PR #14, opened 2016-01-29
Fix the imports for Python 3

pittwolfe:


rabernat:
I hope this doesn't break anything for python 2.7. Too bad we don't have tests yet...

PR #46, opened 2016-10-11
Read llc files

rabernat:
Major updates and reorganization happening here.


rabernat:
This test fails because dask/dask#1646 has not been released yet.

PR #52, opened 2017-03-16
Fix doc typos

rabernat:
Hi, this is amazing package, so cool! But you don't follow PEP8 standards in your code. So I fixed it.


rabernat:
This actually sucks really bad.

PR #53, opened 2017-03-16
Dhruv_fixes_colons

dhruvbalwada:
Your colons were messed up on the install page. So I fixed it.


ci/environment-py35.yml
```diff
@@ -5,7 +5,6 @@ dependencies:
   - dask
   - numpy
   - pytest
-  - future
```
rabernat: you need to put future back in...

Issue #73, opened 2017-09-01
Automatic grid generation


rabernat:
cc @jbusecke, @naomi-henderson

All of the basic functionality in xgcm now could easily be applied to generic gridded datasets (e.g. satellite, model data without auxiliary coordinates, etc.), _if_ we had the ability to automatically generate other axis variables. (This may also be necessary for @naomi-henderson to move forward with pangeo-data/pangeo-discussion#1.)

What would this look like? Imagine we have an xarray dataset that looks like this
```python
import xarray as xr
import numpy as np
ds = xr.DataArray(np.random.rand(180, 360),
                         dims=['lat', 'lon'],
                         coords={'lon': np.arange(0, 360)+0.5,
                                 'lat': np.arange(-90, 90)+0.5}
                        ).to_dataset(name='somedata')
ds
```
```
<xarray.Dataset>
Dimensions:   (lat: 180, lon: 360)
Coordinates:
  * lon       (lon) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5 10.5 ...
  * lat       (lat) float64 -89.5 -88.5 -87.5 -86.5 -85.5 -84.5 -83.5 -82.5 ...
Data variables:
    somedata  (lat, lon) float64 0.2838 0.3383 0.1441 0.8124 0.6643 0.6623 ...
```
This is a dataset with the values located at the midpoints between lat and lon.

It would be amazing to be able to say this
```python
grid = xgcm.autogenerate_axis(ds, axes={'X': 'lon', 'Y': 'lat'})
```
There are lots of potential options we would want to include, such as the coordinate positions to generate (i.e. left, right, inner, outer). I'm also assuming we would want to treat the original coordinates as the cell centers, but maybe that's not always the case.

This could be implemented via a lower level function called `autogenerate_axis` that just works on one axis at a time (the way the `Grid` object works now). I would recommend creating a new module called `autogenerate` for these functions.

A reminder of the lexicon of grid positions
```
 Center
 |------o-------|------o-------|------o-------|------o-------|
       [0]            [1]            [2]            [3]

 Left
 |------o-------|------o-------|------o-------|------o-------|
[0]            [1]            [2]            [3]

 Right
 |------o-------|------o-------|------o-------|------o-------|
               [0]            [1]            [2]            [3]

 Inner
 |------o-------|------o-------|------o-------|------o-------|
               [0]            [1]            [2]

 Outer
[...]

PR #74, opened 2017-10-04
basic grid autogenerate logic

jbusecke:
Adds basic functionality to create 'c-grid style' coordinates to xarray datasets (e.g. observational datasets)

### DIFF
--- a/xgcm/autogenerate.py
+++ b/xgcm/autogenerate.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 from future.utils import iteritems
 from xgcm.grid import Axis, raw_interp_function
+import xarray as xr
 
 
 def generate_axis(ds,
@@ -46,6 +47,9 @@ def generate_axis(ds,
         enabled for dimensions and deactivated for multidimensional
         coordinates. These can only be calculated after the dims are created.
     """
+    if not isinstance(ds, xr.Dataset):
+        raise RuntimeError("'ds' needs to be xarray.Dataset")
+
     if new_name is None:
         new_name = name+'_inferred'
 
--- a/xgcm/autogenerate.py
+++ b/xgcm/autogenerate.py
@@ -92,22 +92,31 @@ def generate_axis(ds,
         # or xgcm.Axis throws error. Will be rewrapped below.
         ds[name] = fill_attrs(ds[name], 'center', axis)
 
-        ax = Axis(ds, axis, periodic=periodic, wrap=wrap, raw_out=True)
-        ds.coords[new_name] = ax.interp(ds[name], relative_pos_to,
-                                        boundary=boundary,
-                                        fill_value=fill_value)
+        ax = Axis(ds, axis, periodic=periodic, wrap=wrap)
+        ds.coords[new_name] = \
+            ax._neighbor_binary_func_raw(ds[name],
+                                         raw_interp,
+                                         relative_pos_to,
+                                         boundary=boundary,
+                                         fill_value=fill_value)
+        # ds.coords[new_name] = ax.interp(ds[name], relative_pos_to,
+        #                                 boundary=boundary,
+        #                                 fill_value=fill_value)
+        # ax__neighbor_binary_func_raw(da, ax., to,
+        #                                           boundary=boundary,
+        #                                           fill_value=fill_value)
 
         # Place the correct attributes
         ds[name] = fill_attrs(ds[name], pos_from, axis)
         ds[new_name] = fill_attrs(ds[new_name], pos_to, axis)
     else:
-        ax = Axis(ds, axis, periodic=periodic, wrap=wrap, raw_out=False)
+        ax = Axis(ds, axis, periodic=periodic, wrap=wrap)
         ds.coords[new_name] = ax.interp(ds[name], pos_to, boundary=boundary,
                                         fill_value=fill_value)
     return ds
 
 
-def autogenerate_grid_ds(ds,
+def generate_grid_ds(ds,
                          axes_dims_dict,
                          axes_coords_dict=None,
                          position=None,
@@ -178,16 +187,17 @@ def autogenerate_grid_ds(ds,
     return ds
 
 
-def parse_wrap_pad(wrap, varname):
-    if isinstance(wrap, dict):
+def parse_wrap_pad(in_val, varname):
+    """Parse wrap or pad parameters"""
+    if isinstance(in_val, dict):
         try:
-            is_wrapped = wrap[varname]
+            is_valued = in_val[varname]
         except KeyError:
             # Set defaults
-            is_wrapped = None
+            is_valued = None
     else:
-        is_wrapped = wrap
-    return is_wrapped
+        is_valued = in_val
+    return is_valued
 
 
 def parse_position(position, axname, pos_default=('center', 'left')):
@@ -222,6 +232,13 @@ def position_to_relative(pos_from, pos_to):
         raise RuntimeError("Cannot infer '%s' coordinates \
     from '%s'" % (pos_to, pos_from))
     return to
+
+
+def raw_interp(data_left, data_right):
+    # TODO: This is not great, but I am not sure how to pass the nested
+    # function in xgcm.grid.Axis.interp to Axis._neighbor_binary_func_raw
+    # instead.
+    return 0.5*(data_left + data_right)
 
 
 def auto_pad(da, dim):
--- a/xgcm/autogenerate.py
+++ b/xgcm/autogenerate.py
@@ -1,4 +1,4 @@
-from .grid import Axis
+from .grid import Axis,Grid
 import docrep
 
 docstrings = docrep.DocstringProcessor(doc_key='My doc string')
@@ -35,7 +35,7 @@ class AutoGenerate(Axis):
 
             wrap_value : float, optional
                 if value is given, cyclic coordinate is wrapped approriately,
-                e.g. longitude is wrapped around 190 or 360 discontinuity
+                e.g. longitude is wrapped around 180 or 360 discontinuity
             Returns
             -------
             da_i : xarray.DataArray
@@ -52,8 +52,7 @@ class AutoGenerate(Axis):
             data_left, data_right = \
                 self._get_neighbor_data_pairs(da, to,
                                               boundary=boundary,
-                                              fill_value=fill_value,
-                                              wrap=wrap)
+                                              fill_value=fill_value)
 
             # Is there a way to use the existing grid.Axis.interp logic?
             # For now just hardcode it
@@ -62,42 +61,162 @@ class AutoGenerate(Axis):
             return data_new
 
 
+def generate(ds, axes, dims, to,
+             x_wrap, y_wrap, z_pad,
+             to_shift, from_shift, dim_switch=True):
+        """
+        Generate new dataarray from coordinates or dimensions, which might be
+        missing in order to get a 'full' c-grid.
+
+        INPUTS:
+
+        ds - xarray.Dataset
+
+        axes = dictionary containing comodo axis and corresponding DataArray
+        in ds (e.g. axes={'X':'llon','Y':'llon'}). These could be
+        multidimensional coordinates like geospatial longitudes/latitudes
+
+        dims = dictionary of dimensons in ds corresponding to comodo axis
+        e.g. (dims={'X':'lon','Y':'lon'}). These have to be names of
+        strictly dimensions of ds.
+
+        to - {'left','right'} Direction of shift, similar as the logic in
+        xgcm.Axis, needed to apply shift correctly
+
+        x_wrap - float: value of discontinuity in x direction (defaults to
+        360 for global longitude)
+
+        y_wrap - float: as x_wrap but for y direction (defaults to
+        180 for global longitude)
+
+        z_pad - float: boundary padding value for the depth interpolation. If
+        not specified it will default to the min() or max() of the appropriate
+        field, depending on 'to'
+
+        dim_switch - Bool: switch between input for dimesions (need to
+        load class from xgcm.autogenerate) and coordinates (uses existing
+        xgc.grid logic)
+        """
+
+        #TODO: I can definitely infer the dims from the comodo stuff, but for
+        #now leave it explicit
+
+        for ki in axes.keys():
+            if ki not in dims.keys():
+                raise RuntimeError('dimension '+ki+' was not specified in axes \
+                                    input')
+            old_name = axes[ki]
+            dim = dims[ki]
+            new_name = old_name+'_inferred'
+
+            if ki == 'X':
+                periodic = True
+                boundary = None
+                fill_value = 0.0
+                wrap = x_wrap
+            elif ki == 'Y':
+                periodic = True
+                boundary = None
+                fill_value = 0.0
+                wrap = y_wrap
+            elif ki == 'Z':
+                periodic = False
+                boundary = 'fill'
+                # For depth decide between surface or bottom fill
+                if z_pad is None:
+                    # Infer limits from depth data
+                    z_min = ds[old_name].min().data
+                    z_max = ds[old_name].max().data
+                    # The value below is an attempt on extrapolating the values
+                    # linearly. The difference between values at the top and
+                    # bottom is used to pad the depth array at the appropriate
+                    # boundary. For multidimensional arrays the min difference
+                    # is chosen. This could lead to undesired results if the
+                    # depth spacing is spatially irregular (though thats a rare
+                    # case IMO)
+                    top_diff = ds[old_name].diff(dim).isel(**{dim: 0}).min()
+                    bot_diff = ds[old_name].diff(dim).isel(**{dim: -1}).min()
+                    top = z_min.data - top_diff.data
+                    bot = z_max.data + bot_diff.data
+                    if to is 'left':
+                        fill_value = top
+                    elif to is 'right':
+                        fill_value = bot
+                else:
+                    fill_value = z_pad
+
+            # Input coordinate has to be declared as center,
+            # or xgcm.Axis throws error. Will be rewrapped below.
+            attrs = ds[old_name].attrs
+            attrs['axis'] = ki
+            attrs.pop('c_grid_axis_shift', None)
+            ds[old_name].attrs = attrs
+
+            if dim_switch:
+                ax = AutoGenerate(ds, ki, periodic=periodic, wrap=wrap)
+                ds.coords[new_name] = ax._generate_coord(ds[old_name], to,
+                                                         boundary=boundary,
+                                                         fill_value=fill_value)
+
+            else:
+                ax = Axis(ds, ki, periodic=periodic, wrap=wrap)
+                ds.coords[new_name] = ax.interp(ds[old_name], boundary=boundary,
+                                         fill_value=fill_value)
+
+            attrs_new = ds[new_name].attrs
+            attrs_new['axis'] = ki
+            attrs_new.pop('c_grid_axis_shift', None)
+            # Reset appropriate attributes for old and new coordinates
+            if abs(to_shift) > 0.0:
+                attrs_new['c_grid_axis_shift'] = to_shift
+            ds[new_name].attrs = attrs_new
+
+            if abs(from_shift) > 0.0:
+                ds[old_name].attrs['c_grid_axis_shift'] = from_shift
+
+        return ds
+
+
 def autogenerate_ds(ds,
                     axes={'X': 'lon', 'Y': 'lat'},
                     position='center',
+                    coord_axes=None,
                     x_wrap=360,
                     y_wrap=180,
-                    z_pad=None):
+                    z_pad=None,
+                    x_coord_wrap=360,
+                    y_coord_wrap=180,
+                    z_coord_pad=None):
     """
-    Add missing axis coordinates to dataset.
+    Regenerate all c-grid information from an existing dataset
 
     Parameters
     ----------
     ds : xarray.Dataset
         The data on which to operate
+    axes : axes on which to regenerate dims (and optional coords, see
+        'coord_axes' input). Needs to be in the form of a dict with keys
+        ('X','Y'or'Z'), the values need to be the corresponding dimension names
+        in ds.
     position : {'center', 'left', 'right'}
         The position of the datapoints relative to the coordinates given
         This will usually be the cell center, but can sometimes be the
         lower left (option:'left') or perhaps upper right (option:'right')
-        corner of the gridcell
-
-    boundary : {None, 'fill', 'extend'}
-        A flag indicating how to handle boundaries:
-
-        * None:  Do not apply any boundary conditions. Raise an error if
-          boundary conditions are required for the operation.
-        * 'fill':  Set values outside the array boundary to fill_value
-          (i.e. a Neumann boundary condition.)
-        * 'extend': Set values outside the array to the nearest array
-          value. (i.e. a limited form of Dirichlet boundary condition.)
-
-    fill_value : float, optional
-         The value to use in the boundary condition with `boundary='fill'`.
+        corner of the gridcell.
+    coord_axes : (optional) Like 'axes' but for multidimensional coordinates in
+        ds
+    x_wrap : float - Value of discontinuity at the 'X' boundary
+    y_wrap : float - Value of discontinuity at the 'Y' boundary
+    z_pad : float -  Depth padding value. If None, it will be determined from
+        data
+    x_coord_wrap : float - Like 'x_wrap' but for coordinate variable
+    y_coord_wrap : float - Like 'y_wrap' but for coordinate variable
+    z_coord_pad : float -  Like 'z_pad' but for coordinate variable
 
     Returns
     -------
-    da_i : xarray.DataArray
-        The differenced data
+    ds : xarray.DataArray
+        Dataset with inferred dims (and coordinates)
     """
 
     # set position of original coordinates and define how to interpolated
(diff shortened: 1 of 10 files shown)


rabernat:
Wow this is great! I will have a close look tomorrow.

jbusecke:
OK @rabernat, this should look a bit better than the first commit! Excited to hear what you think.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+                    top_diff = ds[old_name].diff(dim).isel(**{dim: 0}).min()
+                    bot_diff = ds[old_name].diff(dim).isel(**{dim: -1}).min()
+                    top = z_min.data - top_diff.data
+                    bot = z_max.data + bot_diff.data
+                    if to is 'left':
+                        fill_value = top
+                    elif to is 'right':
+                        fill_value = bot
+                else:
+                    fill_value = z_pad
```
rabernat: I'm not sure if the design of this function is optimal. Could we not just have a generic `generate_axis` function that has sufficient flexibility via kwargs to apply to any scenario? Then we would loop over axes in `autogenerate_ds` and call this `generate_axis` function on each.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+        The data on which to operate
+    axes : axes on which to regenerate dims (and optional coords, see
+        'coord_axes' input). Needs to be in the form of a dict with keys
+        ('X','Y'or'Z'), the values need to be the corresponding dimension names
+        in ds.
+    position : {'center', 'left', 'right'}
+        The position of the datapoints relative to the coordinates given
+        This will usually be the cell center, but can sometimes be the
+        lower left (option:'left') or perhaps upper right (option:'right')
+        corner of the gridcell.
```
rabernat: Shouldn't this be axis dependent? Could one specific a dict for position like `{'X': 'center', 'Y': 'right', 'Z' left'}`?

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+    coord_axes : (optional) Like 'axes' but for multidimensional coordinates in
+        ds
+    x_wrap : float - Value of discontinuity at the 'X' boundary
+    y_wrap : float - Value of discontinuity at the 'Y' boundary
+    z_pad : float -  Depth padding value. If None, it will be determined from
+        data
+    x_coord_wrap : float - Like 'x_wrap' but for coordinate variable
+    y_coord_wrap : float - Like 'y_wrap' but for coordinate variable
+    z_coord_pad : float -  Like 'z_pad' but for coordinate variable
```
rabernat: Again, I am opposed to hard-coding the possible axes. The design should be more general and less redundant.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+        to = 'right'
+        to_shift = 0
+        from_shift = -0.5
+
+    # Do we want to modify in place? or copy?
+    ds = ds.copy()
+
+    # Loop over all specified axes and build new coordinates
+    ds = generate(ds, axes, axes, to, x_wrap, y_wrap,
+                  z_pad, to_shift, from_shift, dim_switch=True)
```
rabernat: I don't see a loop. Why do you say "loop over all specified axes"? Maybe that loop should be in this function, not `generate`...

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # Generate shifted multidimensional coordinates if specified in
+    # 'coord_axes'
+    if coord_axes:
+        ds = generate(ds, coord_axes, axes, to, x_coord_wrap, y_coord_wrap,
+                      z_coord_pad, to_shift, from_shift, dim_switch=False)
```
rabernat: It seems weird to be calling `generate` again. Can you explain what is going on here?

xgcm/grid.py
```diff
@@ -555,6 +569,15 @@ def cumsum(self, da, axis, **kwargs):
         return ax.cumsum(da, **kwargs)
 
 
+def add_slice(da, dim, sl, added):
+    axis = da.get_axis_num(dim)
+    t_order = list(range(len(da.data.shape)))
+    t_order[axis] = 0
+    t_order[0] = axis
+    data = np.transpose(da.data.copy(), axes=t_order)
```
rabernat: Does calling transpose on dask data trigger computation? If so, this is forbidden.

rabernat: Can you explain what this function does?

xgcm/test/test_autogenerate.py
```diff
@@ -0,0 +1,152 @@
...
+
+    left = autogenerate_ds(ds_original, axes=axes,
+                           position='left', coord_axes=coord_axes)
+
+    right = autogenerate_ds(ds_original, axes=axes,
+                            position='right', coord_axes=coord_axes)
+
+    xr.testing.assert_identical(center, ds_out_center)
+    xr.testing.assert_identical(left, ds_out_left)
+    xr.testing.assert_identical(right, ds_out_right)
```
rabernat: I really like your test!

It is, however, a pretty high level test. It might be valuable to break down your new module into smaller units and write some unit tests.

xgcm/test/test_grid.py
```diff
@@ -432,3 +432,16 @@ def test_grid_ops(all_datasets):
...
+    da = xr.DataArray(np.ones([2, 2, 3]),
+                      dims=['lat', 'z', 'lon'])
+
+    new = add_slice(da, 'lon', 1, 3.0)
+
+    ref = np.array([[[1., 4., 1.],
+                    [1., 4., 1.]],
+                    [[1., 4., 1.],
+                    [1., 4., 1.]]])
+    np.testing.assert_allclose(new, ref)
```
rabernat: Can you test this function on dask data and ensure that it does not trigger computation?

rabernat:
We probably could have discussed the API and design a bit more before diving in, but I'm really glad you got this started.

Here's how I would have suggested approaching it.

- Put all the heavy lifting into a function called something like `autogenerate_axis_coords` which operates on a single axis at a time. It would have lots of options (boundary conditions, etc.). It would return a new dataset with extra coords in it.
- Have a simple function called `autogenerate_grid_coords` which just loops over the axes and calls `autogenerate_axis_coords` for each one.
- Have convenience methods called `autogenerate_grid` and `autogenerate_coords` which call the above methods and then make `Axis` and `Grid` objects out of them.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+            wrap_value : float, optional
+                if value is given, cyclic coordinate is wrapped approriately,
+                e.g. longitude is wrapped around 180 or 360 discontinuity
+            Returns
+            -------
+            da_i : xarray.DataArray
+                The differenced data
+            """
+            #TODO I think we can modify xgcm.Axis so that this can all be
+            # absorbed in there.
```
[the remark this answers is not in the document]
jbusecke: I think we can achieve this functionality (which is mostly copied and pasted) by modifying the xgcm.Axes object, then we can completely remove this from here.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+            if to is None:
+                to = self._default_shifts[position_from]
+
+            # get the two neighboring sets of raw data
+            data_left, data_right = \
+                self._get_neighbor_data_pairs(da, to,
+                                              boundary=boundary,
+                                              fill_value=fill_value)
+
+            # Is there a way to use the existing grid.Axis.interp logic?
```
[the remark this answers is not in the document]
jbusecke: Even more reason to absorb into xgcm.Axis

xgcm/autogenerate.py
```diff
@@ -0,0 +1,254 @@
...
+        strictly dimensions of ds.
+
+        to - {'left','right'} Direction of shift, similar as the logic in
+        xgcm.Axis, needed to apply shift correctly
+
+        x_wrap - float: value of discontinuity in x direction (defaults to
+        360 for global longitude)
+
+        y_wrap - float: as x_wrap but for y direction (defaults to
+        180 for global longitude)
```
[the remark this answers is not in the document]
jbusecke: My intention was to make this work relatively smoothly "out of the box with common datasets (like satellite products). The logic should work with any succession of values that need to be wrapped around a discontinuity. This would have to be explained better in the docstring, but if you have an arbitrary dataset with x coordinates 1,2,...10 you would have to specify 'x_wrap=10' and it should do the right thing. Should have a test too.

jbusecke:
Thanks for the detailed input, Ryan. I agree that this can all be made much more general, but I just wanted to start somewhere and get something that achieves the basics. I think a lot of the hard coding can be removed with proper 'bookeeping' of the kwargs.

Ill get working on that on the train. Yay, commuting!

xgcm/grid.py
jbusecke: This function was my attempt to enable the 'wrapping' of values around a discontinuity. The basic idea is to add a 'wrap value' to the first or last (depending on the shift_to direction) slice from an arbitrary axis (xarray dimension). Is there a way to selectively replace values in a dask array without triggering computation?

Alternatively we might have to compute and then return to a dask array?

xgcm/autogenerate.py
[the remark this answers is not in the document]
jbusecke: Oh sorry I guess I only read the last part of the comment. Ill see what I can do!

xgcm/autogenerate.py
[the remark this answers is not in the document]
rabernat: I would avoid modifying the Axis class. This can all be accomplished with a simple function. No classes needed. (See my comment below).

xgcm/autogenerate.py
```diff
@@ -1,254 +1,240 @@
...
+                  pad=None,
+                  raw_switch=True):
+    """
+    Parameters
+    ----------
+    ds : xarray.Dataset
+        Dataset with gridinformation used to construct c-grid
+    axis_dict : dict
+        Dict with information on the dimension in ds corrsponding to the xgcm
+        axis. E.g. {'X':'lon'}
```
rabernat: I don't see `axis_dict` in the function arguments

xgcm/autogenerate.py
```diff
@@ -1,254 +1,240 @@
...
-                ds.coords[new_name] = ax._generate_coord(ds[old_name], to,
-                                                         boundary=boundary,
-                                                         fill_value=fill_value)
+from .grid import Axis
+
+
+def generate_axis(ds,
+                  axis,
+                  name,
+                  axis_dim,
```
rabernat: The docstring is not consistent with the function arguments

xgcm/autogenerate.py
```diff
@@ -1,254 +1,240 @@
...
+                  raw_switch=True):
+    """
+    Parameters
+    ----------
+    ds : xarray.Dataset
+        Dataset with gridinformation used to construct c-grid
+    axis_dict : dict
+        Dict with information on the dimension in ds corrsponding to the xgcm
+        axis. E.g. {'X':'lon'}
+    pos_from : str
```
rabernat: should be `pos_from : {'center','left','right'}`

jbusecke:
Ok I rewrote the module substantially. I think I at least started addressing most of your comments. I have still a bunch of cleaning up to do, but I wanted to get some feedback first.
A typical usecase would be this:
```python
from xgcm import Axis, Grid
import xarray as xr
import numpy as np
from dask.array import from_array
from xgcm.autogenerate import autogenerate_grid_ds

# create test datasets
dx = 0.5
dy = 1.0
dz = 0.5
a = np.random.rand(180, int(360/dx), int(10/dz))

x = np.arange(-180, 180, dx)
y = np.arange(-90, 90, dy)
z = np.arange(0, 10, dz)

xx, yy = np.meshgrid(x, y)
_, _, zz = np.meshgrid(x, y, z)

ds_original = xr.Dataset(
                         {'somedata': (['lat', 'lon', 'z'], a)},
                         coords={'lon': (['lon', ], x+(dx/2.0)),
                                 'lat': (['lat', ], y+(dy/2.0)),
                                 'z': (['z', ], z+(dx/2.0)),
                                 'llon': (['lat', 'lon'], xx+(dx/2.0)),
                                 'llat': (['lat', 'lon'], yy+(dy/2.0)),
                                 'zz': (['lat', 'lon', 'z'], zz+(dx/2.0))}
                        )

axis_dims =  {'X':'lon','Y':'lat','Z':'z'}
axis_coords =  {'X':'llon','Y':'llat','Z':'zz'}
ds_new = autogenerate_grid_ds(ds_original,
                              axis_dims,
                              axis_coords,
                              wrap={'lon':360, 'lat':180,
                                    'llon':360, 'llat':180},
                              pad={'z':'auto','zz':'auto'})
ds_new
```

xgcm/autogenerate.py
```diff
@@ -1,254 +1,240 @@
...
+        periodic = True
+        fill_value = 0.0
+        boundary = None
+    else:
+        raise RuntimeError('Either "wrap" or "pad" have to be specified')
+
+    if raw_switch:
+        # Input coordinate has to be declared as center,
+        # or xgcm.Axis throws error. Will be rewrapped below.
+        ds[name] = fill_attrs(ds[name], 'center', axis)
```
rabernat: I would probably copy `ds` before adding new variables to it

xgcm/grid.py
```diff
@@ -227,13 +227,17 @@ def _neighbor_binary_func(self, da, f, to, boundary=None, fill_value=0.0):
             to = self._default_shifts[position_from]
 
         # get the two neighboring sets of raw data
-        data_left, data_right = self._get_neighbor_data_pairs(da, to,
-                                      boundary=boundary, fill_value=fill_value)
+        data_left, data_right = self._get_neighbor_data_pairs(da, to,\
```
rabernat: please take this `\` out

xgcm/grid.py
```diff
@@ -74,6 +73,7 @@ def __init__(self, ds, axis_name, periodic=True, default_shifts={},
         self._name = axis_name
         self._periodic = periodic
         self._wrap = wrap
+        self._raw_out = raw_out
```
rabernat: What is this attribute?!?

xgcm/grid.py
```diff
@@ -569,14 +573,21 @@ def cumsum(self, da, axis, **kwargs):
...
-    t_order[0] = axis
-    data = np.transpose(da.data.copy(), axes=t_order)
-    data[sl, ...] = data[sl, ...] + added
-    return np.transpose(data, t_order)
+def add_to_slice(da, dim, sl, value):
+    # split array into before, middle and after (if slice is the
+    # beginning or end before or after will be empty)
+    before = da[{dim: slice(0, sl)}]
+    middle = da[{dim: sl}]
+    after = da[{dim: slice(sl+1, None)}]
```
rabernat: perfect...this should be dask friendly

xgcm/autogenerate.py
```diff
@@ -0,0 +1,240 @@
...
+                  axis,
+                  name,
+                  axis_dim,
+                  pos_from='center',
+                  pos_to='left',
+                  wrap=None,
+                  pad=None,
+                  raw_switch=True):
+    """
+    Parameters
```
rabernat: We need a description of what the function does.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,240 @@
...
+        If float is specified, that value is used as padding. Auto attempts to
+        pad linearly extrapolated values. Can be useful for e.g. depth
+        coordinates (to reconstruct 0 depth). Can lead to unexpected values
+        when coordinate is multidimensional.
+    raw_switch : bool
+        Determines if the attributes are created from scratch. Should be
+        enabled for dimensions and deactivated for multidimensional
+        coordinates. These can only be calculated after the dims are created.
+    """
+    new_name = name+'_inferred'
```
rabernat: this suffix should probably be a keyword argument

xgcm/autogenerate.py
```diff
@@ -0,0 +1,240 @@
...
+                                        boundary=boundary,
+                                        fill_value=fill_value)
+
+        # Place the correct attributes
+        ds[name] = fill_attrs(ds[name], pos_from, axis)
+        ds[new_name] = fill_attrs(ds[new_name], pos_to, axis)
+    else:
+        ax = Axis(ds, axis, periodic=periodic, wrap=wrap, raw_out=False)
+        ds.coords[new_name] = ax.interp(ds[name], pos_to, boundary=boundary,
+                                        fill_value=fill_value)
```
rabernat: Can you help me understand what is happening here? What is `raw_switch`? What do these two cases represent?

jbusecke: Ok so for a set of coordinates there are two fundamental cases. The coordinates are a) one dimensional (dimensions) or 2) multidimensional. These are separated by the keyword  raw_switch.

These two cases are treated differently because for each dataset we need to recreate all a) cases before we can proceed to 2), hence this is really the 'raw' data processing step. If we have working one dimensional coordinates (e.g. after we looped over the `axes_dims_dict`, we can use the regular xgcm.Axis to interpolate multidimensional coordinates. 
I designed it this way so that any changes to the Axis.interp method can directly propagate to this module.

Does that make any sense?

xgcm/autogenerate.py
jbusecke: Fixed it.

xgcm/autogenerate.py
jbusecke: Good call, that caused some issues in the tests earlier. Added `ds=ds.copy` before

xgcm/grid.py
jbusecke: is there a preferred way to test if a computation was triggered?

xgcm/grid.py
jbusecke: Done.

xgcm/grid.py
jbusecke: This is basically a switch to omit the call to `_wrap_and_replace_coords` at the end. This way I can use all the logic from `xgcm.Axis` to do the interpolation (as mentioned above). I think this is much better then keeping a branched out interp function in the autogenerate module.

jbusecke:
Just pushed a commit with fixes requested above.

xgcm/autogenerate.py
rabernat: Now I understand and it makes sense.

That would have been a good comment to include in the code.

jbusecke: It is included now

jbusecke:
@rabernat, could you take a look at why the python 2 test is failing? Is there an easy way to fix that? Can I set the broadcasting rule like in python 3?

jbusecke:
I will work on getting the test coverage up.

jbusecke:
@rabernat, should I clean and finish this part as is right now or also try to integrate some more logic, to calculate distances, areas, and volumes? This would be complimentary to development in #75. I think we should probably keep that seperate and finish this now. Just want to hear what you think.

rabernat:
An alternative instead of the `raw_switch` approach.

Rewrite `Grid._neighbor_binary_ufunc` as
```
    # returns numpy or dask array
    data_new = self._binary_ufunc_raw(...)
    return self._wrap_and_replace_coords(da, data_new, to)
```

And then in your code, you call `_binary_ufunc_raw`. This function would need to be defined in `Grid`.

xgcm/grid.py
```diff
@@ -284,12 +293,22 @@ def _get_neighbor_data_pairs(self, da, position_to, boundary=None,
             left = da.data
         elif (self._periodic and ((transition == ('center', 'left')) or
                                   (transition == ('right', 'center')))):
-            left = da.roll(**{dim: 1}).data
+
+            if self._wrap is not None:
+                left = da.roll(**{dim: 1})
+                left = add_to_slice(left, dim, 0, -self._wrap).data
+            else:
+                left = da.roll(**{dim: 1}).data
```
rabernat: can we put this wrap logic in standalone function?

xgcm/grid.py
```diff
@@ -284,12 +293,22 @@ def _get_neighbor_data_pairs(self, da, position_to, boundary=None,
...
             right = da.data
         elif (self._periodic and ((transition == ('center', 'right')) or
                                   (transition == ('left', 'center')))):
             left = da.data
-            right = da.roll(**{dim: -1}).data
+            if self._wrap is not None:
+                right = da.roll(**{dim: -1})
+                right = add_to_slice(right, dim, -1, self._wrap).data
+            else:
+                right = da.roll(**{dim: -1}).data
```
rabernat: Ditto

jbusecke:
I am not able to “access” the basic logic in ‘interp_function’ within Axis.interp. Can You advise me how to pass the inner function as an input?

xgcm/grid.py
jbusecke: I think yes, but I would probably have to repeat all the positional logic. I did sinplify the wrap logic, soo maybe that is acceptable?

jbusecke:
Ok I added tests for everything and removed the 'raw' switch.

xgcm/autogenerate.py
```diff
@@ -224,6 +234,13 @@ def position_to_relative(pos_from, pos_to):
     return to
 
 
+def raw_interp(data_left, data_right):
+    # TODO: This is not great, but I am not sure how to pass the nested
+    # function in xgcm.grid.Axis.interp to Axis._neighbor_binary_func_raw
+    # instead.
+    return 0.5*(data_left + data_right)
```
rabernat: You are correct that you can't extract that function because it is defined within a another. But there is a simple solution: define it at the top level of the grid module namespace and then use the same function in your routine and `interp`.

What I mean is to just define
```python
def interp_raw(data_left, data_right):
    # linear, centered interpolation
    # TODO: generalize to higher order interpolation
    return 0.5*(data_left + data_right)

def diff_raw ...
```
in grid.py, change the `diff` and `interp` methods to call these functions, and then import the same `_raw` functions in your module.

Does this make sense to you?

xgcm/autogenerate.py
```diff
@@ -0,0 +1,274 @@
...
+        If float is specified, that value is used as padding. Auto attempts to
+        pad linearly extrapolated values. Can be useful for e.g. depth
+        coordinates (to reconstruct 0 depth). Can lead to unexpected values
+        when coordinate is multidimensional.
+    new_name : str
+        Name of the inferred grid variable. Defaults to name+'_inferred'
+    raw_switch : bool
+        Determines if the attributes are created from scratch. Should be
+        enabled for dimensions and deactivated for multidimensional
+        coordinates. These can only be calculated after the dims are created.
```
rabernat: I would still prefer a different name for `raw_switch`. Why not `create_attributes_from_scratch`? Something more verbose that corresponds with what the option actually does.

rabernat:
Still need:

- [ ] documentation (like a new page called `doc/grid_generation.rst` and / or an example notebook modeled on `/doc/example_mitgcm.ipynb`)
- [ ] fix the python 2.7 test failure

jbusecke:
Regarding the python 2.7 test failure: It seems like some numpy functions behave differently (e.g. the assert_isclose() function). Is there a simple way to import the python 3 behaviour as for the print statement?

rabernat:
> Regarding the python 2.7 test failure: It seems like some numpy functions behave differently (e.g. the assert_isclose() function). Is there a simple way to import the python 3 behaviour as for the print statement?

You're gonna have to sort this out on your own.

xgcm/autogenerate.py
```diff
@@ -272,3 +285,28 @@ def fill_attrs(da, pos, axis):
...
+    dy  : xarray.DataArray distance inferred from dlat
+    """
+
+    # First attempt with super cheap approach...
+    #     111km for each deg lat and then scale that by cos(lat) for lon
+    dll_factor = 111000.0
+    dx = dlon * np.cos(np.deg2rad(lat.data)) * dll_factor
+    dy = dlat * dll_factor
+
+    return dx, dy
```
rabernat: This function looks useful, but why is it in this PR? It seems like a new, separate feature.

jbusecke:
Seems like the Travis build is passing now. While writing the docs I decided to implement the distance calculation at this point already. This requires a few more functions and mostly tests. Should be ready soon.

rabernat:
> While writing the docs I decided to implement the distance calculation at this point already. 
Not in this PR. We need to discuss the design and API for that stuff separately.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,312 @@
...
+     variable ({dict} e.g. {'z':'auto','latitude':0.0})
+    new_name : str
+     Name of the inferred grid variable. Defaults to name+'_inferred'
+    generate_distance : Bool
+     Switch for the generation of distances based on the axes_coords_dict
+    distance_name : str
+     Name prefix for distance coordinates. Defaults to 'd'+name.
+    ll_dist : Bool
+     Activates the conversion of lon/lat coordinates to distances in meters.
+     Requires coordinates to be specified on 'X' and 'Y'.
```
rabernat: It doesn't look like this option is ever utilized.

I'm confused about what it would mean. How can you covert lon/lat coordinates to distance without specifying a projection?

rabernat:
Just to be clear...I am very very happy with this PR :clap: :clap: :clap:!

I just don't want to clutter it up with new features just as we are getting ready to merge.

jbusecke:
Ahh shoot. This was a big git fail...I am moving these new features into a new branch where they should have gone in the first place. Need to finish the docs still.

xgcm/autogenerate.py
```diff
@@ -0,0 +1,272 @@
...
+        If float is specified, that value is used as padding. Auto attempts to
+        pad linearly extrapolated values. Can be useful for e.g. depth
+        coordinates (to reconstruct 0 depth). Can lead to unexpected values
+        when coordinate is multidimensional.
+    new_name : str
+        Name of the inferred grid variable. Defaults to name+'_inferred'
+    attrs_from_scratch : bool
+        Determines if the attributes are created from scratch. Should be
+        enabled for dimensions and deactivated for multidimensional
+        coordinates. These can only be calculated after the dims are created.
```
rabernat: all of the optional keyword arguments need an `optional` tag, e.g.
```
    attrs_from_scratch : bool, optional
        ...
```

(see numpydoc HOWTO: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

xgcm/autogenerate.py
```diff
@@ -0,0 +1,272 @@
...
+    """Parse wrap or pad parameters"""
+    if isinstance(in_val, dict):
+        try:
+            is_valued = in_val[varname]
+        except KeyError:
+            # Set defaults
+            is_valued = None
+    else:
+        is_valued = in_val
+    return is_valued
```
rabernat: are these private functions? If so, they need to be renamed with a leading underscore (i.e. `_parse_wrap_pad`)

xgcm/grid.py
```diff
@@ -45,7 +45,8 @@ class Axis:
     differentiated by their length.
     """
 
-    def __init__(self, ds, axis_name, periodic=True, default_shifts={}):
+    def __init__(self, ds, axis_name, periodic=True, default_shifts={},
+                 wrap=None):
```
rabernat: the `wrap` option needs to be added to the docstring for the class.

What does `wrap` do? How does it relate to the `periodic` option?

rabernat:
I am worried there is some redundant treatment of periodic coordinates.

Before this PR, "wrapping" of coordinates was already supported via the `periodic` options. I don't understand what the new `wrap` keyword does in the context of `Grid` and `Axis` objects.

jbusecke:
Coordinate values are discontinuous, even if the boundary is periodic. So in order to interpolate/diff/etc. a tracer field across a periodic boundary it is enough to 'know' the position of the neighboring grid cell (this is done in xgcm already)

For the coordinate values themselves this is not sufficient: If one wants to interpolate the longitude value between the first and the last longitude cell (e.g. lon1= 0 and lon2=359) one ends up with 179.5 which is wrong. Depending on the 'direction' of interpolation the expected result should be 359.5 (achieved by adding the 'wrap' value 360 to lon1) or 0.5 (achieved by subtracting the same value from lon2). This functionality is added with the 'wrap' keyword and the private wrap function.

So the 'wrap' parameter is defining a subset of a periodic boundary, where values show a well-defined discontinuity across said boundary.

I am not sure if there are other use cases for this, but I would argue that it is simple enough and extends the generality of the gcm axis. 

Does that make things clearer? Maybe we change the name to periodic_discontinuity?

rabernat:
Thanks for the explanation. But I still don't understand why `Axis` needs to know about this. As far as I can see, it only comes up when interpolating coordinates. How does `Axis.interp` know whether the variable it is interpolating is a coordinate that needs to be wrapped or a regular variable that does not?

> Maybe we change the name to periodic_discontinuity?
👍  What about `coordinate_discontinuity`? "Wrap" is just too similar to "periodic."

jbusecke:
> What about coordinate_discontinuity? "Wrap" is just too similar to "periodic."
I can do that but I think coordinate is too explicit...maybe just boundary_discontinuity?

The reason i put it in axis is so that it is applied on the level of picking the neighboring points (which does most of the heavy lifting in terms of logic). This ensures that the discontinuity is respected for all functions in axis (diff, interp, cumsum, and potential future ones). I chose this over the option to put wrap in each of the functions for simplicity, since the 'deal with discontinuity' function needs to know the direction of the shift (as explained in the above example).

...but I see now how that is a problem if you want to use the axis object on different fields. 
Let me attempt to transfer that into the function arguments instead of the axis init arguments...

jbusecke:
Ok this seems more useful now. This way people can run `xgcm.Axis.interp(...., boundary_discontinuity=360)` and `xgcm.Axis.diff(...., boundary_discontinuity=360)`, while still using the same axis for the default behaviour (contious fields across boundary). I did not attempt to add it to cumsum.

jbusecke:
Perhaps you could add a test case for the new keyword once everything else is in order? I am still not able to completely decipher you test cases :-)

rabernat:
I like this much more! I think this is just about ready to merge!

xgcm/test/test_autogenerate.py
```diff
@@ -0,0 +1,286 @@
...
+                                         (('center', 'right'), 'right'),
+                                         (('right', 'center'), 'left')])
+def test_position_to_relative(p, relative):
+    assert _position_to_relative(p[0], p[1]) == relative
+
+    with pytest.raises(RuntimeError):
+        _position_to_relative('left', 'right')
+
+
+def test__():
```
rabernat: don't you need a better name for this test function?

rabernat:
OMG I'm so excited!

jbusecke:
Uhoh. Found an error in the documentation. How do I best go about fixing that? New small PR?

rabernat:
> New small PR?

Yes. I assume there will be other issues that come up, so no huge rush.

jbusecke:
Ok I can work on it while advancing #73.

rabernat:
This PR closes #73.

PR #85, opened 2017-11-08
Updated defaults and documentation

jbusecke:


rabernat:
I think we are good to go here!

Issue #89, opened 2018-01-19
create Grid without comodo attributes


rabernat:
Currently the only way to pass grid coordinate information to xgcm is via the dataset attributes.

Instead, we would like to be able to do
```python
grid = xgcm.Grid(ds, axis_geometry={'X': {'left': 'XG', 'center': 'XC'},
                                    'Y': {'left': 'YG', 'center': 'YC'}})
```

At this point maybe we can just skip passing `ds` completely and pass DataArrays directly within `axis_geometry`. That would be a major redesign.

jbusecke:
A question I had, was if this functionality should be an argument in xgcm.Grid or just a stand-alone parser function? 

Perhaps grid specific coordinates, like depth, area, volume, partial cells could be added this way (preferably with some sort of auto search, or presets for common names). Perhaps in the future this could enable a syntax like `grid.weighted_mean(field, weight='area')` where the area is picked out automatically based on the axis information of `field` and the additional attributes mentioned above.

In that case the keyword argument would get very long, and perhaps a stand-alone function would be nicer?

PR #93, opened 2018-02-21
Bypass comodo

rabernat:
This allows us to more easily create `Grid` objects on datasets that don't have comodo attributes set.

Still needs:
- [x] tests
- [ ] doc update


rabernat:
Fixes #89

Issue #92, opened 2018-02-21
grid autogeneration is broken


rabernat:
#88 broke grid autogeneration:
https://travis-ci.org/xgcm/xgcm/jobs/344396979#L919

Needs to be fixed asap.

PR #96, opened 2018-03-26
Fix grid generation

jbusecke:
Fixes the original issue in #92 and adds .min/max functions.


rabernat:
Sorry I have neglected this for so long.

Just so I understand, what is the point of the min/max functions?

jbusecke:
In MOM5 the cell depth at velocity points (B-grid, so upper right corner) is determined by the minimum depth of the sourrounding 4 tracer points  (MOM5 manual section 14.5.1 and 14.5.2). 
It would be nice to have this functionality in xgcm, since it takes care of the shift from tracer point to velocity point. I was on vacation and am now lying in bed with a cold, so no worries about the delay. Have to get back into this myself.

jbusecke:
If I remember correctly, they pass if I take all the cases with the 'autopad' option out (This was only done locally not in this PR). This means the core functionality of the generate_* functions seems restored, but something about the padding/filling is still broken. This option is very helpful for e.g. depth data (I have used it a bunch in the past weeks). 
I think we need some robust tests for cases where the data is not periodic. As far as I could see these cases were not tested.

jbusecke:
I would also like to be able to apply the grid generation logic to go from 'center' to 'outer' coordinates. This will make it trivial to calculate cell distances. 
Not sure if that will make it into this PR, but I could probably invest some time on Thu evening. Hope that helps.

xgcm/test/test_grid.py
```diff
@@ -11,12 +11,22 @@
...
     axis_objs = {ax: Axis(ds, ax) for ax in all_axes}
     return axis_objs
 
 
+def test_get_edge_data():
+    ds = datasets['1d_left']
+    axis = Axis(ds, 'X')
+    assert axis._get_edge_data(
+        ds.XC, fill_value=10000) != axis._get_edge_data(ds.XC)
+    # somehow the fill value does nothing...the results are the same...
```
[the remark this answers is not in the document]
rabernat: @jbusecke now that you are online and active, what do you think about this?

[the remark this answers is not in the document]
jbusecke: Doesnt that also assume the axis is periodic?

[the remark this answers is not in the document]
jbusecke: The need for this functionality is popping up all over the map these days, so this is a high priority.

jbusecke:
I added some new tests and all the old ones are now passing. Noteable addition is that we can now use  the 'outer' position as target for the grid generation. This should facilitate calculating grid distances from observational products! This still needs some more documentation, but I wanted to see if the core changes are ok with you.

jbusecke:
I am having trouble interpreting the failing tests for python 2.7. 
It seems like the `ds.XC.data` is full of strings? Is there some backwards compatibility module that needs to be loaded?

rabernat:
Sorry, not sure what is going on with that.

rabernat:
Based on the output of your print statements:
```
        print(ds.XC)
        print(axis._extend_right(ds.XC, **kw))
```
it looks to me like `axis._extend_right(ds.XC, **kw)` is not returning an xarray dataset but rather a numpy array. When you subsequently access the `.data` attribute, you get a raw memory view, which is then misinterpreted in the assert statement.

So you need to figure out why this function returns a numpy array in python 2.7.

rabernat:
If you want to stop stickler from fixing the same style errors in your PR every time you push, you should merge the local branch to the remote branch. (Then you won't have to push with `--force`.)

jbusecke:
Ok I fixed it. I was expecting the wrong type. I am not sure why it actually passed in the 3.x version but now its correct.
Let me know what you think.

xgcm/autogenerate.py
```diff
@@ -56,25 +56,38 @@ def generate_axis(ds,
 
     # Determine the relative position to interpolate to based on current and
     # desired position
+
     relative_pos_to = _position_to_relative(pos_from, pos_to)
 
+    # This is bloated. We can probably retire the 'auto' logic in favor of
+    # using 'boundary' and 'fill_value'. But first lets see if this all works.
```
rabernat: Agreed. Let's first get things working again and then refactor if necessary.

xgcm/grid.py
```diff
@@ -352,43 +354,74 @@ def face_edge_data(fnum, face_axis, count=1):
...
+            if boundary_discontinuity:
+                if is_left_edge:
+                    edge_data = edge_data - boundary_discontinuity
+                elif not is_left_edge:
+                    edge_data = edge_data + boundary_discontinuity
+
+        # Check consistency of output data
+        if not isinstance(edge_data, type(da.data)):
+            raise RuntimeError('Types do not match edge(%s)/da.data(%s)'
+                               % (type(edge_data), type(da.data)))
```
rabernat: How would we ever hit this situation? I'm not sure I can follow what is happening here.

xgcm/grid.py
```diff
@@ -352,43 +354,74 @@ def face_edge_data(fnum, face_axis, count=1):
...
         return edge_data
 
-
     def _extend_left(self, da, boundary=None, fill_value=0.0,
-                     ignore_connections=False, vector_partner=None):
+                     ignore_connections=False, vector_partner=None,
+                     boundary_discontinuity=None):
+        if not isinstance(da, xr.DataArray):
+            raise RuntimeError('Input `da` needs to be a xarray.DataArray. \
+                                Is:%s' % (type(da)))
```
rabernat: I would prefer not to do type checking on private methods. Since they are private (starting with underscore), we can assume they will not be misused. Where in our code would this error be raised?

xgcm/grid.py
```diff
@@ -352,43 +354,74 @@ def face_edge_data(fnum, face_axis, count=1):
...
+            boundary=boundary,
+            fill_value=fill_value,
+            ignore_connections=ignore_connections,
+            vector_partner=vector_partner,
+            boundary_discontinuity=boundary_discontinuity
+        )
+        edge_data = self._get_edge_data(da, **kw)
+        if not isinstance(edge_data, type(da.data)):
+            raise RuntimeError('Mixed data type for edge (%s) and data(%s)'
+                               % (type(edge_data), type(da.data)))
```
rabernat: same

xgcm/grid.py
```diff
@@ -352,43 +354,74 @@ def face_edge_data(fnum, face_axis, count=1):
...
         return concatenate([edge_data, da.data], axis=axis_num)
 
-
     def _extend_right(self, da, boundary=None, fill_value=0.0,
-                      ignore_connections=False, vector_partner=None):
+                      ignore_connections=False, vector_partner=None,
+                      boundary_discontinuity=None):
+        if not isinstance(da, xr.DataArray):
+            raise RuntimeError('Input `da` needs to be a xarray.DataArray. \
+                                Is:%s' % (type(da)))
```
rabernat: same

xgcm/grid.py
```diff
@@ -352,43 +354,74 @@ def face_edge_data(fnum, face_axis, count=1):
...
+            boundary=boundary,
+            fill_value=fill_value,
+            ignore_connections=ignore_connections,
+            vector_partner=vector_partner,
+            boundary_discontinuity=boundary_discontinuity
+        )
+        edge_data = self._get_edge_data(da, **kw)
+        if not isinstance(edge_data, type(da.data)):
+            raise RuntimeError('Mixed data type for edge (%s) and data(%s)'
+                               % (type(edge_data), type(da.data)))
```
rabernat: same

xgcm/grid.py
```diff
@@ -399,14 +432,19 @@ def _get_neighbor_data_pairs(self, da, position_to, boundary=None,
             raise ValueError("`%s` is not a valid axis position name. Valid "
                              "names are %r." % (position_to, valid_positions))
 
-        if position_to not in self.coords:
-            raise ValueError("This axis doesn't contain a `%s` position"
-                             % position_to)
+        # This prevents the grid generation to work, I added an optional
+        # kwarg that deactivates this check
+        # (only set False from autogenerate/generate_grid_ds)
```
rabernat: I think I understand. 👍

xgcm/test/test_grid.py
```diff
@@ -3,20 +3,107 @@
...
 
+
 # helper function to produce axes from datasets
 def _get_axes(ds):
     all_axes = {ds[c].attrs['axis'] for c in ds.dims if 'axis' in ds[c].attrs}
     axis_objs = {ax: Axis(ds, ax) for ax in all_axes}
     return axis_objs
 
 
+# duck array ops should maybe get its own test module?
```
rabernat: 👍

rabernat: I would be happy if you would factor those out into a new test module.

xgcm/test/test_grid.py
```diff
@@ -3,20 +3,107 @@
...
+    else:
+        ref = discontinuity
+
+    kw = {'boundary_discontinuity': discontinuity}
+    if right:
+        extended_raw = axis._extend_right(ds.XC, **kw)
+        extended = extended_raw[-1]
+        expected = ds.XC.data[0] + ref
+        with pytest.raises(RuntimeError):
+            axis._extend_right(ds.XC.data, **kw)
```
rabernat: Maybe `ValueError` would be more appropriate to raise here? `RuntimeError` is a sort of vague catchall:
https://docs.python.org/2/library/exceptions.html

xgcm/test/test_grid.py
```diff
@@ -3,20 +3,107 @@
...
+
+
+@pytest.mark.parametrize('fill_value', [0, 10, 20])
+@pytest.mark.parametrize('boundary', ['fill', 'extend', 'extrapolate'])
+@pytest.mark.parametrize('periodic', [True, False])
+@pytest.mark.parametrize('is_left_edge', [True, False])
+@pytest.mark.parametrize('boundary_discontinuity', [None, 360])
+def test_get_edge_data(periodic, fill_value,
+                       boundary, is_left_edge,
+                       boundary_discontinuity):
```
rabernat: It's great to have these tests. Thanks!

xgcm/test/test_grid.py
jbusecke: I deleted this since I dont test the types anymore (as you suggested).

xgcm/test/test_grid.py
jbusecke: Done

xgcm/test/test_grid.py
jbusecke: I think it would be helpful to run those tests on a small 'tiled' dataset aswell, but I think you could do that much more efficiently. I am still trying to understand the get edge logic in detail!

rabernat:
Ok I think this looks good to go! Thanks so much Julius!

Issue #81, opened 2017-11-01
Support for grid metrics


jbusecke:
Ok now that #74 is merged 🎆 🍾 , I would like to talk about the next step: Recalculating distances from given coordinates.

This is much easier to implement, I believe. However, I think it is worth talking about the actual calculation in detail.
The case worth discussing is the calculation of physical [meters] from positions on the globe, since distances for e.g. depth (already in meters) are now easy to implement using the xgcm.Grid/Axis diff function.

I put together a simple function to calculate the distance in meters from dlon/dlat (longitude/latitude differentials in deg) and the actual longitude/latitude.

```python
    def dll_dist(dlon, dlat, lon, lat):
        """Converts lat/lon differentials into distances
   
        PARAMETERS
        ----------
        dlon : xarray.DataArray longitude differentials
        dlat : xarray.DataArray latitude differentials
        lon  : xarray.DataArray longitude values
        lat  : xarray.DataArray latitude values
   
        RETURNS
        -------
        dx  : xarray.DataArray distance inferred from dlon
        dy  : xarray.DataArray distance inferred from dlat
        """
   
        # First attempt with super cheap approach...
        #     111km for each deg lat and then scale that by cos(lat) for lon
        dll_factor = 111000.0
        dx = dlon * np.cos(np.deg2rad(lat.data)) * dll_factor
        dy = dlat * dll_factor
   
        return dx, dy
```
As @rabernat mentioned in #74 this assumes a projection already (Mercator I guess?). Is there a sleek way to do this with existing packages like cartopy?

rabernat:
Before we get to this issue, I think it is important to think about how xgcm will interact with these "[metrics](https://en.wikipedia.org/wiki/Metric_(mathematics))" (seems like a good term) in more generality. Many datasets already come with them defined. How do the metrics related to `Grid` an `Axis` objects?

The metrics define the distance between axis. Inspired by MITgcm, I imagine metrics `dxc` and `dxg` associated with an `Axis` in the following way:

```
|------o-------|------o-------|------o-------|------o-------|
<-----dxg------>
       <-----dxc------>
```

However, one could make this higher order by doing
```
|------o-------|------o-------|------o-------|------o-------|
 <-dxl-><-dxr->
```
(I believe this is the approach used by MOM.)

By associating each `Axis` with the appropriate metrics, we can turn our difference operation into a derivative, and our sum / cumsum into an integration. (The metrics also open the door to more accurate interpolation methods, such as those used in MOM.)

[...]

PR #99, opened 2018-06-08
Metrics and initial calculus functions

rabernat:
I had no wifi on the train, so I took a first step towards addressing #81, the long-desired support for higher order calculus functions.

I just wrote a test for what this might look like. @jbusecke please let me know what you think of this API in terms of specifying the metrics and taking the derivatives.


xgcm/test/test_calculus.py
```diff
@@ -0,0 +1,23 @@
...
+    # a non-uniform grid would provide a more rigorous test
+    dx = 10.
+    ds = xr.Dataset({'foo': (('XC',), [1., 2., 4., 3.])},
+                    coords={'XC': (('XC',), [0.5, 1.5, 2.5, 3.5]),
+                            'XG': (('XG',), [0, 1., 2., 3.]),
+                            'dXC': (('XC',), [dx, dx, dx, dx]),
+                            'dXG': (('XG',), [dx, dx, dx, dx])})
+
+    grid = Grid(ds, coords={'X': {'center': 'XC', 'left': 'XG'}},
+                metrics={'XC': 'dXC', 'XG': 'dXG'},
```
jbusecke: Ughhhh talking about wifi...I thought I had it but it seems like it wasnt sufficient to upload my comment. 
So here it goes again: I am basically wondering how this `metrics=` statement would look like with multiple dimensions.
I think it might be better to define only the 'type' of metric, e.g. area or xdistance etc, and let xgcm figure out to which set of dims this belongs.
Something like this:
```
...
metrics={'dxc':'dx', 'dxg':'dx'}
...
```
We know that dxc is located at the cell face (from its dimensions), so we can make the input a little less cluttered if we move to more dimensions.

Internally we might parse this info into a dict structure like I mentioned in #81:
For this particular case:
`{['XC']:{'dx':'dXC'}}` and `{['XG']:{'dx':'dXG'}}`, which is not much different from your approach above.
If you add a Y dimension it would look like this:
`{['XC', 'YC]:{'dx':'dXC', 'dy':'dYC', 'area':'rA'}, {['XG', 'YG]:{'dx':'dXG', 'dy':'dYG'}, ...}` and so forth for all unique combinations of dimensions.

Looking at this now it seems quite complicated, but i think this way we future proof the code to include other metrics that depend on more than one dimension (area and volume). 

eventually I imagine something like this:

```
...
metrics={['dXC', 'dXG']:'dx', ['dYC', 'dYG']:'dy', ['rAC']:'area'}
...
```
with many metrics to define this would be significantly shorter.

What do you think?

rabernat: My approach is to start with the simplest case and then build on that. I agree that we want multidimensional metrics. We should use a set as the key since its not order dependent.

What I don't like about your proposed syntax is that it involves special keywords like `dy` and `area`. What if we did this instead:
```python
metrics={'XG': 'dXC',
         'YG': 'dYG', 
         {'XG', 'YG'}: 'rAC'}
```

jbusecke: I see your point. I see one issue with this (again this is more thinking about the future, but it might save some time down the road):

If I understand the logic above, you are assuming that each unique combo of dimensions has only one metric, correct? How does xgcm know from the above that `'dXC'` is a distance in the x-direction?

I am thinking this might lead to problems when we want to implement things like a finite-volume divergence, for which we need the cross-sectional length (`'dYG'`).
I dont see how that would be passed with your above syntax without needing additonal information to distinguish between `'dXC'` and `'dYG'` which are both located at the same dimensions.

My rational behind the special keywords is that we know that each point has 3 distances associated with it, each combination of two points has an area associated, and each combination of 3 points has a volume associated with it. This would be a complete description of the grid, If I am not missing something here.

This approach disables one step of association: If I have both 'dxC' and 'dxG', from the label I know that it is an x-distance, xgcm can figure out where in the grid it is located, by its dimensions. 
I find myself checking all the time which naming convention is used for tracer vs. cell values, but the axis is usually pretty clear (e.g. usually some form of '*dx*').

Does this make any sense?

jbusecke: Actually we could use something like `'X': 'dXC'` and `{'X', 'Y'}: 'rAC'}` as a compromise.
This links each metric to an axis (already part of the xgcm logic), and the numer of axes specifiers passed determines whether it is a distance (1 axis parameter), an area (2 axis parameter)....

Perhaps the best of both worlds?

jbusecke:
I am eager to work further on this. Should I submit my own PR, or modify this one here?

jbusecke:
I have been thinking and tinkering with this approach a fair bit recently.
I have been working with my own module [here](https://github.com/jbusecke/xarrayutils/blob/master/xarrayutils/xgcm_utils.py).

These are all based on passing a list of metrics, e.g. distances along the x-axis and then chosing the correct one based on the dimensions. I think this would simplify the input of metrics.

I propose a slightly different syntax for ingesting the metrics:

```
grid = Grid(..., metrics={
    ‘distance’:{
        ‘X’: (‘dXC’, ‘dXG’, ...),
        ‘Y’: (‘dYC’, ‘dYG’),
        ...}
    ‘area’:{
        ‘X’: (‘cellfaceU’, ...) #area of cellface perpendicular to X, e.g. for zonal fluxes
        ‘Z’: (‘areaC’, ‘areaU’, ‘areaV’) # horizontal area
        ...})
```

This is still verbose but it gets rid of the dataset specific dimension names as dictionary keys and uses the more intuitive axis notation.

For each call we can them check which metric matches the dataarray passed (e.g. find the dx distance matching the u input) and use it to calculate derivatives etc.

We could later implement some optional “autogenerate” logic. For instance if we are missing the cell face area at the u-point but have the distance along x and z at that point we can create the cell face by multiplication.

Do you have any issues with this approach @rabernat? If so we will have to fight about it over lunch 😜.

rabernat:
I'm not sure why codecov is not happy, but I'm gonna merge.

Issue #86, opened 2017-11-09
more typos in autogenerate docs


rabernat:
I wish I had caught all these before merging #85, but I am eager to move fast.

Here are the rendered versions on rtd:
http://xgcm.readthedocs.io/en/latest/grid_generation.html
http://xgcm.readthedocs.io/en/latest/api.html#autogenerate

I recommend the following

- [ ] link explicitly to the API docs from your grid_generation doc page. (You can see how to do this in the other doc pages.)
- [ ] re-run your example from grid_generation.md to make the variable names consistent with the new naming conventions
- [ ] fix the typos in the module docstrings

Not urgent, just wanted to document this via an issue.

cc @jbusecke

jbusecke:
I need to update the docs still, but thanks for the fast merge on #96. I really needed that functionality back!

rabernat:
Will you add an example notebook working with real observational data?

jbusecke:
Yes that was the plan. Now that the generation of the 'outer' position works, it is easy to calculate grid distances. That is what I use this module for most and ill add a nice example, emphasizing the effect of proper area weighting like here (http://xarray.pydata.org/en/stable/examples/monthly-means.html) and maye a gradient calculation? Any other ideas that would be helpful to demonstrate?

Can you assign me to this issue? Its easier to keep track of my todos that way.

rabernat:
For some reason I can't assign you.

jbusecke:
Should we integrate this into a larger "Improved docs" Milestone?

jbusecke:
This was covered by #106. Closing

Issue #77, opened 2017-11-01
new autogenerate functions are missing from API docs


rabernat:
http://xgcm.readthedocs.io/en/latest/api.html

This can be fixed with a few lines here:
https://github.com/xgcm/xgcm/blob/master/doc/api.rst

cc: @jbusecke

jbusecke:
Just added the lines to #78

PR #106, opened 2018-06-29
Jbusecke/autogenerate docs

jbusecke:
Here is a more extensive documentation of the autogenerate module.

This addresses #77 and #86.

Is there a better way to link from a notebook to the API section of the documentation? When I render it locally it looks quite ugly with the full link.
screen shot 2018-06-29 at 12 28 22 pm

This is the code in my notebook:

```
The [`autogenerate`](http://xgcm.readthedocs.io/en/latest/api.html#module-xgcm.autogenerate) module
```

@rabernat, with regards to #86, I think the latest improvements of the automatic padding (e.g. extrapolation) implemented in #96 make the default settings much more useful. Could you give it a test run perhaps?


willirath:
> Is there a better way to link from a notebook to the API section of the documentation? When I render it locally it looks quite ugly with the full link.

Looks fine for me on a local Jupyterlab, Jupyter-Notebook and also on Github.

jbusecke:
Oh another thing: I added a dataset for my global example. I guess we would want to avoid bloating the repo with netcdf files. I was not able to find a global example in the xarray tutorial datasets. Should we follow the same strategy as xarray in the long-term, and outsource the actual data files?

rabernat:
> Is there a better way to link from a notebook to the API section of the documentation?

I found this page in the nbsphinx docs:
https://nbsphinx.readthedocs.io/en/0.3.4/markdown-cells.html#Links-to-Domain-Objects
It explains how to use autodoc links within the notebook. Is that what you have done? In any case, that's what I recommend.

> I guess we would want to avoid bloating the repo with netcdf files. I was not able to find a global example in the xarray tutorial datasets. Should we follow the same strategy as xarray in the long-term, and outsource the actual data files?

This is a very good point. I'm not sure the best way to solve it though. I once recall finding a package that managed the test datasets automatically. But I can't track it down now. Let's not let that hold things up.

Overall, I really like you example and think it will make a great addition to the docs!

Once you have decided you are happy with the linking issue, I will merge this.

jbusecke:
Ook. I removed the old autogenerate docs, fixed some typos and implemented the API links.

PR #130, opened 2019-05-19
add integrate function

jbusecke:
Adds the ability to use the new metrics functionality (#99) to integrate fields using `grid.integrate(da, axis)` syntax. 

This PR also includes substantial changes in the metrics 'internals'. This was necessary due to a bug in #99, where the `Grid._assign_metric` function did not assign a list that was passed. Instead onlyt the last list item was assigned. 

This PR adds more substantial tests for both internal `Grid` metric handling as well as the integrate functions.


jbusecke:
There are a lot of changes that are just due to my `black` code formatter...Sorry if this causes confusion. Maybe I should open a seperate PR, which reformats the whole code according to black and we rebase this one on top of that?

jbusecke:
Ok I rebased this PR onto the lastest `master`. It should be much easier to see changes now.

jbusecke:
Any thought on this one @rabernat?

jbusecke:
@rabernat I just wanted to ping this again. Since I made some changes to your metric logic, I would love if you could have a look at it. There is probably also a way to achieve this fix more elegantly.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

xgcm/grid.py
```diff
@@ -1111,6 +1152,36 @@ def derivative(self, da, axis, **kwargs):
         dx = self.get_metric(diff, ("X",))
         return diff / dx
 
+    @docstrings.dedent
+    def integrate(self, da, axis, **kwargs):
+        """
+        Take the centered-difference derivative along specified axis.
```
jbusecke: This needs to be updated.

rabernat:
I found a bug in the metrics code, fixed in #136. It was not working properly when you tried to have more than one metric defined per axis (e.g. `{('X'): ['dxC', 'dxG']}`).

Once that is merge you should try to rebase this.

Sorry for the slow responses and bugs.

xgcm/grid.py
```diff
@@ -1111,6 +1152,36 @@ def derivative(self, da, axis, **kwargs):
...
+        Parameters
+        ----------
+        axis : str, list of str
+            Name of the axis on which to act
+        %(neighbor_binary_func.parameters.no_f)s
+
+        Returns
+        -------
+        da_i : xarray.DataArray
+            The integrateddata
```
rabernat: typo

xgcm/grid.py
```diff
@@ -1111,6 +1152,36 @@ def derivative(self, da, axis, **kwargs):
...
+        """
+        # convert axis to list if passed as str
+        if isinstance(axis, str):
+            axis = [axis]
+        weight = self.get_metric(da, set(axis))
```
rabernat: don't need to use a set here

xgcm/grid.py
```diff
@@ -1111,6 +1152,36 @@ def derivative(self, da, axis, **kwargs):
         dx = self.get_metric(diff, ("X",))
         return diff / dx
 
+    @docstrings.dedent
+    def integrate(self, da, axis, **kwargs):
```
rabernat: what if I want to integrate over multiple dimensions at once, e.g. ('X', 'Y')

xgcm/test/test_integrate.py
```diff
@@ -0,0 +1,355 @@
...
+                dy_n,
+                dy_e,
+                dy_t,
+                dz_t,
+                dz_w,
+                area_ne,
+                area_n,
+                area_e,
+                area_t,
+                volume_t,
```
rabernat: I know this is how black wants it, but I can't stand this formatting. Is there some config we can tweak to change it?

jbusecke: Their argument is that is produces nicer diffs. And there are no ways to tweak black, thats pretty much the idea behind it I think.
>Black reformats entire files in place. It is not configurable. It doesn't take previous formatting into account. It doesn't reformat blocks that start with # fmt: off and end with # fmt: on. # fmt: on/off have to be on the same level of indentation. It also recognizes YAPF's block comments to the same effect, as a courtesy for straddling code.

Is this such a concern that we should revert to another formatter? 
FWIW, I really like it since its enables easy commenting of input arguments, but I can see how it has a weird asthetic.

rabernat: Ok fine, no problem.

xgcm/grid.py
jbusecke: That works, and is tested in `test_integrate()` in the test module. I however only use lists like `['X','Y']`. Should I add tuples?

xgcm/test/test_integrate.py
```diff
@@ -0,0 +1,355 @@
...
+        "B": ds_b,
+        "C": ds_c,
+        # "fail_gridtype": ds_fail,
+        # "fail_dimtype": ds_fail2,
+        "coords": coords,
+        "metrics": metrics,
+        # "fail_coords": coords_outer,
+    }
+
+
```
jbusecke: I spent quit a bit of time on these tests. Could you move them into #136? That would confirm that all bugs mentioned here are 100% solved by that commit. Then I can remove all the metric code from this PR.

xgcm/test/test_integrate.py
```diff
@@ -0,0 +1,355 @@
...
+        with pytest.raises(expected_error):
+            metric = grid.get_metric(ds[data_var], axes)
+    else:
+        metric = grid.get_metric(ds[data_var], axes)
+        expected_metrics = [
+            ds[me].reset_coords(drop=True) for me in metric_expected_list
+        ]
+        expected = functools.reduce(operator.mul, expected_metrics, 1)
+        assert metric.equals(expected)
+
```
jbusecke: Up until here stuff should probably go to #136

xgcm/test/test_integrate.py
jbusecke: For this very long list I could use a list of tuples directly (instead of zipping), which would cut this thing in half at least?

xgcm/test/test_integrate.py
rabernat: The problem with these tests is that they test private methods (i.e. you use the private method `._assign_metrics` and check agains the `._metrics` attribute). I realize that other parts of xgcm do this, but I have since realized it is very bad practice. We should test the user-facing functions. The private methods are implementation details, and we should feel free to change them without breaking tests.

I will try to refactor your tests to keep the functionality without doing this.

jbusecke: Ah ok I wasnt aware of that. I just wanted to ensure that everything works on the most basic level.

jbusecke: Curious to see how to do it properly.

jbusecke:
Ok this works nicely on a datarray now. Do you think it would be desirable to have this working on a full dataset with different variables? I think that would be very cool. 
I would implement it like this maybe:
```
def integrate(da, ...):
    if isinstance(da, xr.dataarray):
        # do what we do now and return the weighted array
    elif isinstance(da, xr.dataset):
        return xr.concat({k, integrate(v, ...) for k,v in da.items()}
```

or can you think of a more elegant way?

rabernat:
> Do you think it would be desirable to have this working on a full dataset with different variables?

No. Let's make small changes, one at a time.

jbusecke:
Sounds good. Do you think this is ready to merge?

xgcm/test/test_integrate.py
rabernat: This is now very redundant with the dataset generator in test_metrics.

Any chance you could factor these out into a standalone module in the tests directory?

jbusecke:
How is this?

xgcm/grid.py
```diff
@@ -1145,6 +1145,33 @@ def derivative(self, da, axis, **kwargs):
...
+            Name of the axis on which to act
+        %(neighbor_binary_func.parameters.no_f)s
+
+        Returns
+        -------
+        da_i : xarray.DataArray
+            The integrated data
+        """
+
+        weight = self.get_metric(da, _maybe_promote_str_to_list(axis))
```
rabernat: you don't need to call _maybe_promote_str_to_list here. It is done inside get_metric.

jbusecke:
Do you think this is good to go @rabernat?

xgcm/grid.py
```diff
@@ -1145,6 +1145,33 @@ def derivative(self, da, axis, **kwargs):
         dx = self.get_metric(diff, ("X",))
         return diff / dx
 
+    @docstrings.dedent
+    def integrate(self, da, axis, **kwargs):
+        """
+        Integrate along specified axis or axes.
+        Grid metrics (e.g. cell length, area, volume) are determined automatically.
```
rabernat: What about "Perform finite volume integrate along specified axis or axes, accounting for grid metrics. (e.g. cell length, area, volume)"

xgcm/grid.py
```diff
@@ -1145,6 +1145,33 @@ def derivative(self, da, axis, **kwargs):
...
+            The integrated data
+        """
+
+        weight = self.get_metric(da, axis)
+        weighted = da * weight
+        # TODO: We should integrate xarray.weighted once available.
+
+        dim = [
+            di for ax in axis for di in self.axes[ax].coords.values() if di in da.dims
+        ]
```
rabernat: What happens if I try to integrate over an axis that is not present in the array? Like `grid.integrate(da, ('X', 'Y, 'Z'))` where da is 2D?

xgcm/test/test_integrate.py
```diff
@@ -0,0 +1,71 @@
...
+            assert integrated.equals(expected)
+    elif gridtype == "C":
+        for axis, metric_name, dim in zip(
+            ["X", "Y", ["X", "Y"]],
+            ["dx_n", "dy_n", "area_n"],  # need more metrics?
+            ["xt", "yu", ["xt", "yu"]],
+        ):
+            integrated = grid.integrate(ds.v, axis)
+            expected = (ds.v * ds[metric_name]).sum(dim)
+            assert integrated.equals(expected)
```
rabernat: Sorry to be nit-picky here, but the style of these tests is important to me...

I would prefer to just have two separate tests, test_integrate_cgrid, test_integrate_bgrid, rather than this internal if / else logic.

xgcm/test/datasets.py
```diff
@@ -238,3 +238,155 @@ def periodic_2d(request):
 def nonperiodic_2d(request):
     ds, periodic = datasets_with_periodicity[request.param]
     return ds, periodic, expected_values[request.param]
+
+
+def datasets_grid_metric():
```
rabernat: what about `def datasets_grid_metric(grid_type='C')`?

xgcm/test/datasets.py
```diff
@@ -238,3 +238,155 @@ def periodic_2d(request):
...
+            {"u": u_b, "v": v_b, "wt": wt, "tracer": tr, "timeseries": timeseries}
+        )
+    )
+    ds_c = _add_metrics(
+        xr.Dataset(
+            {"u": u_c, "v": v_c, "wt": wt, "tracer": tr, "timeseries": timeseries}
+        )
+    )
+
+    return {"B": ds_b, "C": ds_c, "coords": coords, "metrics": metrics}
```
rabernat: I would prefer to just return one dataset at a time.

xgcm/test/test_integrate.py
```diff
@@ -0,0 +1,71 @@
...
+        ["xt", "yt", "zt", ["xt", "yt"], ["xt", "yt", "zt"]],
+    ):
+        integrated = grid.integrate(ds.tracer, axis)
+        expected = (ds.tracer * ds[metric_name]).sum(dim)
+        assert integrated.equals(expected)
+        # test with tuple input if list is provided
+        if isinstance(axis, list):
+            integrated = grid.integrate(ds.tracer, tuple(axis))
+            expected = (ds.tracer * ds[metric_name]).sum(dim)
+            assert integrated.equals(expected)
```
rabernat: This would be a place to test for the error of integrating over a dataset that doesn't have the requested axes.

rabernat:
Julius, thanks for your patience. This is almost there.

I have a few suggestions for how to restructure the tests, and for how to deal with the arguments to integrate.

xgcm/test/test_integrate.py
jbusecke: Thats a good point. Ill get onto that soon.

xgcm/test/datasets.py
jbusecke: Could we do a separate PR for this? #132 rebased on this style and I think it is important to fix that bug. I promise Ill work swiftly on a PR that makes this dataset routine nicer. Then I can also add a non-uniform case.

xgcm/test/datasets.py
jbusecke: see below

xgcm/test/test_integrate.py
jbusecke: Roger that.

xgcm/grid.py
jbusecke: intergration? I think that sounds good.

xgcm/test/test_integrate.py
jbusecke: I am implementing the changes but just out of curiosity? Is this so the pytest error messages are more specific? Just trying to learn along the way :-)

rabernat: I just don't like lots of if / else blocks in parameterized tests. Parameterization makes sense if you can run basically the same test on different inputs. But if you have to have big if / else blocks to deal with the different inputs, you don't gain anything by parameterizing other than unnecessary complexity. At that point you just have two separate functions already.

This is a stylistic choice, and there is no right and wrong here. But xgcm is my 👶and I just have these opinions.

xgcm/test/test_integrate.py
```diff
@@ -5,22 +5,26 @@
...
-    ds = ds_full[gridtype]
-    grid = Grid(
-        ds,
-        coords=ds_full["coords"],
-        metrics=ds_full["metrics"],
-        # periodic=True,
-    )
+    ds = ds_full[grid_type]
+    return ds, ds_full["coords"], ds_full["metrics"]
+
```
jbusecke: This is just me being lazy and not refactoring the metric dataset just yet. Mostly I want to speed this up in order to get #132 merged asap.

xgcm/test/test_integrate.py
```diff
@@ -5,22 +5,26 @@
...
     for axis, metric_name, dim in zip(
         ["X", "Y", "Z", ["X", "Y"], ["X", "Y", "Z"]],
         ["dx_t", "dy_t", "dz_t", "area_t", "volume_t"],
         ["xt", "yt", "zt", ["xt", "yt"], ["xt", "yt", "zt"]],
     ):
+        print(axis)
+        print(ds.tracer)
+        print(metrics)
+        print(grid.get_metric(ds.tracer, axis))
         integrated = grid.integrate(ds.tracer, axis)
```
jbusecke: oops let me erase those print statements

xgcm/test/test_integrate.py
```diff
@@ -29,43 +33,105 @@ def test_integrate(gridtype):
...
+        ["dx_n", "dy_n", "area_n"],  # need more metrics?
+        ["xt", "yu", ["xt", "yu"]],
+    ):
+        integrated = grid.integrate(ds.v, axis)
+        expected = (ds.v * ds[metric_name]).sum(dim)
+        assert integrated.equals(expected)
+
+
+@pytest.mark.parametrize("axis", ["X", "Y", "Z"])
+def test_integrate_missingaxis(axis):
```
jbusecke: So I am testing two error cases: 
1) An axis specified in `integrate` missing in the grid object
2) A missing dimension  in the input dataarray (e.g. because it was sliced or averaged)

Can you think of anything else we want to catch here?

xgcm/test/test_integrate.py
jbusecke: Done in the newest push.

xgcm/test/test_integrate.py
jbusecke: No thats fair. And it does make sense.

xgcm/test/test_integrate.py
rabernat: Nothing else comes to mind.

jbusecke:
Ok I rebased this on master. Is the revised testing ok in your eyes @rabernat?

rabernat:
Fantastic! This is awesome. I just had a quick look through overall it looks awesome. But we need narrative documentation. I recommend writing a new page on "Metrics and Integrals".

jbusecke:
Thanks! Ill put something together hopefully later today. I feel a schematic would help tremendously. Is there anything online so we do not have to build it ourselves?

rabernat:
> I feel a schematic would help tremendously. Is there anything online so we do not have to build it ourselves?

I'm not sure what sort of schematic you have in mind, so hard to offer advice. I have been loving https://www.lucidchart.com and now have a paid account. That's what I would use. It would be nice to redo all our schematics there.

jbusecke:
Ill try to come up with some sketches.

———————————————
juliusbusecke.com

jbusecke:
Ok so I have added some docs. 

A couple of things:

1. The build from the upstream master is breaking now, haven't looked into why.
2. Should we request [this](https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html) so PRs can be built in rtd before merging.

jbusecke:
Ok got it. Apparently xarray changed the way they handle replacing coordinates in a dataset. 

So this:
```
    ds_chopped = ds.copy()
    del ds_chopped["data_g"]
    ds_chopped["XG"] = ds_chopped["XG"].isel(XG=slice(None, 3))
    print(ds_chopped)
```
end up as a datarray with a nan padded `XG`

```
Dimensions:  (XC: 100, XG: 100)
Coordinates:
  * XG       (XG) float64 0.0 0.06283 0.1257 nan nan nan ... nan nan nan nan nan
  * XC       (XC) float64 0.03142 0.09425 0.1571 0.2199 ... 6.126 6.189 6.252
Data variables:
    data_c   (XC) float64 0.9834 0.802 0.3393 0.627 ... 0.3024 0.1027 0.08346
```

By selecting the dimension and then dropping the datavariable. This works in all cases.

dcherian:
Please open a bug report. We just merged a big PR refactoring `merge` that may have changed behaviour

jbusecke:
Will do later today.

jbusecke:
@rabernat, is there a way to preview the rtd output for a PR? I rendered it locally but would love to check if it looks ok in the final form. 
Also I fixed #158, which was caused by a typo in the `index.rst`.

jbusecke:
Any comments on the added docs? If so I could finish them up tonight. Eager to move ahead.

rabernat:
Docs are great! I say let's merge and move on!

PR #131, opened 2019-05-22
black formatting

jbusecke:
I am a big fan of [black](https://github.com/python/black) as a code formatter. This PR is intended to switch xgcms code style to black. 

This is currently a work in progress, where I just ran `black .` on the root directory and added a badge 😁 
I still need to find a way to force black formatting in the CI process.


jbusecke:
Ok the reformatted code went through and I think the latest commit will enable 'checking' the code with black. This will error out the build if something is not black conform. 
@rabernat I am also thinking of setting up a pre-commit hook that auto formats code according to [this](https://www.mattlayman.com/blog/2018/python-code-black/). Thoughts?

jbusecke:
Ok so the way I have it set up now is that if `pre-commit` is installed the code will be checked with black before pushing. Should I add `pre-commit` to the py36 CI environment? Should we make a note in the docs/readme that we enforce black style?

jbusecke:
I am not sure why codecov complains but the reformatting might have just upset the diff calculation. @rabernat let me know if you think this is good to go, so that I can start rebasing #130.

jbusecke:
Do you think we should add pre-commit to one of the environments? Or should we just add some text that the user can use this to auto format commits before sending them out?

I assume the default `dev` environment should be `ci/py36` right?

jbusecke:
Can we move ahead and merge this? I would love to rebase #130 and get @rabernat opinion on that.

rabernat:
Let's try it and see how it goes! :)

PR #132, opened 2019-07-04
Small typo in grid.derivative

timothyas:
In `grid.derivative`, the call to `self.get_metric` had a hard coded axis argument. This simply changes that to whatever the user's input axis is.


jbusecke:
Thanks for catching this! The metrics implementation has still a few bugs (see #130), but I did not see this one. 

The failing CI build seems to be related to a deprecated `message` parameter in pytest (see [here](https://docs.pytest.org/en/4.6-maintenance/deprecations.html#message-parameter-of-pytest-raises)). Apparently this does not check the error message. 
I'll try to fix that in a seperate PR. 
@rabernat, what do you think it the best order of fixing these?

I think:
1) Check and merge #130 
2) Change the error handling in `xgcm/test/test_grid.py` ([e.g. here and in following lines](https://github.com/xgcm/xgcm/blob/f9b58ff986da3cddc99044f6a3b576a467d09b94/xgcm/test/test_grid.py#L523)). I will put up a PR shortly.
3) Rebase #132 (with added tests)

xgcm/grid.py
```diff
@@ -1108,7 +1108,7 @@ def derivative(self, da, axis, **kwargs):
 
         ax = self.axes[axis]
         diff = ax.diff(da, **kwargs)
-        dx = self.get_metric(diff, ("X",))
+        dx = self.get_metric(diff, (axis,))
```
jbusecke: Nice catch. Could you add a test for the derivative function in your PR?

timothyas: Yep! I made a quick addition to the uniform test case: just adding another dimension. It was a bit rushed and not the most clever addition, but tests what this PR fixes.

timothyas:
For what it's worth, that sounds good to me! To give you some feedback @jbusecke, I found #130 really useful - it cleans up some things with the metrics stuff so that `grid.derivative` was working fine with MITgcm output.

timothyas:
PS this is funny timing, I will take off for a few days. I just wanted to say that in case anything happens with this PR. I will be back on it in ~1 week and happy to make any further changes! Thanks for the review @jbusecke.

xgcm/grid.py
jbusecke: To me this looks fine. I will put some more work into these parts soon and might add some more cases, but for now this is great. We recently started using [black](https://github.com/python/black) as a code formatter. It seems to be upset with some of the formatting of your PR. This can usually be fixed by running `black` over the file you changed.

jbusecke:
Glad to hear that. Thanks for testing this and helping to weed out the kinks. Ill wait for some feedback from @rabernat (who might be on vacation right now...) and then we can get moving on this.

jbusecke:
That sound good. Enjoy the time off. Ill ping you when things start moving on the other PRs.

jbusecke:
Ok the backlog has been pretty much cleared. This should be easy to rebase. @timothyas are you able to do this some time soon? 

cc'ing @rabernat

xgcm/grid.py
timothyas: See latest commit, black had no problem with grid.py but fixed the `test_derivatives` formatting.

timothyas:
Thanks @jbusecke, it was no problem to rebase this PR so it should be good to go. As mentioned in my review comment, I ran black and it fixed the `test_derivatives` formatting (very nice package, super easy to use!). 

I have a to-do list of coding stuff and making a non-uniform grid test case is on it for a future PR. I'll get to it when I can, but for now let me know if you see any trouble with this one.

xgcm/test/test_derivatives.py
```diff
@@ -8,23 +8,43 @@ def test_derivative_uniform_grid():
     # this is a uniform grid
     # a non-uniform grid would provide a more rigorous test
```
jbusecke: I have refactored a similar one in `tests/datasets/datasets_metric`, could you see if that works for this test? That way we can change that to a nonuniform spacing and have it update in all tests.

timothyas: Nice! I was able to use that, see the latest push for a C grid derivative tests. A couple of comments on this: 
1. Could easily also add a B grid test if you think it's necessary
2. For some reason, the `.equals( )` always returns `False` for me, even when the absolute value of the difference between `expected` and the test derivative is zero. This may be my version of pytest, or something, but I'd be interested if you understand why or see something I'm missing. 
3. There is no Z axis derivative for the velocity points. I even tried to add a `dz_e` living at `(xu, yt, zw)`, and similarly for `dz_y`, but xgcm couldn't figure out the metric to use. Is this because we would need another axis, like 'ZE' or something in addition to 'Z'? I'm thinking this is because there is no defined xgcm axis for these points on the grid.

jbusecke: 1) I think a B-grid test would be great. Could you add that? It seems to me that `test_derivative_uniform_grid():` is now pretty similar to the other test? Perhaps you can delete that test and include both c and b grid test in a `test_derivative` function? @rabernat what is your style opinion here?
2) Weird. Could it be that the coordinates/attributes are somehow different? Does [this](https://github.com/xgcm/xgcm/blob/cac66afeb2426546b142aef07b7752bd02ea0ebb/xgcm/test/test_metrics.py#L34) make it pass?
3) Could you explain a bit more (or share a notebook or code) how you added those metrics? I am surprised that it didnt handle it, and we should definitely find out why.

jbusecke: [This](https://github.com/xgcm/xgcm/pull/130/files/415e89f0a9c69ae39666910de14b26f4985f11ae#r303041847) seems to answer what Ryan thinks about the style of tests. So please disregard 
>Perhaps you can delete that test and include both c and b grid test in a test_derivative function?

timothyas: I couldn't see the link to what you meant about what Ryan's test style preference, but I'm assuming you meant to not delete the old `test_derivative_uniform_grid()` function. Let me know if you meant something else. 

1. B grid added with latest commit. I could have written both B and C grid in a more nested for loop structure, but I think this is nicer because things are more explicit. Let me know if you agree with that.

2. Nice catch! Yeah that did the trick and I made that change. 

3. Coming soon...

timothyas: Re 3: Turns out this is no problem! I just had to draw myself a picture of a finite volume grid to do it right. I ended up changing how the `dz` elements are created, I made them now 1D vectors just like they are in e.g. MITgcm. Let me know if this breaks anything. 

I made that change after doing things more complicated the first go around. I rolled all of that back, but here's a description of one of the intermediate commits just in case you're curious...

b55986e added the `dz` elements which live on each of horizontal velocity points, and while doing that I changed the way all the `dz` elements are referenced in order to be as clear as possible since I think it got a bit confusing. I mentioned it in the comments, but it works like: dz_<1>_<2> where <1> refers to location in horizontal, <2> location in vertical. So the old `dz_t` -> `dz_t_t` and `dz_w` -> `dz_t_w`. After realizing the time dimension was in the `dz` elements, I realized we can just make them 1D vectors, so the additions here don't seem necessary and just make things confusing.

timothyas: Ok it looks like changing `dz` to 1D vectors breaks `test_metrics.test_get_metric()` in the case where 

`axes=('X', 'Y', 'Z'), data_var= 'u' , drop_vars=None, metric_expected_list=['volume_t'], expected_error=KeyError` 

However, in that test `grid.get_metric()` now returns the product of `area_e` and `dz_t`, which is the correct volume element for that grid cell location on a C grid. In my opinion we should change the test because, for example, this is how the vertical grid elements are handled in the MITgcm, so it seems like a legitimate case. 

Sorry this PR is dragging on! Let me know what you think!

timothyas:
Hey @jbusecke (and @rabernat) I just wanted to give this PR a quick refresh for whenever you have the chance to look. To summarize the last comments I made, I see two ways for this PR to incorporate vertical derivative tests for both B and C grids (as far as I can tell!): 
1. Create 3D `dz` fields in `test_metrics.py`, making one 3D field for each possible horizontal location on the grid (tracer, N, E, NE) for the `dz` located vertically in the center of the cell. I did something like this in b55986e but felt it was just overly complicated for what we wanted... 
2. Create 2, 1D `dz` vectors in `test_metrics.py`.

To me, 2 seems like a natural choice because it reflects vertical coordinates in ocean models (this is the state of the PR currently). However, this breaks one of the cases for `test_get_metric`, although I think with a legitimate case that could occur, see [this comment](https://github.com/xgcm/xgcm/pull/132#discussion_r303244414). So to go this route, I would also need to modify the test, for instance by asserting this test returns `area_e`*`dz`, or finding a different error that should be produced. 

I'm hesitant to do that though, since it involves re-writing a test made previous to me. I'm also hesitant about option 1 since it involves adding a lot of extra lines and potentially making these test cases even more difficult to read with all the different vertical coordinate fields. Let me know what you think is best and I'd be happy to finish this one up (and use the functionality :D )

jbusecke:
Dear @timothyas, thank you so much for your patience. Apologies for my long hiatus. I have just refactored the `datasets_grid_metric` in #130 [here](https://github.com/xgcm/xgcm/pull/130/files#diff-07f0b457ea6d9517cc726a3229c9dec3R243). The test dataset has now vertical dimensions too. Could you take a look and see if you think that your tests can work with this? If yes, I hope that #130 can finally be merged after we take care of the bug in #155.

xgcm/test/test_derivatives.py
jbusecke: My intention was to first check if all `combined` metrics (e.g. area, cell_faces, volume) are correctly applied, but you are right: We should also test for datasets where certain combinations are not given as an array but instead as a combo of e.g. area and depth (the logic written by @rabernat, handles that very elegantly). As in my earlier comment, I think it would be best to expand this on top of #130, by dropping for instance the volume and then asserting that the returned metric is correctly reconstructed.

jbusecke:
Just a quick update: #155 was merged. I think #130 is very close. Pending @rabernat s approval, it could be merged soon and then we can tackle this. Have you had a chance to look if your tests would work with the new version of `dataset_grid_metric` in #130 ?

rabernat:
#130 is good to go code wise, but it needs documentation before we can merge.

jbusecke:
Awesome! Ill work on the docs. @timothyas could you rebase this on #130 and add refactor your tests with those datasets? 
Thanks a lot for your patience and work on this PR.

timothyas:
Hey @jbusecke, thanks for the update. I totally understand, and I also forgot this PR was "in line" behind some others. 

I rebased to your branch and using `dataset_grid_metric` raises the same errors. I'll look into this today and see if I'm just missing something stupid and/or what needs to be done to get all the appropriate vertical derivative checks. I'll get back to you as soon as I can.

jbusecke:
Awesome. I am happy to review once you have a revised version. I am eager to fix this issue.

timothyas:
Ok I went with route 2 described [earlier](https://github.com/xgcm/xgcm/pull/132#issuecomment-532383273). I modified `dz_t` and `dz_w` to be 1D vectors rather than the 4D fields that they were. This made the added C and B grid vertical derivative tests to be a little cleaner, and rather than defining a dz element for each horizontal location on the grid. 

As mentioned, that breaks the `test_get_metric` because `get_metric` no longer errors out since it elegantly reconstructs a volume element. So I just modified the test to drop `dz_t`, and then it errors out as expected. I also added this as a reconstructed case (without dropping `dz_t`) just for kicks... I'm not sure how to get permalinks for a PR, so I'll point you to where these two changes are as comments. Whenever you get the chance, let me know what you think - thanks!

xgcm/test/test_metrics.py
```diff
@@ -283,23 +111,22 @@ def test_assign_metric(key, metric_vars):
         (["X", "Y", "Z"], "tracer", None, ["volume_t"], None),
         (["X"], "u", None, ["dx_e"], None),
         (["X", "Y"], "u", None, ["area_e"], None),
-        (("X", "Y", "Z"), "u", None, ["volume_t"], KeyError),  # This should error out
+        (("X", "Y", "Z"), "u", ["dz_t"], ["volume_t"], KeyError),  # This should error out
```
timothyas: This is the test that I modified. By dropping `dz_t` xgcm can no longer create the volume element at the zonal velocity location and so errors out as expected.

xgcm/test/test_metrics.py
```diff
@@ -283,23 +111,22 @@ def test_assign_metric(key, metric_vars):
...
         # reconstructed cases
         (["X", "Y"], "tracer", ["area_t"], ["dx_t", "dy_t"], None),
         (["X", "Y", "Z"], "tracer", ["volume_t"], ["area_t", "dz_t"], None),
+        (["X", "Y", "Z"], "u", None, ["area_e", "dz_t"], None),
```
timothyas: I just added this as a reconstruct test case, without dropping `dz_t` because why not - sometimes those velocity points are tricky...

jbusecke:
To fix those black issues, you can use `pre-commit` explained [here](https://xgcm.readthedocs.io/en/latest/contributor_guide.html). Somehow this doesnt show up in the docs, Ill raise a seperate issue about that.

xgcm/test/test_metrics.py
jbusecke: This is a bit troubling...why does xgcm reconstruct the volume on the u point from the tracer depth? I think this is a bug.

xgcm/test/test_metrics.py
jbusecke: Same thing here. This should not work! `dz_t` is located at `xt` and `yt`, the vertical data to recostruct the volume for `u` should be located at the same point as `area_e`.

xgcm/test/test_metrics.py
timothyas: Here xgcm doesn't recreate `volume_t`, it creates the product of `area_e` and `dz_t`. What was happening was the test was failing because the expected `KeyError` was not being raised, so I'm dropping `dz_t` to get the expected error as before.

xgcm/test/test_metrics.py
timothyas: So this happens because of how I changed the `dz_t` and `dz_w` elements. Because they are now 1D vectors (just like in e.g. MITgcm) they are not tied to particular horizontal points, and therefore xgcm "reconstructs" volume at the `u` point with the product `area_e` and `dz_t`. Note that `area_e` * `dz_t` is not "equal to" `volume_t`, so if instead of having `area_e`,`dz_t` in the expected returned metrics, the test will error out. However, `area_e` * `dz_t` should be the correct volumetric element at the `u` point because the vertical grid spacing does not change from the tracer point on the horizontal grid to any of the velocity points. 

If you want me to make the `dz` elements so that they are instead 3D fields, defined at specific points on the horizontal grid, then I can do this. I just didn't do this because (1) the notation got pretty hairy and I had to make 6 different dz elements which seemed a bit unnecessary and (2) at least the MITgcm uses 1D vertical coordinates so it seemed like a more realistic case... Basically I made this change because I was able to get the same `KeyError` as you had written in `test_get_metric`, and the same behavior in all the other previous tests, and this seemed like an easier solution than making a bunch of 3D fields for `dz` elements. I'm open to being convinced otherwise though :)

jbusecke: I see. Thanks for the explanation. I must have missed the changes in the `dz_t` (github does not offer an option to compare your PR to another (#130) as far as I know, which makes it difficult to see the exact changes).

I think your test case is very relevant, but I would prefer them to be implemented as another `grid_type`. E.g. `ds = dataset_grid_metric('C_fixed_dz`)` (maybe there is a better naming hehe). 
The reason why I want to test the most general case extensively (3D or actually 4D vertical spacing) is that many models use some sort of variable depth spacing for the cells (MOM5 uses z* coordinates and MOM6 has a completely arbitrary vertical coordinate system) and I believe xgcm can be tremendously useful for these grid layouts. 

We need to take care that our methods are not specific to a single model, but it is a great idea to test special cases as you suggested. 

My suggestion is: 
1) Create a separate dataset with a fixed vertical spacing. This should be achievable within ` dataset_grid_metric` relatively easy.
2) Add a corresponding test function `test_derivative_c_grid_fixed_z`/ `test_get_metric_c_grid_fixed_z` or similarly named.

That way all my test stay intact and we add the important case of 1d fixed vertical cell depth to the tests.

xgcm/test/test_derivatives.py
```diff
@@ -1,30 +1,135 @@
 from __future__ import print_function
 import xarray as xr
 
+import numpy as np
+
 from xgcm.grid import Grid, Axis
+from xgcm.test.datasets import datasets_grid_metric
 
 
 def test_derivative_uniform_grid():
```
jbusecke: Do we still need this test function? Or is it superseded by the ones implemented below? @rabernat

jbusecke:
Thanks for all the hard work @timothyas, I think this will be ready soon.

xgcm/test/test_metrics.py
timothyas: Ok, I think that makes sense to have 3D/4D dz grid elements then. Thanks for all the explanation, I guess I should take another model for a spin (heh) to get a flavor for other grids. (It's always been interesting to me that MITgcm has the z* coordinate system but still the vertical grid is printed as a 1D, static vector). 

I'll be able to do this later today or this weekend, it should not be difficult as you said. Coming soon!

jbusecke: Thats why we work as a team on this! Different perspectives give us a more general approach.

>It's always been interesting to me that MITgcm has the z* coordinate system but still the vertical grid is printed as a 1D, static vector

MOM5 does the same thing but the vertical cell depth is in fact changing in time (particularly near the surface), there is probably an option to output the 'real' `dz` as a diagnostic. It is also possible to reconstruct a z* coordinate from the SSH and ocean depth. I have implemented the reconstruction [here](https://github.com/jbusecke/mom_xtools/blob/928f780d59570d96c629e12d87d08b366ffbb024/mom_xtools/mom_budgets.py#L185) *Disclaimer*: This might be MOM specific. 
There are execellent (although not a short read) explanation on that topic in the [MOM5 docs](https://mom-ocean.github.io/assets/pdfs/MOM5_manual.pdf)

README.rst
```diff
@@ -1,20 +1,7 @@
...
-========= ============== ============================================================================
-Link      Provider       Description
-========= ============== ============================================================================
-|Binder|  mybinder.org   Basic self-contained example
-|PBinder| Pangeo Binder  More complex examples integrated with other Pangeo tools (dask, zarr, etc.)
-========= ============== ============================================================================
-
-Description
------------
+|pypi| |conda forge| |Build Status| |codecov| |docs| |DOI| |license|
```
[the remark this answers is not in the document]
timothyas: I brought this back the manual way, and now seems to be back so I'll go ahead and resolve this one

xgcm/test/test_metrics.py
timothyas: I reverted this one back to how it was before so resolving it...

xgcm/test/test_metrics.py
timothyas: Nice! Yeah this looks MOM specific, but with `xgcm.diff` it would not be difficult to implement for MITgcm diagnostics... 

As for the PR, I just added dz elements for velocity points to the `datasets_grid_metric`. Let me know what you think!

xgcm/test/datasets.py
```diff
@@ -238,3 +238,176 @@ def periodic_2d(request):
...
+    # dz elements at velocity points
+    dz_w_ne = xr.DataArray(
+        data_generator() * dz, coords=[("xu", xu), ("yu", yu), ("time", t), ("zw", zw)]
+    )
+    dz_w_n = xr.DataArray(
+        data_generator() * dz, coords=[("xt", xt), ("yu", yu), ("time", t), ("zw", zw)]
+    )
+    dz_w_e = xr.DataArray(
+        data_generator() * dz, coords=[("xu", xu), ("yt", yt), ("time", t), ("zw", zw)]
+    )
```
timothyas: @jbusecke these are the added `dz` elements. The only difference between your and my version of `datasets.py` is this block and where they are added to metrics below. 

It seems like this should be good to go, and I'll just wait for the other PR's in line to get merged, then I can rebase it to master before merging... but let me know what you think! Thanks for all the rstar references!

jbusecke:
Thank a lot! I will give this a final review once #130 merges, so that it is easier to tell the differences. The error in the CI will also be fixed by #130.

jbusecke:
#130 was just merged. Could you rebase this? I will give it a thorough review asap. Thanks for the patience @timothyas

timothyas:
Rebased to master. Thanks for the reviews and for crushing all the other PR's @jbusecke! Nice work!

xgcm/test/test_derivatives.py
```diff
@@ -1,30 +1,135 @@
...
+
+    ds, coords, metrics = datasets_grid_metric("C")
+    grid = Grid(ds, coords=coords, metrics=metrics)
+
+    # run this for each axis and each field in dataset
+    def test_single_derivative(axis, fld, dx):
+
+        dvar_dx = grid.derivative(fld, axis)
+        expected = grid.diff(fld, axis) / dx
+        abs_diff = np.sum(np.abs(dvar_dx - expected))
```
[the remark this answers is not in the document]
timothyas: Oops, that was from earlier. Before adding `reset_coords` (per your suggestion) I couldn't get `.equals` to work, so I was testing the absolute difference in the values. This is gone now

jbusecke:
@timothyas, would it be possible for you to add some examples to the docs? I would prefer to do that in a separate PR, but since you have dealt a lot with `.differentiate()` now, I think you could do that really quick. 
I would add a few examples after the intro to metrics [here](https://xgcm.readthedocs.io/en/latest/grid_metrics.html#Using-metrics-with-xgcm)

jbusecke:
Oh. Seems like black is complaining...

timothyas:
> @timothyas, would it be possible for you to add some examples to the docs? I would prefer to do that in a separate PR, but since you have dealt a lot with `.differentiate()` now, I think you could do that really quick.
> I would add a few examples after the intro to metrics [here](https://xgcm.readthedocs.io/en/latest/grid_metrics.html#Using-metrics-with-xgcm)

Good call, I can add this today

jbusecke:
Awesome. Thanks! 

Also, love the commit message
image
🤣

timothyas:
Hey @jbusecke, I can just add differentiation to `doc/grid_metrics.ipynb` right after integration. Maybe that's what you were suggesting, do you agree?

jbusecke:
> Hey @jbusecke, I can just add differentiation to `doc/grid_metrics.ipynb` right after integration. Maybe that's what you were suggesting, do you agree?

Exactly!

jbusecke:
But I still prefer to have it in a separate PR. Is that cool?

rabernat:
@jbusecke - go ahead and merge this when you're happy.

Thanks @timothyas!

timothyas:
Ah ok! Yes that makes sense, it's cleaner. No problem!

jbusecke:
🍾

PR #135, opened 2019-07-05
corrected error testing function

jbusecke:
Changes the checking procedure for error messages (see #133).

PR #136, opened 2019-07-10
Fix metric bug

rabernat:
This fixes a bug in related to the internal storing of metrics. In the metric specification, we use something like `metrics={("X",): ["dXC", "dXG"]}`. However, in the internal storage, we did not use a list for the metric variables. This meant that we actually couldn't store multiple metrics for a single axis.

~This still needs a test.~


rabernat:
@jbusecke - it would be great if you could confirm that this solves the same bug you identified in #130. If so, I would prefer to do it this way.

jbusecke:
It is the same bug I noticed, yes. I would like to move my tests from #130 to confirm that this indeed addresses all the issues I experienced. Could you move them into this PR?
 
Also if you could check on #135, that would `repair` the CI which fails for everything right now due to changes in pytest.
 
Once that works and this is merged, Ill start rebasing #130 onto this. Sound good?

jbusecke:
Also thanks for putting this together, this looks of course a lot nicer than my solution haha.

xgcm/test/test_metrics.py
```diff
@@ -44,3 +70,275 @@ def test_metrics_2d_grid():
...
+    }
+
+
+@pytest.mark.parametrize(
+    "key, metric_vars",
+    [
+        # (("X"), ["dx_t"]), # this fails, can we raise a more useful error?
+        (("X",), ["dx_t"]),
+        (("X",), []),  # Should this pass?
+        (("X",), "dx_t"),
```
rabernat: This is the only test that is failing. I require the metrics to be a list. What do you think about that?

jbusecke: For the sake of useability I would prefer to convert any string input to a list internally, but if a proper error is raised I could live without it. Great that the other all pass.

rabernat: What do you suggest is the best way to do this?

xgcm/grid.py
```diff
@@ -905,31 +907,47 @@ def check_neighbor(link, position):
...
         """
         metrics should look like
-           {('X', 'Y'): 'rAC'}
+           {('X', 'Y'): ['rAC']}
         check to make sure everything is a valid dimension
         """
 
         self._metrics = {}
 
         for key, metric_vars in metrics.items():
```
jbusecke: Could you add 
```
if isinstance(metric_vars, str):
    metric_vars = [metric_vars]
    ...
```
here to accept str input for a single metric per axis?

xgcm/test/test_metrics.py
jbusecke: I made a suggestion in the definition of `_assign_metrics`. 

Not sure if this is good practice, but the more I think about it, it would be a nice thing to have. I am again thinking of a very simple use case of e.g. an obervational dataset that has only the tracer cell dimensions. I think `Grid(..., metrics={'X':'dx, 'Y':'dy'})` looks a lot cleaner than `Grid(..., metrics={'X':['dx], 'Y':['dy']})`

xgcm/test/test_metrics.py
```diff
@@ -44,3 +70,273 @@ def test_metrics_2d_grid():
...
+        ("Z",): ["dz_t", "dz_w"],
+        ("X", "Y"): ["area_t", "area_n", "area_e", "area_ne"],
+        ("X", "Y", "Z"): ["volume_t"],
+    }
+    # Used previously for some fail test. Retain for now.
+    # coords_outer = {
+    #     "X": {"center": "xt", "outer": "xu"},
+    #     "Y": {"center": "yt", "outer": "yu"},
+    #     "Z": {"center": "zt", "outer": "zw"},
+    # }
```
rabernat: I would like to remove / resolve all of these commented lines before merging. Can you recommend what to cut / keep?

xgcm/test/test_metrics.py
```diff
@@ -44,3 +70,273 @@ def test_metrics_2d_grid():
...
+    # ds_fail = _add_metrics(
+    #     xr.Dataset(
+    #         {"u": u_b, "v": v_c, "tracer": tr, "timeseries": timeseries}
+    #     )
+    # )
+    # ds_fail2 = _add_metrics(
+    #     xr.Dataset(
+    #         {"u": u_b, "v": v_c, "tracer": tr, "timeseries": timeseries}
+    #     )
+    # )
```
rabernat: Also these

rabernat:
@jbusecke I added a more generic approach to this type checking. All tests pass. I would just like to remove the commented code if you can recommended what to do there.

xgcm/test/test_metrics.py
jbusecke: Go ahead and remove. These were just copied over from the prototype I built [here](https://github.com/jbusecke/xarrayutils/blob/master/xarrayutils/test/test_xgcm_utils.py).
I can easily restore them, once I migrate the weighted mean to xgcm proper.

xgcm/test/test_metrics.py
jbusecke: Same. These can be removed

xgcm/grid.py
```diff
@@ -16,6 +16,14 @@
 docstrings = docrep.DocstringProcessor(doc_key="My doc string")
 
 
+def _maybe_promote_str_to_list(a):
```
jbusecke: I like this!

jbusecke:
I see that you are fixing the error message matching here. Should I close #135?

rabernat:
So the test is failing because of a black issue.
```
$ if [[ "$CONDA_ENV" == py36 ]]; then black --check xgcm; fi
would reformat /home/travis/build/xgcm/xgcm/xgcm/grid.py
would reformat /home/travis/build/xgcm/xgcm/xgcm/test/test_metrics.py
would reformat /home/travis/build/xgcm/xgcm/xgcm/test/test_grid.py
All done! 💥 💔 💥
3 files would be reformatted, 14 files would be left unchanged.
The command "if [[ "$CONDA_ENV" == py36 ]]; then black --check xgcm; fi" exited with 1.
```

@jbusecke - I must not have my black auto formatting set up correctly. Can you explain how to do that? (And preferably document it somewhere in this repo?)

rabernat:
Ok I figured it out.
```
pip install pre-commit
pre-commit install
```

jbusecke:
Thats it! Should I add a brief 'developer guide' to the docs?

rabernat:
> Thats it! Should I add a brief 'developer guide' to the docs?

👍

jbusecke:
I personally use the atom package [python-black](https://atom.io/packages/python-black), which formats on each save. Just FYI

jbusecke:
Huh? Why is it still failing?

jbusecke:
Contributer guide PR #137

jbusecke:
To me this is good to go. What do you think?

rabernat:
Please merge!

jbusecke:
👏

Issue #134, opened 2019-07-05
Rename conda test env


jbusecke:
Our ci yml files create a test environment with the generic name [`test_env`](https://github.com/xgcm/xgcm/blob/f9b58ff986da3cddc99044f6a3b576a467d09b94/ci/environment-py36.yml#L1), which is problematic since it is also used by e.g. xarray.

I propose to rename our environment to `test_env_xgcm`, so that one doesn have to reinstall environments when submitting PRs to different packages.

jbusecke:
Closed via #139

PR #139, opened 2019-07-10
Changed ci env names to unique name.

jbusecke:
I ran into problems when working on xarray and xgcm PRs since both use the name `test_env` for their environments. This gives our conda env the unique name `test_env_xgcm` so folks dont have to reinstall environments when switching between the packages.

Addresses #134


rabernat:
Good idea!

PR #153, opened 2019-09-27
[WIP] add Grid.shift

### LINKED ISSUE #151: xgcm.shift?
Sometimes it is useful to multiply values at cell centers by face heights that at cell edges. 

I'm thinking it would be nice to have `grid.shift` to blindly change locations from centers to faces and vice versa. 

e.g `grid.shift(center_var, axis="x", direction="right") * face_var` (result at face points with no interpolation)

dcherian:
Closes #151 

hacked this together quickly, comments requested

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1123,6 +1123,32 @@ class Grid:
         ax = self.axes[axis]
         return ax.diff(da, **kwargs)
 
+
+    @docstrings.dedent
+    def shift(self, da, axis, to=None):
+        """
+        Shift DataArray to
+
+        Parameters
+        ----------
+        da : xarray.DataArray
+
+        axis : str
+            axis name
+
+        to : str
+            new position to shift to
+
+        Returns
+        -------
+        da_s : xarray.DataArray
+            The shifted data
+        """
+
+        # TODO: probably need boundary here
+        return self.axes[axis]._wrap_and_replace_coords(da, da.data, position_to=to)
+
+
     @docstrings.dedent
     def derivative(self, da, axis, **kwargs):
         """


jbusecke:
I like this a lot. This helps also with fixing mislabelled datasets. I remember coding something similar a while ago, but never creating a PR...

jbusecke:
See #155 for an attempted fix of the error that is causing your tests to fail (I think that is unrelated to your changes).

xgcm/grid.py
```diff
@@ -1123,6 +1123,32 @@ def diff(self, da, axis, **kwargs):
...
+        to : str
+            new position to shift to
+
+        Returns
+        -------
+        da_s : xarray.DataArray
+            The shifted data
+        """
+
+        # TODO: probably need boundary here
```
jbusecke: Would you allow all the boundary options, like in the other operations like `.diff` etc?

The way I understand it, this would just rename the dimensions of the dataarray. If the array on e.g. the cell center has the same size as the one on the cell boundary, that's all you need. The data in the array stays untouched.

The only case where we need to intervene is if we have 'inner' and 'outer' coordinates. Then one side would potentially be cut or we would have to add values. I am just wondering if we should introduce a different keyword. On the one hand, `boundary` is familiar, but then again I am worried about the cutting. E.g. in the case of going from 'outer' to 'inner' coordinates, how would you determine which side of the array loses a slice?

jbusecke:
The CI errors should be resolved when rebasing this on the current master.

rabernat:
I really appreciate your contribution @dcherian! Thanks for working on this.

I have to say though, I'm not sure I understand the use case for this. Isn't the whole point of xgcm to avoid incorrectly applying operations across incompatible values on different grids? I'm worried it will confuse users, and that people will mistakenly use shift when interp is what they probably want.

Can you explain the use case in more detail? Ideally whatever the explanation is, it would make it into the docs.

dcherian:
@ALDepp  and I were trying to do some POP budgets. There was some weirdness where we had to a take a precomputed quantity at u points and multiple by cell heights at T points (or some such) because that's how the model does it. Fundamentally, to recreate the model's procedure we had to do this and ignore coordinate labels. This seemed like the clean way to do that within the xgcm framework..

Currently Anna is using `to_index()` and `to_xgcm()` to switch between the two representations (label and indexes) to get through the calculation.

jbusecke:
To weigh in quickly: I also had a situation with some postprocessed files at GFDL (where the dimensions were messed up), where this would come in handy to clean up data.
But I also see @rabernat 's point. 
Can we name this different or have the user load it from a different module, so we have the functionality but it remains clear that it is not a 'regular' xgcm function?

rabernat:
Sorry for letting this hang...I'd like to get it resolved.

Isn't the simplest thing to do in these cases just to use xarray's `rename` function to rename the dimension as needed? Why does xgcm have to be involved if there is no actual interpolation or difference happening?

rabernat:
Hi @dcherian - I'd like to get this PR either merged or closed. Any thoughts on my questions above?

As long as someone can explain to me (and in the docs) why this functionality is need (and not possible without xgcm), I'm happy to merge this PR.

dcherian:
I'll close it. I can't remember why `rename_dims` wasn't good enough.

Issue #154, opened 2019-10-01
Current build broken?


jbusecke:
It seems that both #153 and #130 both have the same errors that seem unrelated at first look.

> xgcm/test/test_duck_array_ops.py::test_concatenate FAILED 
> xgcm/test/test_grid.py::test_add_to_slice FAILED

The last build of the master branch was 3 months ago, did something change with the `xr.concat` function, that breaks `add_to_slice`?

I just pulled the latest master and I can confirm the same tests failing for me locally.

dcherian:
See https://github.com/pydata/xarray/pull/3362

Thanks for finding this. I think `xgcm` should test using xarray master so that these bugs can be caught before release.

rabernat:
Thanks for working on this guys!

jbusecke:
Ok I am happy to implement this. @dcherian, is there another package that does this, so I can get inspiration how to implement? 
I feel it would also be needed to trigger a rebuild of the master branch on travis (our current build is 3 months old), this would have at least alerted us a bit earlier.

dcherian:
Xarray does this with dask master. Really helps

jbusecke:
Closed via #155

PR #155, opened 2019-10-01
Quick fix for the broken build

jbusecke:
Addresses #154. 

cc'ing @dcherian @rabernat 

There are two problems caused by updates of xarray and numpy upstream (see details below). 

In this PR I want to:
- [ ] fix the issue related to the numpy __array_function__ implemented in 1.16?
- [x] Implement testing from the master branch of xarray to make it easier to catch these errors in the future.
- [x] Potentially drop support for older python versions
- [x] Add CI testing for python 3.7 environment


xgcm/grid.py
```diff
@@ -1235,7 +1235,10 @@ def add_to_slice(da, dim, sl, value):
     # beginning or end before or after will be empty)
     before = da[{dim: slice(0, sl)}]
     middle = da[{dim: sl}]
+    # the newest xarray version seems to not retain the singular dimension
+    middle = middle.expand_dims(dim)
```
jbusecke: it seems like the newest version of xarray does not retain a singular dimension by default. This fix restores that behaviour.

xgcm/test/test_duck_array_ops.py
```diff
@@ -14,4 +14,6 @@ def test_concatenate():
     concat_mixed = concatenate([a, b_dask], axis=0)
     assert isinstance(concat, np.ndarray)
     assert isinstance(concat_dask, dsk.array.Array)
-    assert isinstance(concat_mixed, np.ndarray)
+    # assert isinstance(concat_mixed, np.ndarray)
+    # the resulting mixed array is a dask array, not np. due to changes in numpy?
+    assert isinstance(concat_mixed, dsk.array.Array)
```
jbusecke: This seems to be a seperate issue, where numpy now returns a dask array if mixed array are passed? a) Is this expected? b) Should we keep this or force a computation on the dask array?

jbusecke: Hmm this only works for the py3.7 environment. Any suggestions how to fix this elegantly?

dcherian: this is likely a numpy __array_function__ protocol thing. What are the numpy versions in your different CI configs?

jbusecke:
I believe we are not specifying versions. This would make sense to me because older python versions likely default to other numpy versions in conda?

I assume we want uniformity in the output, but do we prefer numpy or dask arrays as result for the mixed case?

———————————————
juliusbusecke.com

jbusecke:
@dcherian do you know an elegant way to set the  *array_function* protocol explicitly?
My (completely uniformed and naive) intuition would be to always keep dask arrays, but I think this is a question for @rabernat.

rabernat:
This array function stuff has given us endless headaches. See https://github.com/pangeo-data/pangeo/issues/690 for some discussion.

I have had the bandwidth to try to understand it all. I appreciate you looking into it!

jbusecke:
Hmm I tried to do 
```
import os

os.environ["NUMPY_EXPERIMENTAL_ARRAY_FUNCTION"] = "0"
```
before importing numpy in both `xgcm/duck_array_ops.py` and `xgcm/tests/test_duck_array_ops.py`, but get the same error. Am I setting the env variable in the wrong spot? Is there a way to set this right when importing a package?

jbusecke:
I also tried to explicitly activate that option, and assert that the output of the mixed concat is a dask array. That again only works on the 3.7 environment.

dcherian:
I would instead ask what the expected output is supposed to be.

Do you want a dask array or a numpy array? If the latter, then explicitly convert to a numpy array in whatever function you're using. If not, rewrite so that it doesn't convert to numpy.

As I understand it, the tradeoff with the array_function stuff is that library maintainers rather than users need to take the burden of explicitly making sure everything works.

jbusecke:
I think we might need to require numpy >=1.16, for which we would have to drop support for python 2.7 and 3.5. What do you think about this?

dcherian:
You could add a different test `py37-upstream-master` so you're testing both the release and master?

jbusecke:
@dcherian do you by any chance know if it's possible to trigger an xgcm build for each release of xarray?

jbusecke:
@dcherian do you by any chance know if it's possible to trigger an xgcm build for each release of xarray?

dcherian:
Don't know.

jbusecke:
Ok thanks. For now this should suffice to alert us earlier. Ill keep looking for a solution on that one. 

Regarding the __array_function__ behavior, I seem to only be able to specifically enable it (resulting in dask arrays returned for concatenation of mixed arrays), not the other way around. Does that present a problem? 

If not we could drop 2.7 (we should do that anyways) and 3.5 support and pending the upstream xarray fix this would be done.

jbusecke:
Ughhh I somehow seem to have fragged the travis integration...will try to fix it. Sorry for the delay...

jbusecke:
Ok I think this is close. I dropped support for py2.7 and 3.5, forced numpy >=0.16 on the py36 environment and for the relevant modules force the behavior of __array_function__ to be consistent for the duck_array test. 
This does result in dask arrays as the concatenated output of the combined arrays (but doesn't break anything else). I added support for python 3.7 and the upstream master of xarray for the CI. 

One question remains: Should we leave the temporary fix for the xarray slicing (and recombination) in place? @dcherian 's fix in over at xarray wont propagate to all of our test envs immediately. So should we keep this temporary fix and remove it when the next xarray version comes around?

jbusecke:
@rabernat could you take a look? It would be good to get these sorted out, so we can move on to some of the pending PRs

.travis.yml
```diff
@@ -7,29 +7,32 @@ notifications:
...
-  - python: 2.7
-    env: CONDA_ENV=py27
-  - python: 3.5
-    env: CONDA_ENV=py35
   - python: 3.6
     env: CONDA_ENV=py36
+  - python: 3.7
+    env: CONDA_ENV=py37
+  - python: 3.7
+    env: CONDA_ENV=py37-upstream-master
```
jbusecke: Added an extra environment that tests xgcm with the upstream xarray master

ci/environment-py36.yml
```diff
@@ -2,10 +2,10 @@ name: test_env_xgcm
 dependencies:
   - python=3.6
   - xarray
-  - dask
-  - numpy
+  - numpy >= 1.16
```
jbusecke: I think this is required to get a consistent `_array_function_` bahavior.

doc/installation.rst
```diff
@@ -5,8 +5,8 @@ Installation
 Requirements
 ^^^^^^^^^^^^
 
-xgcm is compatible with python 3 and python 2.7. It requires xarray_
-(>= version 0.8.2) and dask_ (>= version 0.11.2).
+xgcm is compatible with python 3 (>= version 3.6). It requires xarray_
+(>= version 0.13.0) numpy_ (>= version 1.16.0) and dask_ (>= version 0.16.0).
```
jbusecke: I am actually not sure which dask version is required at the very least. The conda install defaults to 2.x.

xgcm/grid.py
jbusecke: @dcherians fix over at xarray will fix this, but our tests will still fail until a new version of xarray is released, or am I missing something here? Should we leave this patch in and mark it for removal in the future?

xgcm/test/test_duck_array_ops.py
jbusecke: We have a consistent output across versions now by enforcing `os.environ["NUMPY_EXPERIMENTAL_ARRAY_FUNCTION"] = "1"`. This however changes the behavior of the output (will now spit out dask arrays for mixed input). Is this acceptable?

.travis.yml
```diff
@@ -7,29 +7,32 @@ notifications:
...
   - conda update -q conda
   - conda info -a
+  - conda list
 
 install:
   - conda env create --file ci/environment-$CONDA_ENV.yml
   - source activate test_env_xgcm
+  # Install xarray from master not conda to be able to react to changes quicker
+  - if [[ "$CONDA_ENV" == "py37-upstream-master" ]]; then
+      pip install --no-deps git+https://github.com/pydata/xarray.git;
```
rabernat: Why not just put this in the environment file?

ci/environment-py37-upstream-master.yml
```diff
@@ -10,3 +9,4 @@ dependencies:
     - docrep
     - codecov
     - pytest-cov
+    - black
```
rabernat:
```
  - git+https://github.com/pydata/xarray.git
```
will install the github master into the environment

xgcm/duck_array_ops.py
```diff
@@ -5,6 +5,11 @@
 from __future__ import division
 from __future__ import print_function
 
+# fix for inconsistent numpy behavior
+import os
+
+os.environ["NUMPY_EXPERIMENTAL_ARRAY_FUNCTION"] = "1"
+
```
rabernat: I'm not really excited about this fix. We should not be modifying user environment variables. It's also not guaranteed to work; if the user imports numpy before importing xgcm, this won't have any effect.

xgcm/grid.py
rabernat: Update the comment with a reference to the xarray PR and version with which it will become unnecessary.

It's fine to leave it in if it has no performance penalty.

rabernat:
In general, I'm fine with dropping 2.7 and 3.5 support for our next release.

xgcm/test/test_duck_array_ops.py
rabernat: We can't count on that environment variable to be set. So we need a different approach.

ci/environment-py37-upstream-master.yml
jbusecke: 🤦‍♂ that is obviously easier. Will fix that right now.

.travis.yml
jbusecke: Done

rabernat:
Can we please remove the `add_to_slice` function?

It looks like it is never called from the code, only from the tests!
https://github.com/xgcm/xgcm/search?q=add_to_slice&unscoped_q=add_to_slice

If it is a utility function for tests, it should live in a test module, not the main code. Moreover, do we really need it?

If we can eliminate this function, all we really care about for this PR is `concatenate`, correct?

xgcm/duck_array_ops.py
jbusecke: Good point. I just dont understand the internals of numpy enough. But let me remove these and see if the tests run (there is a good chance that requiring numpy >=1.16 solves this on its own.

ci/environment-py37-upstream-master.yml
```diff
@@ -1,7 +1,7 @@
 name: test_env_xgcm
 dependencies:
-  - python=3.5
-  - xarray
+  - python=3.7
+  - git+https://github.com/pydata/xarray.git
```
rabernat: This does not work here. It needs to be in the pip section. See https://github.com/MITgcm/xmitgcm/blob/master/ci/environment-py36-xarraymaster.yml for a working example.

That's why the build failed: https://travis-ci.org/xgcm/xgcm/jobs/592737894

xgcm/test/test_duck_array_ops.py
jbusecke: it seems to work now that we have dropped the old python versions and require numpy >1.16. Should I see if it also works without that numpy version specified?

jbusecke:
Neat. This seems to work. So it turned out that we were testing an orphaned function in the code which we now removed and problems with the changes in the numpy internals were fixed by dropping support for old python versions. 

Thanks so much for the help here @dcherian and @rabernat.

dcherian:
hahahaha well at least it caught an xarray bug.

jbusecke:
Hahaha exactly 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #157, opened 2019-10-03
Accept multiple axes in `.interp()`


jbusecke:
I often find myself interpolating variables from the B-grid velocity point (north-east) to the tracer point, or vice versa. 
Currently this looks like this:
```
da_new = grid.interp(grid.interp(da, 'X'), 'Y')
```
I would love to be able to pass a list of axes to achive the same behavior like this:
```
da_new = grid.interp(da, ['X', 'Z'])
```

This would be a nice addition to have also for the `min()` and `max()` functions. A use case for this would be to determine the cell depth of z-star cells in MOM5, which at the velocity points are defined as the minimum of the 4 surrounding tracer cells.

dcherian:
+20.

rabernat:
Great idea.

Implementation question: does the order matter here?

dcherian:
Hmm it's a bunch of linear operations so order shouldn't matter?

jbusecke:
Hmm I have always assumed that not, but perhaps there is a special case. 
We could still accomodate that by internally following the order of the list. E.g. ‘grid.interp(..., [‘X’, ‘Y’])’ vs ‘grid.interp(..., [‘Y’, ‘X’])’
What do you think?

Julius J.M. Busecke, Ph.D. (he/him)
———————————————
Postdoctoral Research Associate
Princeton University • Geosciences
408A Guyot Hall • Princeton NJ
juliusbusecke.com

jbusecke:
I just put out an idea how to implement this in #159, feedback greatly appreciated.

dcherian:
THis can be closed?

jbusecke:
Yes. Closed via #159

PR #159, opened 2019-10-09
WIP: Enable 'single axis' operations to accept multiple axes

jbusecke:
Here is a first attempt to implement #157.

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1054,7 +1054,6 @@ class Grid:
         out = da
         for axx in axis:
             kwargs = {k: v[axx] for k, v in multi_kwargs.items()}
-            print("internal", axx)
             ax = self.axes[axx]
             func = getattr(ax, funcname)
             metric_weighted = kwargs.pop("metric_weighted", False)
@@ -1063,7 +1062,7 @@ class Grid:
                 metric_weighted = (metric_weighted,)
 
             if metric_weighted:
-                metric = self.get_metric(da, metric_weighted)
+                metric = self.get_metric(out, metric_weighted)
                 out = out * metric
 
             out = func(out, **kwargs)
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1191,7 +1191,7 @@ class Grid:
         --------
         Each keyword argument can be provided as a `per-axis` dictionary. For instance,
         if we want to find the minimum of sourrounding grid cells for a global 2D dataset
-        in both X and Y axis, but the fill value at the boundary should be differenc
+        in both X and Y axis, but the fill value at the boundary should be different
         for each axis, we can do this:
 
         >>> grid.min(da, ['X', 'Y'], fill_value={'X':0, 'Y':100})
@@ -1220,7 +1220,7 @@ class Grid:
         --------
         Each keyword argument can be provided as a `per-axis` dictionary. For instance,
         if we want to find the maximum of sourrounding grid cells for a global 2D dataset
-        in both X and Y axis, but the fill value at the boundary should be differenc
+        in both X and Y axis, but the fill value at the boundary should be different
         for each axis, we can do this:
 
         >>> grid.max(da, ['X', 'Y'], fill_value={'X':0, 'Y':100})
@@ -1236,12 +1236,24 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        %(neighbor_binary_func.parameters.no_f)s
+
+        Examples
+        --------
+        %(grid_func.examples)s
 
         Returns
         -------
         da_i : xarray.DataArray
             The cumsummed data
+
+        Examples
+        --------
+        Each keyword argument can be provided as a `per-axis` dictionary. For instance,
+        if we want to compute the cumulative sum of global 2D dataset
+        in both X and Y axis, but the fill value at the boundary should be different
+        for each axis, we can do this:
+
+        >>> grid.max(da, ['X', 'Y'], fill_value={'X':0, 'Y':100})
         """
         return self._grid_func("cumsum", da, axis, **kwargs)
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1256,18 +1256,7 @@ class Grid:
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
-        # This is a workaround, that should be eliminated
-        # once cumsum can axcept multiple axes
-        # (see https://github.com/xgcm/xgcm/pull/159)
-
-        if isinstance(axis, str):
-            axis = [axis]
-
-        out = weighted
-        for ax in axis:
-            out = self.cumsum(out, ax, **kwargs)
-
-        return out
+        return self.cumsum(weighted, axis, **kwargs)
 
     @docstrings.dedent
     def average(self, da, axis, **kwargs):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1228,6 +1228,24 @@ class Grid:
         return self._grid_func("max", da, axis, **kwargs)
 
     @docstrings.dedent
+    def cumsum(self, da, axis, **kwargs):
+        """
+        Cumulatively sum a DataArray, transforming to the intermediate axis
+        position.
+
+        Parameters
+        ----------
+        %(grid_func.parameters)s
+        %(neighbor_binary_func.parameters.no_f)s
+
+        Returns
+        -------
+        da_i : xarray.DataArray
+            The cumsummed data
+        """
+        return self._grid_func("cumsum", da, axis, **kwargs)
+
+    @docstrings.dedent
     def _apply_vector_function(self, function, vector, **kwargs):
         # the keys, should be axis names
         assert len(vector) == 2
@@ -1422,24 +1440,6 @@ class Grid:
         """
 
         return self._apply_vector_function(Axis.diff, vector, **kwargs)
-
-    @docstrings.dedent
-    def cumsum(self, da, axis, **kwargs):
-        """
-        Cumulatively sum a DataArray, transforming to the intermediate axis
-        position.
-
-        Parameters
-        ----------
-        %(grid_func.parameters)s
-        %(neighbor_binary_func.parameters.no_f)s
-
-        Returns
-        -------
-        da_i : xarray.DataArray
-            The cumsummed data
-        """
-        return self._grid_func("cumsum", da, axis, **kwargs)
 
 
 def raw_interp_function(data_left, data_right):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1048,26 +1048,27 @@ class Grid:
         da_i : xarray.DataArray
             The interpolated data
         """
-        if isinstance(axis, list):
+        if isinstance(axis, list) or isinstance(axis, tuple):
             out = da
             for axx in axis:
-                out = self.interp(out, axx, **kwargs)
+                out = self.interp(out, axx, metric_weighted=metric_weighted, **kwargs)
                 #!!!the kwargs needs to be provided in a dict here?
+                # Also what about a case where one would want to preserve different metrics?
             return out
+        else:
+            ax = self.axes[axis]
 
-        ax = self.axes[axis]
+            if isinstance(metric_weighted, str):
+                metric_weighted = (metric_weighted,)
 
-        if isinstance(metric_weighted, str):
-            metric_weighted = (metric_weighted,)
-
-        if metric_weighted:
-            metric = self.get_metric(da, metric_weighted)
-            da = da * metric
-        out = ax.interp(da, **kwargs)
-        if metric_weighted:
-            metric_new = self.get_metric(out, metric_weighted)
-            out = out / metric_new
-        return out
+            if metric_weighted:
+                metric = self.get_metric(da, metric_weighted)
+                da = da * metric
+            out = ax.interp(da, **kwargs)
+            if metric_weighted:
+                metric_new = self.get_metric(out, metric_weighted)
+                out = out / metric_new
+            return out
 
     @docstrings.dedent
     def _apply_vector_function(self, function, vector, **kwargs):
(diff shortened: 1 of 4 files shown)


xgcm/grid.py
```diff
@@ -1031,6 +1031,12 @@ def interp(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The interpolated data
         """
+        if isinstance(axis, list):
+            out = da
+            for axx in axis:
+                out = self.interp(out, axx, **kwargs)
+                #!!!the kwargs needs to be provided in a dict here?
```
jbusecke: I am not sure if this is the most elegant way, but it fundamentally works. I am not sure how we would parse the kwargs here. We shurely would want to be able to pass e.g. different boundaries for each axis.

xgcm/test/test_grid.py
```diff
@@ -640,4 +640,18 @@ def test_grid_ops(all_datasets):
...
+# @pytest.mark.parametrize("func", ["interp", "diff", "max", "min"])
+@pytest.mark.parametrize("func", ["interp"])
+def test_multi_axis_input(all_datasets, func):
+    ds, periodic, expected = all_datasets
+    grid = Grid(ds)
+    axes = list(grid.axes.keys())
+    for varname in ["data_c", "data_g"]:
+        serial = ds[varname]
+        for axis in axes:
+            serial = getattr(grid, func)(serial, axis)
```
jbusecke: I am also wondering if we can use something like this to generalize all the functions that act on a single axis (diff, interp, min, max, shift) and save ourselves some boilerplate code? Not necessarily important right now.

jbusecke:
The CI error will be fixed with #130

jbusecke:
Yes I can put this on my list for tomorrow probably. Today I will be occupied with the GFDL Lab Review until late.

———————————————
jbusecke.github.io<http://jbusecke.github.io>

On Oct 29, 2019, at 12:21 PM, Ryan Abernathey <notifications@github.com<mailto:notifications@github.com>> wrote:


Julius do you want to rebase this and see if we can get the tests passing?

jbusecke:
So this works in principle. 

We will have to make a decision how to pass e.g. boundary inputs. I would think we could do something like `grid.interp(da, ['X','Y'], boundary='fill')`, which would apply the boundary to both axes or `grid.interp(da, ['X', 'Y'], boundary={'X':'fill', 'Y':'extend'}), which would apply different boundaries for each axis? What do you think it the most elegant way?

jbusecke:
Oh yes, I meant it as both 😆. Ill have a look later. I think `Grid` must do something similar too?

jbusecke:
I have implemented `_parse_axes_kwargs` in `Grid`, for every 'multi-axis' input this automatically converts input like `boundary='fill` into a format like this
```
boundary = {'X':'fill','Y':'fill'}
```
One can also pass such a dictionary to apply different boundary conditions (I am currently testing how to pass `metric_weighted` in the same way, but encounter some errors. 

We also need docs for this, but I wanted to ask about the general design first.

xgcm/grid.py
```diff
@@ -838,6 +838,17 @@ def __init__(
         if metrics is not None:
             self._assign_metrics(metrics)
 
+    def _parse_axes_kwargs(self, kwargs):
+        """Parse string kwargs input to dictionary for each axis.
+        Used e.g. for `boundary`"""
+        parsed_kwargs = dict()
+        if isinstance(kwargs, dict):
+            parsed_kwargs = kwargs
```
rabernat: Do we want to do a consistency check here? Raise an error if the axis is not valid? Or does that happen downstream?

xgcm/grid.py
```diff
@@ -1049,20 +1060,32 @@ def interp(self, da, axis, metric_weighted=False, **kwargs):
...
         """
+        if isinstance(axis, list) or isinstance(axis, tuple):
+            # parse multi axis kwargs like `boundary`
+            multi_kwargs = {k: self._parse_axes_kwargs(v) for k, v in kwargs.items()}
+            out = da
+            for axx in axis:
+                out = self.interp(
+                    out, axx, **{k: v[axx] for k, v in multi_kwargs.items()}
+                )
+            return out
```
rabernat: This block should probably be factored out as well, since it is reused often

xgcm/grid.py
jbusecke: In the end the downstream functions like `diff`, `interp` etc will just 'see' `boundary='fill'` for a single axis. The errors should be raised there. Currently axis that do not exist (e.g. if passing `{..., 'Z':{...}}` to a dataset that has only x and y axes) are just silently omitted. Do you think we should raise a warning/error?

xgcm/grid.py
jbusecke: I was thinking about that too, will do with the next step.

rabernat:
> what exactly do you mean by API docs? Inline comments? Or examples how to use it (honestly dont know precisely what that means).

The docstring of the functions, which describe in detail the expected inputs and outputs. 

https://xgcm.readthedocs.io/en/latest/api.html

jbusecke:
Got it. Will work on that on the train ride back tonight.

rabernat:
@jbusecke - can you give us an update on this PR? That train ride was a quite a while ago. 🤣

jbusecke:
LOL. So many train rides ago. I have not worked on this in a while, sorry for that. Just took a look at it and I was able to get the test passed locally. I will polish it and see if it gets through the CI.

xgcm/grid.py
jbusecke: I think this is a more general question and could be handled in a seperate PR?

xgcm/grid.py
jbusecke: I am not sure how to achieve this properly. Another class method, e.g. `mulit_axis_func(self, funcname)` which will grab e.g. `self.interp` and execute it over each axis? Or is this to convoluted?

jbusecke:
Ok cool, the tests pass. There are a few design questions still needing review (see detailed comments on the code).

xgcm/grid.py
jbusecke: I think this has been addressed in the test. See `test_missingdim` and `test_missingaxis` in `test_metric_ops_multi_axis.py`. Or do you think we need more fine grained control? Could be nice to have a warning that mentions specifically the argument that is not conform. Let me know.

xgcm/grid.py
jbusecke: Thinking about it even the code below is very repetitive between e.g. `interp`/`max` etc (e.g. the metric weighted step). Is there a sleek way to factor this out? 

And should this be done now, or in a seperate PR?

jbusecke:
Ok this looks good in terms of tests. I will try to work on extending the docs for these cases (cant promise I will get to it today though). But if there are no major concerns, I will try to get this done soon, so we can include it in the next [release](#168).

xgcm/grid.py
rabernat: I think it's okay for now.

xgcm/grid.py
rabernat: I am still not comfortable with how much repetition is introduced with this. We do need to refactor the code somehow. It needs to be done here, in this PR, because this PR is the one that introduced the pattern. Otherwise we build up a lot of technical debt.

I am not sure the best way to do it. A class method seems reasonable.

xgcm/grid.py
```diff
@@ -1148,9 +1171,18 @@ def diff(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The differenced data
         """
-
-        ax = self.axes[axis]
-        return ax.diff(da, **kwargs)
+        if isinstance(axis, list) or isinstance(axis, tuple):
```
rabernat: here

xgcm/grid.py
```diff
@@ -1284,9 +1305,16 @@ def min(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The minimal data
         """
-
-        ax = self.axes[axis]
-        return ax.min(da, **kwargs)
+        if isinstance(axis, list) or isinstance(axis, tuple):
```
rabernat: here

xgcm/grid.py
```diff
@@ -1304,9 +1332,16 @@ def max(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The maximal data
         """
-
-        ax = self.axes[axis]
-        return ax.max(da, **kwargs)
+        if isinstance(axis, list) or isinstance(axis, tuple):
```
rabernat: here

xgcm/grid.py
jbusecke: I agree with you, and have started looking into it.

jbusecke:
I have introduced a class method `_grid_func`, which calls the axis functions (interp, max, min, diff). It also handles the repeated execution if multiple axes are passed and also the weighting. What do you think about this @rabernat?

xgcm/grid.py
```diff
@@ -1028,41 +1039,66 @@ def __repr__(self):
...
+
+    @docstrings.dedent
+    def interp(self, da, axis, **kwargs):
+        """
+        Interpolate neighboring points to the intermediate grid point along
+        this axis.
+
+        Parameters
+        ----------
+        %(grid_func.parameters.no_f)s
```
jbusecke: @rabernat maybe you know how to achieve this: I want to parse the input parameters from `_grid_func`, is that possible?

xgcm/grid.py
```diff
@@ -1028,41 +1039,66 @@ def __repr__(self):
...
         ----------
-        axis : str
-            Name of the axis on which to act
+        axis : {str, list, tuple}
+            Name of the axis on which to act. Multiple axes can be passed as list or
+            tuple (e.g. ['X', 'Y']). Functions will be executed over each axis in the
+            given order.
         metric_weighted : str or tuple of str
             If an axis or list of axes is specified,
             the grid metrics will be used to determined the weight for interpolation.
```
jbusecke: I need to replace the `interpolation` here with a more general word

xgcm/test/test_metrics_ops_single_axis.py
```diff
@@ -7,20 +7,59 @@
...
+        new = func(
+            ds[variable], axis, metric_weighted=metric_weighted, boundary=boundary
+        )
+        assert new.equals(expected)
+
+    @pytest.mark.parametrize("multi_axis", ["X", ["X"], ("Y"), ["X", "Y"], ("Y", "X")])
+    @pytest.mark.parametrize(
+        "metric_weighted", ["X", ["X", "Y"], {"X": "Y", "Y": ["X", "Y"]}]
+    )  #
+    def test_interp_conservative_multi_axis(
```
jbusecke: should be renamed since it tests more than just interp now

xgcm/test/test_metrics_ops_single_axis.py
```diff
@@ -7,20 +7,59 @@
...
-    new = grid.interp(ds[variable], axis, metric_weighted=metric_weighted)
-    assert new.equals(expected)
+class TestParametrized:
+    @pytest.mark.parametrize("axis", ["X", "Y"])
+    @pytest.mark.parametrize("metric_weighted", ["X", ("Y",), ("X", "Y"), ["X", "Y"]])
+    @pytest.mark.parametrize(
+        "periodic", ["True", "False", {"X": True, "Y": False}, {"X": False, "Y": True}]
+    )
+    @pytest.mark.parametrize("boundary", ["fill", "extend", "extrapolate"])
+    def test_interp_conservative(
```
jbusecke: same here

xgcm/grid.py
```diff
@@ -838,6 +838,17 @@ def __init__(
         if metrics is not None:
             self._assign_metrics(metrics)
 
+    def _parse_axes_kwargs(self, kwargs):
+        """Parse string kwargs input to dictionary for each axis.
+        Used e.g. for `boundary`"""
```
rabernat: Here it would be great to just have an example of in --> out

xgcm/grid.py
rabernat: Is it `_grid_func`?

xgcm/grid.py
```diff
@@ -1028,41 +1039,66 @@ def __repr__(self):
...
-
-        if metric_weighted:
-            metric = self.get_metric(da, metric_weighted)
-            da = da * metric
-        out = ax.interp(da, **kwargs)
-        if metric_weighted:
-            metric_new = self.get_metric(out, metric_weighted)
-            out = out / metric_new
-        return out
+        return self._grid_func("interp", da, axis, **kwargs)
```
rabernat: Can we skip writing "interp" and get the function name from context?

I think the answer is no. So it's okay.

xgcm/test/test_grid.py
```diff
@@ -639,4 +633,28 @@ def test_grid_ops(all_datasets):
...
+    grid = Grid(ds, periodic=periodic)
+    axes = list(grid.axes.keys())
+    for varname in ["data_c", "data_g"]:
+        serial = ds[varname]
+        for axis in axes:
+            boundary_axis = boundary
+            if isinstance(boundary, dict):
+                boundary_axis = boundary[axis]
+            serial = getattr(grid, func)(serial, axis, boundary=boundary_axis)
+        full = getattr(grid, func)(ds[varname], axes, boundary=boundary)
```
rabernat: clever :)

xgcm/test/test_metrics_ops_single_axis.py
```diff
@@ -7,20 +7,59 @@
 from xgcm.test.datasets import datasets_grid_metric
 
 
+@pytest.mark.parametrize("funcname", ["interp", "diff", "min", "max"])
```
rabernat: Do we need cumsum here too?

xgcm/test/test_metrics_ops_single_axis.py
rabernat: ok then please rename it

xgcm/grid.py
jbusecke: We probably can. But it doesnt seem that easy. So maybe we stick with a bit of repetition now and check on this later?

PR #162, opened 2019-10-10
WIP: Adding weighted average and cumsum

jbusecke:
Added average, generalized tests for integrate/average


jbusecke:
Regarding the naming:
I opted for:
`average` as the metric-aware of `mean`.
`integrate` as the metric-aware of `diff`.

What would be a good name for the metric version of `cumsum`?

I also want a metric aware version of `interp` (which multiplies values, then interpolates and divides by metric again), my suggestion is `remap` (MOM lingo), but that might be confusing. 

What do you all think?

jbusecke:
cc ing @timothyas, @dcherian

jbusecke:
One important aspect here is: Do we check the weights for consistency?

E.g. if for some reason the surface area and tracer values have misaligned values (I have encountered this situation in MOM5), the weighted global average would be biased. 
One could select both data and metric based on an AND between `xr.isnan(data)` and `xr.isnan(metric)` but that would add overhead. 

I think it would be good to eventually implement this with [`xarray.weighted`](https://github.com/pydata/xarray/issues/422), which does account for these things afaik. 

Any comments/opinions appreciated.

xgcm/grid.py
```diff
@@ -1162,29 +1183,49 @@ def integrate(self, da, axis, **kwargs):
...
+            Name of the axis on which to act
+        %(neighbor_binary_func.parameters.no_f)s
+
+        Returns
+        -------
+        da_i : xarray.DataArray
+            The averaged data
+        """
+        # check if axis input is compatible
+        self._check_axes(axis)
```
rabernat: Doesn't this check happen in `get_metric` anyway? Why do we need to call it twice?

xgcm/grid.py
```diff
@@ -941,6 +941,27 @@ def _assign_metrics(self, metrics):
                 # check for duplicate dimensions among each axis metric
                 self._metrics[metric_axes].append(metric_var)
 
+    def _check_axes(self, axis):
+        """Checks if axis(or axes) in `axis` are part of the grid"""
+        check_axes = [ax for ax in axis if ax not in self.axes]
+        check_axes_str = ",".join(check_axes)
+        if any(check_axes):
+            raise ValueError("Axis %s not found in grid object" % check_axes_str)
```
rabernat: I'm not sure I understand the logic for splitting these things out. Isn't it an unnecessary layer of indirection. 

http://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

xgcm/grid.py
```diff
@@ -941,6 +941,27 @@ def _assign_metrics(self, metrics):
...
+            all_dim = self.axes[ax].coords.values()
+            matching_dim = [di for di in all_dim if di in da.dims]
+            if len(matching_dim) == 1:
+                dim.append(matching_dim[0])
+            else:
+                raise ValueError(
+                    "Did not find single matching dimension corresponding to axis %s. Got (%s)"
+                    % (ax, matching_dim)
+                )
+        return dim
```
rabernat: This makes sense to split out because we call it twice and it's pretty complicated.

rabernat:
In my review I forgot to mention...**documentation**! 😁

jbusecke:
> In my review I forgot to mention...**documentation**! 😁

But of course! This was just the first swing! Glad you like so far. Let me get some stuff done and get right back to this **so amped**

xgcm/grid.py
jbusecke: Oh you might be right. Ill investigate

jbusecke:
Also thoughts on the naming of additional functions @rabernat ?

rabernat:
> What would be a good name for the metric version of `cumsum`?

`cumint`

> I also want a metric aware version of `interp` (which multiplies values, then interpolates and divides by metric again), my suggestion is `remap` (MOM lingo), but that might be confusing.

Can we instead make this an option for interp?

jbusecke:
Like this? `grid.interp(da, axis, conservative=True)`?

dhruvbalwada:
- I think the metric aware function of mean should just be mean with an option for weights as an argument. But since numpy follows the same convention, this is probably appropriate. 
- Why is the metric aware 'diff' 'integrate'? Should it not be 'differentiate'?

jbusecke:
> * I think the metric aware function of mean should just be mean with an option for weights as an argument. But since numpy follows the same convention, this is probably appropriate.

I think it is good to have a descriptively named function, to make people more aware. And yes, staying in line with numpy/xarray is another compelling reason.

> * Why is the metric aware 'diff' 'integrate'? Should it not be 'differentiate'?

Oh yes, that was a typo. Fixed the initial message.

xgcm/grid.py
jbusecke: Ok i have dug a little. Consider this small example:
```
from xgcm import Grid

axis = 'X'
funcname = 'integrate'

del coords[axis]

del_metrics = [k for k in metrics.keys() if axis in k]
for dm in del_metrics:
    del metrics[dm]

grid = Grid(ds, coords=coords, metrics=metrics)

func = getattr(grid, funcname)

grid.get_metric(ds.u, ['X'])
```

This does indeed raise an error, but it is somewhat cryptic: 
```
"Unable to find any combinations of metrics for array dims {'yt', 'time', 'xu', 'zt'} and axes ['X']"
```

So while I agree that the checking should probably be removed, should we work on the error here?

xgcm/grid.py
jbusecke: I think ill remove this. See below.

xgcm/grid.py
jbusecke: And I will reuse it even more often for the other ones.

xgcm/grid.py
jbusecke: We can discuss this, for now I will remove the additional function and check that this error is properly raised in the tests.

jbusecke:
Added a (hacky) version of `grid.cumint` (this will do for now, but should be cleaner once `grid.cumsum` supports multiple axes #159).

@rabernat I also removed that additional check and refactored the tests. 

I will work on the docs tomorrow.

jbusecke:
Ok so as a last feature I added the option `conservative` to `grid.interp()` (it is False by default to preserve the old behavior. 
`grid.interp(da, axis, conservative=True)` does `grid.interp(da*dx, 'X') / dx_on_new_position` or `grid.interp(da*dy, 'Y') / dy_on_new_position`, which will conserve e.g the total tracerflux during an interpolation. Now writing the docs. I would love to get this merged before Monday so that @jhamman has a chance to integrate these into the hackathon environments.

jbusecke:
I have generalized the conservation option in `grid.interp`: You can pass an axis (or several) to `grid.interp` like this:

```
grid.interp(tracer, 'X', conserve=['X', 'Y'])
```
What this does is to multiply tracer with the area of the tracer cell, interpolate it onto the u cell and divide it by the area of the u cell. The horizontal tracer content is exactly conserved.

I think this is an important option to have but I wasnt able to give a good example in the docs, which is why I left it out right now (the example dataset is quite smooth, so the differences between the "regular" interp and the conservative one were miniscule).

I however think that this is a way to implement finite volume differentiation as an option (see the discussion over at #163). We should chat about this over there, and move on with this PR. 

@rabernat are you ok leaving this functionality 'hidden' for now? It is properly tested. Also if there is any chance you could review this before Monday that would be awesome. I will try to put out a demo binder for the hackathon.

xgcm/grid.py
```diff
@@ -1162,28 +1211,79 @@ def integrate(self, da, axis, **kwargs):
...
 
         weight = self.get_metric(da, axis)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
-        return weighted.sum(dim)
+        # get dimension(s) corresponding
+        # to `da` and `axis` input
+        dim = self._get_dims_from_axis(da, axis)
+        # do we need to pass kwargs?
```
jbusecke: What kind of kwargs would be passed here? Since it is a reducing operation, the boundaries should not matter?

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -0,0 +1,218 @@
+from __future__ import print_function
+import pytest
+
+from xgcm.grid import Grid
+from xgcm.test.datasets import datasets_grid_metric
+
+
+def _expected_result(da, metric, grid, dim, axes, funcname):
```
timothyas: After this is merged I could consolidate the derivative tests to this module if you agree.

jbusecke: I did that (I think) in `test_metric_ops_single_axis.py`. The idea here is to seperate the operations that can be executed on one axis after another (e.g. derivative, interp) and others that need to consider multidimensional metrics. This will probably have to be generalized if we add the finite volume ones (I have a plan how to do this elegantly, but will postpone until after the hackathon).

jbusecke: I guess I only renamed the `test_derivative` for now. This could def need some work afterwards!

xgcm/grid.py
jbusecke: Good point. Ill add that.

jbusecke:
@timothyas, that is a great suggestion! We should definitely do that.

Would it be possible to add these + the derivative docs in #163? As mentioned before, I would like to get this merged quickly. 

@rabernat the functionality is explained in a very basic way, would you consider merging this so we can have it in the standard packages for the hackathon? We could still merge the docs tomorrow or later so that users can access them via rtd (no need for those changes to be on the hackathon environments!)

xgcm/grid.py
```diff
@@ -1015,7 +1029,7 @@ def __repr__(self):
         return "\n".join(summary)
 
     @docstrings.dedent
-    def interp(self, da, axis, **kwargs):
+    def interp(self, da, axis, conserve=False, **kwargs):
```
rabernat: can we make this `metric_weighted=False` instead?

xgcm/grid.py
```diff
@@ -1024,6 +1038,11 @@ def interp(self, da, axis, **kwargs):
         ----------
         axis : str
             Name of the axis on which to act
+        conserve : str or tuple of str
+             Axis along to conserve data. The data will be multiplied with the metric
+             appropriate to `conserve` and the interpolated data will be divided by
+             the same metric, except located at the new grid location
+             (the default is False).
```
rabernat: Let's eliminate discussion of conservation and go with a simpler name here.
```
metric_weighted : str or tuple of str
    If an axis or list of axes is specified, the grid metrics will be used to determined the weight for interpolation.
    If `False` (default), the points will be weighted equally.
```

xgcm/grid.py
```diff
@@ -1162,28 +1192,79 @@ def integrate(self, da, axis, **kwargs):
...
-                    % (ax, matching_dim)
-                )
+            out = self.cumsum(out, ax, **kwargs)
+
+        return out
+
+    @docstrings.dedent
+    def average(self, da, axis, **kwargs):
+        """
+        Perform weighted average along specified axis or axes,
```
rabernat: "weighted average" => "weighted mean reduction"

xgcm/grid.py
```diff
@@ -1162,28 +1192,79 @@ def integrate(self, da, axis, **kwargs):
...
+        accounting for grid metrics. (e.g. cell length, area, volume)
+
+        Parameters
+        ----------
+        axis : str, list of str
+            Name of the axis on which to act
+        %(neighbor_binary_func.parameters.no_f)s
```
rabernat: Need docstring entry for `**kwargs`.

xgcm/test/test_metric_ops_multi_axis.py
timothyas: Ah I see now. Yeah going to FV will require some generalization but it should be do-able. Thanks

timothyas:
> Would it be possible to add these + the derivative docs in #163? As mentioned before, I would like to get this merged quickly.

No problem!

xgcm/grid.py
jbusecke: Good suggestion. Done

xgcm/grid.py
jbusecke: Done

xgcm/grid.py
jbusecke: I was wrong...its just `average` and `integrate` that need a rewrite. I put it in manually, we can see if we want to import the docstring from xarray later? I was not able to get that to work right now.

jbusecke:
Ok @rabernat, I made the changes requested.

jbusecke:
🍾

Issue #168, opened 2019-12-10
Next release?


andersy005:
Is there a timeline on when the next "stable" version of xgcm will be available? it's been a while since the last release.

rabernat:
It would be great to have a release soon. I've been swamped this semester so have not taken it on myself. If one of the other contributors wanted to organize a release, I would be 100% 👍 .

jbusecke:
I was super swamped until now, but would be happy to drive this forward. Is there a tutorial on which steps are involved? Still very unfamiliar with the whole process but eager to learn.

andersy005:
>  Is there a tutorial on which steps are involved?


@jbusecke, here are the steps that I would do for PyPI.

0. Update `~/.pypirc` with your PyPI credentials

```
[distutils]
index-servers=
    pypi

[pypi]
username: xxxxxxxxxx
password: xxxxxxxxxxxxxxxxxx
```

1.  Tag a release and push to github

        git tag -a 1.0.0 -m "Version 1.0.0"
        git push origin master --tags

2.  Build and publish release on PyPI

    ```
    git clean -xfd  # remove any files not checked into git
    python setup.py sdist bdist_wheel   # build package
    twine upload dist/*  # register and push to pypi. This requires having `twine` installed (It's available on PyPI and conda-forge)
    ```

jbusecke:
Thank you very much @andersy005, Ill try to get to it tomorrow. Is is kosher to save username and password unencrypted like that?

andersy005:
You are welcome! 

> Is is ok to save username and password unencrypted like that?

There's a much safer way, **using an API token**:

- Head over to https://pypi.org/manage/account/ to create an API token
- Here's the documentation on how to update the `~/.pypirc` file to use the API token: https://pypi.org/help/#apitoken
- Remember to set the permission of `~/.pypirc` with  `chmod 600 ~/.pypirc`

jbusecke:
Awesome, thanks. Will hopefully get to it tomorrow.

rabernat:
What is up with this? Are we going to do a release or not? Anything needed from me?

jbusecke:
Hey @rabernat and @andersy005, sorry for dropping the ball on this. I was pretty swamped over here.

@andersy005, if I understand your changes correctly (and I might very well miss something here), we only need to trigger a release on github (v.0.3.0, I assume) and then the release on pypi will be triggered? Do we need to do anything after that for conda? Or is this based on the pypi release?

[...]

PR #170, opened 2020-01-16
Add GitHub Action to publish xgcm to PyPI on release.

andersy005:
Towards fixing #168 

For this workflow to automatically publish xgcm to PyPI (after a release is created), the repo admin will need to add `PYPI_USERNAME` and `PYPI_PASSWORD` secrets in https://github.com/xgcm/xgcm/settings/secrets.

This workflow will be triggered whenever a release is created on Github.

### DIFF
--- a/.github/workflows/pythonpublish.yaml
+++ b/.github/workflows/pythonpublish.yaml
@@ -1,0 +1,26 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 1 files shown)


andersy005:
Cc @jbusecke

rabernat:
This is great @andersy005. Thanks so much! I'll try to add the secrets.

rabernat:
Just added secrets. So we should be good to go with #168.

PR #174, opened 2020-02-03
Fix GitHub action YAML indentation

andersy005:
I noticed that [this action](https://github.com/xgcm/xgcm/actions) didn't run after the release of v0.3.0. I don't know the reason behind this yet. In the meantime, This PR fixes the YAML indentation (not sure the prior indentation was the culprit).

### DIFF
--- a/.github/workflows/pythonpublish.yaml
+++ b/.github/workflows/pythonpublish.yaml
@@ -8,19 +8,19 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Set up Python
-      uses: actions/setup-python@v1
-      with:
-        python-version: '3.x'
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install setuptools setuptools-scm wheel twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-      run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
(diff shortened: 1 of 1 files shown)


jbusecke:
Thanks @andersy005, is there a way to trigger the action manually? Or do we need to release 0.3.1 😆

andersy005:
> is there a way to trigger the action manually?

Not that I know of :(. 

- One solution is to update the activity types to `[created, edited, published, prereleased]`

https://github.com/xgcm/xgcm/blob/8cd73f5e66d946dd7174a1bfec9f3d8e91b0f7cd/.github/workflows/pythonpublish.yaml#L3-L5

And then to add some trivial edits to the release https://github.com/xgcm/xgcm/releases/tag/v0.3.0 . This will trigger the GitHub Action :) 

- Another solution is to merge this PR, delete the existing release, and re-release/publish it again 

- Another alternative would consist of 

  1. Checking out the latest tag of xgcm to a local machine, building the distribution, and then publishing it:
```bash
python setup.py sdist bdist_wheel
 twine upload dist/*
```
 2. Merging this PR. The GitHub action will be trigger after the next planned release

rabernat:
Should we manually release 0.3.0 then?

andersy005:
> Should we manually release 0.3.0 then?

👍

jbusecke:
Does @rabernat need to do this because of the authentification?

jbusecke:
Should I try this route first?
> Another solution is to merge this PR, delete the existing release, and re-release/publish it again

That way we are sure that it will work automatically next time?

rabernat:
I don't believe that releases can be deleted. I'll manually release 0.3.0 and we can use this going forward.

rabernat:
I manually released 0.3.0: https://pypi.org/project/xgcm/

PR #179, opened 2020-02-16
credit Gael for LLC grid image

rabernat:
Add proper attribution to @gaelforget for LLC grid.

Gael is this okay, or would you prefer different language?

Sorry for neglecting this earlier.

### DIFF
--- a/doc/grid_topology.rst
+++ b/doc/grid_topology.rst
@@ -24,7 +24,10 @@ xgcm is capable of understanding the connections between these grid
   :scale: 20 %
   :alt: Cubed Sphre Grid
 
-  Example of a cubed-sphere grid from the MIT General Circulation Model.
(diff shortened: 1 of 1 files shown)


gaelforget:
Thanks Ryan! This is great. 

One of those things that's been stuck on my todo list forever is putting DOIs on the grid files in ftp://mit.ecco-group.org/ecco_for_las/version_4/grids

Once I get that done, I could add another DOI here or something like that -- would that be useful?

rabernat:
> One of those things that's been stuck on my todo list forever is putting DOIs on the grid files in ftp://mit.ecco-group.org/ecco_for_las/version_4/grids

An easy solution is to just stick them on https://zenodo.org/

In this case, we are really just using the figure as an illustration for the documentation. We are not actually reading the grid file. So it would not be useful specifically here. But in general, yes, I support you sharing the grid files this way.

PR #184, opened 2020-03-05
Add note in documentation on reversing `cumint` direction

ali-ramadhan:
Had to calculate a cumulative integral from the bottom up but documentation did not provide any hints on how to do this so this PR adds a sentence in case other users would like to do the same.

I just had to do `ds = ds.reindex(Z=ds.Z[::-1], Zl=ds.Zl[::-1])`.

Hint from https://stackoverflow.com/questions/54677161/xarray-reverse-an-array-along-one-coordinate

I edited the Jupyter notebook doc directly to avoid modifying a ton of Jupyter metadata.

I'm computing geostrophic velocity profiles from thermal wind balance if anyone is interested: https://github.com/ali-ramadhan/LESbrary.jl/blob/3041967c8bea4b4a58d7a208d6db1fbece591078/src/sose_data.py#L64-L115

### DIFF
--- a/doc/grid_metrics.ipynb
+++ b/doc/grid_metrics.ipynb
@@ -498,7 +498,10 @@
     "grid.cumsum( -grid.integrate(ds.UVEL,'Z') * ds.dyG, 'Y', boundary='fill')\n",
     "```\n",
     "\n",
-    "Except that, once again, one does not have to remember the matching metric while using `cumint`."
+    "Except that, once again, one does not have to remember the matching metric while using `cumint`.",
+    "\n",
+    "To reverse the integral bounds you can reindex the coordinate along which the cumulative integral\n",
+    "is to be evaluated for example, with `ds.reindex(Z=ds.Z[::-1], Zl=ds.Zl[::-1])` before calling `cumint`."
    ]
   },
   {


jbusecke:
Hi @ali-ramadhan,

thanks for using xgcm and also adding to the documentation. Your example looks really cool

> `Σdz_dΘdx` I was not aware you can use greek letters in python. 🤯

I believe this here would be equivalent, and perhaps more xgm'ish:
 
`ds_cumint_reverse = ds.integrate(axis) - ds.cumint(axis)`

Could you verify that this would work in your case?

rabernat:
> I believe this here would be equivalent, and perhaps more xgm'ish:
> 
> `ds_cumint_reverse = ds.integrate(axis) - ds.cumint(axis)`

This would be a bit inefficient, because it effectively does the computation twice.

I would support adding a keyword `reverse=False` to all the cumsum / cumint operators. So you could do
```python
ds.cumint(axis, reverse=True)
```

jbusecke:
Good point. How do you think this could be implemented best? Flip the numpy/dask array before the computation? Or do it on the xarray level with `sortby(-dim)`?

rabernat:
> Flip the numpy/dask array before the computation?

This seems like the simplest, but it is not currently possible because we use xarray's cumsum method:
https://github.com/xgcm/xgcm/blob/03c165c225545e5bb4aa22eb08ad9a366b4879e3/xgcm/grid.py#L615-L616

> Or do it on the xarray level

This introduces some extra overhead with indexes that we don't really need.

The solution may be to refactor `grid.cumsum` to call cumsum at the numpy / dask level.

dcherian:
flip is just `.isel(dim=slice(0, -1, -1)`. I don't know how much overhead this will introduce relative to the overhead of refactoring working code :)

rabernat:
so we are talking about

```python
if reverse:
    da_rev = da.isel(dim=slice(0, -1, -1))
    da_cum = da_rev.cumsum(dim=dim).isel(dim=slice(0, -1, -1))
else:
    da_cum = da.cumsum(dim=dim)
```

jbusecke:
This seems easy enough to implement and test. Should we do that here or raise an issue upstream in xarray? I just did a quick search through their issue and couldnt find anything. @dcherian do you know if this would be a major thing? 
Given the propsed solution is literally 3 lines, I think we should just go ahead and implement it in xgcm for the time being?

dcherian:
You could open an issue about `flip`. 

do we need a function / kwarg for this or just a tip in the docstring?

dcherian:
> Or do you mean I should propose one?

:+1:

My question is, can we not expect users to do `grid.cumsum(da.isel(Z=slice(None, None, -1)), "Z")`). I guess the real issue is that `Z` is different for different variables and that makes it annoying (but easily fixed with cf-xarray!)

jbusecke:
> I guess the real issue is that Z is different for different variables and that makes it annoying (but easily fixed with cf-xarray!)

I mean that is sort of the core functionality of xgcm to deal with that 😝. So I think we should be able to handle it more elegantly.

jbusecke:
Closing this in favor of #256. There are still some issues to work out, any feedback over there would be very much appreciated.

Issue #150, opened 2019-09-17
grid.diff dropping coordinates


ALDepp:
Hi all,  I am wondering about the behaviour of grid.diff. 

For example, when I find the differences of a variable ds.DIA_IMPVF_TEMP, output:

<xarray.DataArray 'DIA_IMPVF_TEMP' (time: 1, z_w_bot: 62, nlat_t: 305, nlon_t: 1301)>
dask.array<shape=(1, 62, 305, 1301), dtype=float32, chunksize=(1, 62, 305, 1301)>
Coordinates:
  * z_w_bot  (z_w_bot) float32 1000.0 2000.0 3000.0 ... 574999.06 599999.06
    TLONG    (nlat_t, nlon_t) float64 dask.array<shape=(305, 1301), chunksize=(305, 1301)>
    TLAT     (nlat_t, nlon_t) float64 dask.array<shape=(305, 1301), chunksize=(305, 1301)>
  * time     (time) object 0052-01-16 05:17:48
  * nlon_t   (nlon_t) float64 0.5 1.5 2.5 3.5 ... 1.298e+03 1.3e+03 1.3e+03
  * nlat_t   (nlat_t) float64 0.5 1.5 2.5 3.5 4.5 ... 301.5 302.5 303.5 304.5
Attributes:
    grid_loc:      3113
    long_name:     TEMP Flux Across Bottom Face from Diabatic Implicit Vertic...
    cell_methods:  time: mean
    units:         degC cm/s

with grid.diff(ds.DIA_IMPVF_TEMP, axis="Z", boundary="extend")
 the coordinates TLONG and TLAT and the attributes are dropped. 
output:
<xarray.DataArray 'sub-bf9600c57ef24f7ffb1ba1512c2a9276' (time: 1, z_t: 62, nlat_t: 305, nlon_t: 1301)>
dask.array<shape=(1, 62, 305, 1301), dtype=float32, chunksize=(1, 1, 305, 1301)>
Coordinates:
  * time     (time) object 0052-01-16 05:17:48
  * z_t      (z_t) float32 500.0 1500.0 2500.0 ... 537500.0 562499.06 587499.06
  * nlat_t   (nlat_t) float64 0.5 1.5 2.5 3.5 4.5 ... 301.5 302.5 303.5 304.5
  * nlon_t   (nlon_t) float64 0.5 1.5 2.5 3.5 ... 1.298e+03 1.3e+03 1.3e+03

[...]

PR #186, opened 2020-03-11
init keep_coords implementation

apatlpo:
we are trying to address #150

### DIFF
--- a/doc/grids.rst
+++ b/doc/grids.rst
@@ -287,6 +287,17 @@ We can reverse the difference operation by taking a cumsum:
 Which is approximately equal to the original ``f``, modulo the numerical errors
 accrued due to the discretization of the data.
 
+By default, these grid operations will drop any coordinate that are not 
+dimensions. The keep_coords argument allow to preserve compatible coordinates.
+For example:
+
+.. ipython:: python
+
+    f2 = f+xr.Dataset(coords={'y': np.arange(1,3)})['y']
+    f2 = f2.assign_coords(h=f2.y**2)
+    print(f2)
+    grid.interp(fy, 'X', keep_coords=True)
+
 So far we have just discussed simple grids (i.e. regular grids with a single
 face).
 Xgcm can also deal with complex topologies such as cubed-sphere and
--- a/doc/grids.rst
+++ b/doc/grids.rst
@@ -296,7 +296,7 @@ For example:
     f2 = f+xr.Dataset(coords={'y': np.arange(1,3)})['y']
     f2 = f2.assign_coords(h=f2.y**2)
     print(f2)
-    grid.interp(fy, 'X', keep_coords=True)
+    grid.interp(f2, 'X', keep_coords=True)
 
 So far we have just discussed simple grids (i.e. regular grids with a single
 face).
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -681,6 +681,7 @@ class Axis:
             keep_coords,
         )
 
+    @docstrings.dedent
     def max(
         self,
         da,
@@ -816,7 +817,9 @@ class Grid:
             coordinates in ``ds``.
         metrics : dict, optional
             Specification of grid metrics
-
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
+            
         REFERENCES
         ----------
         .. [1] Comodo Conventions https://web.archive.org/web/20160417032300/http://pycomodo.forge.imag.fr/norm.html
@@ -1138,8 +1141,6 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1168,8 +1169,6 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1198,8 +1197,6 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1229,8 +1226,6 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1261,8 +1256,6 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
 
         Examples
         --------
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1354,9 +1354,7 @@ class Grid:
         ----------
         axis : str
             Name of the axis on which to act
-        %(neighbor_binary_func.parameters.no_f)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
+        %(grid_func.parameters)s
 
         Returns
         -------
@@ -1408,9 +1406,7 @@ class Grid:
         ----------
         axis : str, list of str
             Name of the axis on which to act
-        %(neighbor_binary_func.parameters.no_f)s
-        keep_coords : boolean, optional
-            Preserves compatible coordinates. False by default.
+        %(grid_func.parameters)s
 
         Returns
         -------
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -709,7 +709,7 @@ class Axis:
             vector_partner,
         )
 
-    def _wrap_and_replace_coords(self, da, data_new, position_to, keep_coords):
+    def _wrap_and_replace_coords(self, da, data_new, position_to, keep_coords=False):
         """
         Take the base coords from da, the data from data_new, and return
         a new DataArray with a coordinate on position_to.
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -220,6 +220,8 @@ class Axis:
             The value to use in the boundary condition with `boundary='fill'`.
         vector_partner : dict, optional
             A single key (string), value (DataArray)
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Returns
         -------
@@ -1136,6 +1138,8 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1164,6 +1168,8 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1192,6 +1198,8 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1221,6 +1229,8 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1251,6 +1261,8 @@ class Grid:
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Examples
         --------
@@ -1350,6 +1362,8 @@ class Grid:
         axis : str
             Name of the axis on which to act
         %(neighbor_binary_func.parameters.no_f)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Returns
         -------
@@ -1402,6 +1416,8 @@ class Grid:
         axis : str, list of str
             Name of the axis on which to act
         %(neighbor_binary_func.parameters.no_f)s
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Returns
         -------
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -653,6 +653,7 @@ class Axis:
         fill_value=0.0,
         boundary_discontinuity=None,
         vector_partner=None,
+        keep_coords=False,
     ):
         """
         Minimum of neighboring points on intermediate grid point.
@@ -675,6 +676,7 @@ class Axis:
             fill_value,
             boundary_discontinuity,
             vector_partner,
+            keep_coords,
         )
 
     def max(
@@ -685,6 +687,7 @@ class Axis:
         fill_value=0.0,
         boundary_discontinuity=None,
         vector_partner=None,
+        keep_coords=False,
     ):
         """
         Maximum of neighboring points on intermediate grid point.
@@ -707,6 +710,7 @@ class Axis:
             fill_value,
             boundary_discontinuity,
             vector_partner,
+            keep_coords,
         )
 
     def _wrap_and_replace_coords(self, da, data_new, position_to, keep_coords=False):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -532,6 +532,7 @@ class Axis:
         fill_value=0.0,
         boundary_discontinuity=None,
         vector_partner=None,
+        keep_coords=False,
     ):
         """
         Interpolate neighboring points to the intermediate grid point along
@@ -557,6 +558,7 @@ class Axis:
             fill_value,
             boundary_discontinuity,
             vector_partner,
+            keep_coords=keep_coords,
         )
 
     @docstrings.dedent
@@ -591,11 +593,11 @@ class Axis:
             fill_value,
             boundary_discontinuity,
             vector_partner,
-            keep_coords,
+            keep_coords=keep_coords,
         )
 
     @docstrings.dedent
-    def cumsum(self, da, to=None, boundary=None, fill_value=0.0):
+    def cumsum(self, da, to=None, boundary=None, fill_value=0.0, keep_coords=False):
         """
         Cumulatively sum a DataArray, transforming to the intermediate axis
         position.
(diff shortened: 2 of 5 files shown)


jbusecke:
Thanks for submitting the PR @apatlpo. As far as I can see your tests are failing because of code formatting. We use black to format our code. You can find instructions how to install black and automatically format in the [Contributors Guide](https://xgcm.readthedocs.io/en/latest/contributor_guide.html)

apatlpo:
hum, many tests failing.
I'll have to investigate what's going on.
it's unfortunately mayhem immediately, will look later today.

apatlpo:
I'd need your opinion about these two points:

- keep_coords is a kwarg everywhere except in `_wrap_and_replace_coords`

- I would need feedback on the core of the implementation:
https://github.com/xgcm/xgcm/blob/91d1cd2198ecbabbaf62b5172a90990481826213/xgcm/grid.py#L742-L750

Additionally, there are several spots where keywords arguments are provided as positional arguments, see for example:
https://github.com/xgcm/xgcm/blob/91d1cd2198ecbabbaf62b5172a90990481826213/xgcm/grid.py#L553-L562
while:
https://github.com/xgcm/xgcm/blob/91d1cd2198ecbabbaf62b5172a90990481826213/xgcm/grid.py#L187-L197
This looks odd to me, shouldn't it?

apatlpo:
pinging @jbusecke for the help

xgcm/grid.py
```diff
@@ -704,7 +709,7 @@ def max(
             vector_partner,
         )
 
-    def _wrap_and_replace_coords(self, da, data_new, position_to):
+    def _wrap_and_replace_coords(self, da, data_new, position_to, keep_coords):
```
jbusecke: Could you set a default right here (e.g. `keep_coords=False`), that could help with all the tests below.

xgcm/test/test_grid.py
```diff
@@ -187,12 +187,12 @@ def test_axis_wrap_and_replace_2d(periodic_2d):
     da_xg_yc = 0 * ds.XG * ds.YC + 1
 
     da_xc_yg_test = axis_objs["Y"]._wrap_and_replace_coords(
-        da_xc_yc, da_xc_yc.data, "left"
+        da_xc_yc, da_xc_yc.data, "left", False
```
jbusecke: With the above default you could avoid changing all this code.

xgcm/test/test_grid.py
```diff
@@ -187,12 +187,12 @@ def test_axis_wrap_and_replace_2d(periodic_2d):
...
     )
     assert da_xc_yg.equals(da_xc_yg_test)
 
     da_xg_yc_test = axis_objs["X"]._wrap_and_replace_coords(
-        da_xc_yc, da_xc_yc.data, "left"
+        da_xc_yc, da_xc_yc.data, "left", False
```
jbusecke: and here and so on...

xgcm/test/test_grid.py
```diff
@@ -631,6 +635,7 @@ def test_grid_ops(all_datasets):
                     da_cumsum = grid.cumsum(ds[varname], axis_name, boundary=boundary)
                     da_cumsum_ax = axis.cumsum(ds[varname], boundary=boundary)
                     assert da_cumsum.equals(da_cumsum_ax)
+                # test diff keep_coords?
```
jbusecke: I personally would write an additional test function that tests the behavior of keep_attrs on a high level, instead of testing this in all the other tests. That way we can be sure that the existing behavior is not modified.

apatlpo: could you please be a bit more specific about this last point, I am not following sorry

jbusecke: I think looking at the [tests for the metrics capability](https://github.com/xgcm/xgcm/blob/master/xgcm/test/test_metric_ops_multi_axis.py), could be helpful to get you started. I was talking about something along the lines of:
```

@pytest.mark.parametrize("funcname", ["diff", "interp", "min", "max", "integrate", "average", "cumint"]) # this should cover every operation that we have available
def test_keep_coords(all_datasets,funcname):
    grid = Grid(ds, coords=coords, metrics=metrics, periodic=periodic)
    func = getattr(grid, funcname) #grab the function from the grid object insead of typing out every instance of grid.diff, grid.interp....
    for axis_name in grid.axes.keys():
        result = func(ds.var, axis_name)
        result_w_coords = func(ds.var, axis_name, keep_coords=True)
        # some logic that asserts that `results` has no coords except the dimensions, and `results_w_coords` does in fact have all of the desired coordinates
```
The test datasets should have plenty of non-dim coordinates at different positions (the grid metrics).

This will make sure that we consistently can pass the argument to all available functions and that indeed the coordinates are preserved when we pass it. Let me know if this makes sense (the last message was written too late at night, apologies).

apatlpo: thanks a bunch for this enlightening guidance.
a81166c implements a first test

apatlpo:
Note that at the moment keep_coords is implemented for `diff`, `interp`, `min`and `max`but not for `integrate`, `average`, and `cumsum`. The reason being they call xarray functions directly and _wrap_and_replace_coords (where keep_coords is implemented).
Should I work toward adding this functionality to the latter functions as well?

xgcm/test/test_grid.py
```diff
@@ -663,3 +659,26 @@ def test_multi_axis_input(all_datasets, func, periodic, boundary):
...
+    ds, coords, metrics = datasets_grid_metric("C")
+    ds = ds.assign_coords(yt_bis=ds["yt"], xt_bis=ds["xt"])
+    grid = Grid(ds, coords=coords, metrics=metrics)
+    func = getattr(grid, funcname)
+    # print(ds)
+    for axis_name in grid.axes.keys():
+        coords = list(func(ds.tracer, axis_name).coords)
+        assert "xt_bis" not in coords
+        assert "yt_bis" not in coords
+        #
```
jbusecke: I like this way of testing that the dimension coordinates are in fact dropped, but I would suggest to borrow some logic from the actual implementation here and be even more explicit. E.g. `test_coords = [c for c in ds.coords if set(c.dims)==set(result.dims)]` where `results=func(ds.tracer, axis_name, keep_coords=True)`. The assert that all the test_coords are in results. 

I also think we should explicitly test for them being dropped if `keep_coords=False`, this will help with bookkeeping between the different methods.

xgcm/test/test_grid.py
```diff
@@ -663,3 +659,26 @@ def test_multi_axis_input(all_datasets, func, periodic, boundary):
             serial = getattr(grid, func)(serial, axis, boundary=boundary_axis)
         full = getattr(grid, func)(ds[varname], axes, boundary=boundary)
         xr.testing.assert_allclose(serial, full)
+
+
+# @pytest.mark.parametrize("funcname", ["diff", "interp", "min", "max", "integrate", "average", "cumint"])
+# integrate, average and cumsum call xarray sum method directly and
+# not _wrap_and_replace_coords
+@pytest.mark.parametrize("funcname", ["diff", "interp", "min", "max"])
```
jbusecke: I forgot `derivative`. Could you add it? See some more discussion about which functions to include in the main comments.

xgcm/grid.py
```diff
@@ -733,6 +742,17 @@ def _wrap_and_replace_coords(self, da, data_new, position_to):
...
                     coords[d] = da.coords[d]
 
+        # add compatible coords
+        if keep_coords:
+            # v0: compact but not readable version:
+            # coords.update(**{c: da[c] for c in da.coords
+            #                 if c not in coords and all([d in coords for d in da[c].dims])})
+            # v1: less compact but more readable version:
+            for c in da.coords:
+                if c not in coords and all([d in coords for d in da[c].dims]):
```
jbusecke: This could be reused for the test.

jbusecke: Also I think you are right, and this version is more readable. So I recommend going with this.

jbusecke:
First of all, this looks great already! 

I agree that we have to think about `integrate` and `average` differently. I would like to include them either way to keep track of the behavior of all available methods of grid. I am actually curious if they preserve the coords as is. In that case you could write a bit of logic to skip the assertion that the coords are dropped in these two cases (see my inline comments).

`cumsum` and `cumint` do actually call `_wrap_and_replace` eventually (they all go through the `axis.cumsum`, so they should be included.

Looking at how verbose this implementation has to be implemented has me thinking how to streamline/refactor the code (Ill try to take care of that, in a later PR). Having these detailed tests will be very valuable later on. So a big thanks for all your work.

xgcm/test/test_grid.py
apatlpo: This comments raises a doubt in mind that needs clarification.
My understanding of the functionality we are trying to implement is that all coordinates of the *DataArray* that is passed to xgcm functions such as `grid` and that are compatible with the output DataArray are transfered.
The test you indicate (`test_coords = [c for c in ds.coords if set(c.dims)==set(result.dims)]`) suggests you'd like all compatible coordinates from the original *Dataset* should be transfered. Is that what you think should happen?
If yes, I don't see how this can be possible as to my understanding all coordinates originally not compatible with the input DataArray are dropped immediately upon selection of the DataArray from the Dataset.

jbusecke: You are absolutely right. This was sloppy notation (I sketched the function without actually running anything). I think I meant to use `ds.tracer` instead of `ds`. Does that make more sense?
Apologies for the confusion.

apatlpo:
1a912ed tries to implement all your suggestions
The logic in _wrap_and_replace has been slightly lightened
you're right about `cumsum`, `cumint`, `derivative` of course, keep_coords is smoothly passed to these.
Passing keep_coords to `integrate` and `average` breaks with an error of the type:
```
xgcm/grid.py:1446: in average
    return weighted.sum(dim, **kwargs) / weight.sum(dim, **kwargs)
../../.miniconda3/envs/equinox/lib/python3.7/site-packages/xarray/core/common.py:46: in wrapped_func
    return self.reduce(func, dim, axis, skipna=skipna, **kwargs)
../../.miniconda3/envs/equinox/lib/python3.7/site-packages/xarray/core/dataarray.py:2235: in reduce
    var = self.variable.reduce(func, dim, axis, keep_attrs, keepdims, **kwargs)
../../.miniconda3/envs/equinox/lib/python3.7/site-packages/xarray/core/variable.py:1535: in reduce
    data = func(input_data, axis=axis, **kwargs)

axis = (0,), skipna = None, kwargs = {'keep_coords': False}, func = <function nansum at 0x119fc3440>, nanname = 'nansum'

    def f(values, axis=None, skipna=None, **kwargs):
        if kwargs.pop("out", None) is not None:
            raise TypeError(f"`out` is not valid for {name}")

        values = asarray(values)

        if coerce_strings and values.dtype.kind in "SU":
            values = values.astype(object)

        func = None
        if skipna or (skipna is None and values.dtype.kind in "cfO"):
            nanname = "nan" + name
            func = getattr(nanops, nanname)
        else:
            func = _dask_or_eager_func(name, dask_module=dask_module)

        try:
>           return func(values, axis=axis, **kwargs)
E           TypeError: nansum() got an unexpected keyword argument 'keep_coords'
```
Logic was added to tests in order to account for this.

xgcm/test/test_grid.py
```diff
@@ -658,3 +659,41 @@ def test_multi_axis_input(all_datasets, func, periodic, boundary):
...
+        "interp",
+        "min",
+        "max",
+        "integrate",
+        "average",
+        "cumsum",
+        "cumint",
+        "derivative",
+    ],
+)
```
jbusecke: It is probably not strictly necessary but could you parameterize the grid type as well?
```python
@pytest.mark.parametrize("gridtype", ["B", "C"])
def test_keep_coords(funcname, gridtype):
    ds, coords, metrics = datasets_grid_metric(gridtype)
```
That way we make sure that all cases are covered.

xgcm/test/test_grid.py
jbusecke: This looks great now. Very readable implementation.

jbusecke:
This looks good to me. @rabernat, could you give it a review before merging?

@apatlpo, dont forget to add an entry in the [whats new section of the docs](https://github.com/xgcm/xgcm/blob/master/doc/whats-new.rst) and it would be amazing if you could document the new feature with an example.

xgcm/test/test_grid.py
jbusecke: Awesome thanks.

apatlpo: done

apatlpo:
grid type is now parametrized in the test

whats-new was updated

I am not sure where to put an example. 
Do you think this modest implementation deserves one?

jbusecke:
The more we have in the doc, the better IMO. 
I would put a really small example behind `Which is approximately equal to the original f, modulo the numerical errors accrued due to the discretization of the data.` in the [Simple Grids Section](https://xgcm.readthedocs.io/en/latest/grids.html#core-grid-operations-diff-interp-and-cumsum).

jbusecke:
I was also just rereading this here: https://github.com/xgcm/xgcm/issues/150#issuecomment-601929988. 
Depending on the answer we should bump our xarray requirements.

apatlpo:
Ok, noted: doc + version

I just woke up in the middle of the night after a nightmare (unrelated to this PR) with a potentially interesting twist to the implementation as currently proposed by the PR. It would be easy I think to allow keep_coords to be a dataset (the original one the dataarray was taken from for examlpe) from which additional coords could be salvaged.
Do you think think could be a good idea?
It yes, I'll add this as well.

allright back to sleep ...

apatlpo:
also I was wondering, what about API, should I add a description of keep_coords in the docstring of `_neighbor_binary_func`?

jbusecke:
I was wondering about the late reply! 

> It would be easy I think to allow keep_coords to be a dataset (the original one the dataarray was taken from for examlpe) from which additional coords could be salvaged.  Do you think think could be a good idea?  It yes, I'll add this as well

I appreciate the idea but I think we want to keep this as easy as possible. If the user wants to add more specific coords, that is easy enough to do manually. 

I think the implementation as it is is good. 

The API is a great point! The keyword would have to added to all user facing functions. This currently means a lot of copy pasting (I am planning to address this in a later PR together with some doctests, but for now please add them manually.

apatlpo:
ok, I am not entirely convinced, the implementation being so straitghforward, but you're the captain onboard.

API and xarray version requirement have been updated.
I added an example. It's not particulary pretty but I cannot think about anything else at the moment.

I still copy below the extra implementation I was talking about for the records:
```
def _augment_coords(dims, coords, d):
    """ gather coords compatible with dims from d (xr.Dataset or DataArray)
    """
    for c in d.coords:
        if c not in coords and set(d[c].dims).issubset(dims):
            coords[c] = d[c]
    return coords
...
        # add compatible coords
        if keep_coords:
            coords = _augment_coords(dims, coords, da)
            if isinstance(keep_coords, (xr.Dataset, xr.DataArray)):
                coords = _augment_coords(dims, coords, keep_coords)  
```

doc/grids.rst
```diff
@@ -287,6 +287,17 @@ We can reverse the difference operation by taking a cumsum:
...
+By default, these grid operations will drop any coordinate that are not 
+dimensions. The keep_coords argument allow to preserve compatible coordinates.
+For example:
+
+.. ipython:: python
+
+    f2 = f+xr.Dataset(coords={'y': np.arange(1,3)})['y']
+    f2 = f2.assign_coords(h=f2.y**2)
+    print(f2)
+    grid.interp(fy, 'X', keep_coords=True)
```
jbusecke: Should be `grid.interp(f2, 'X', keep_coords=True)`?

jbusecke:
> ok, I am not entirely convinced, the implementation being so straitghforward, but you're the captain onboard.

To kind but if anyone is the captain its @rabernat 😁 . I think about it this way: If we implement it now and later decide to drop it, that is easy. The other way around requires significantly more effort. I thank you very much for taking the extra time right now. I really believe there is no such thing as too much docs.

It seems that your example in the docs in [failing](https://readthedocs.org/projects/xgcm/builds/10674748/), see my inline comment.

apatlpo:
oops, fixed !

xgcm/grid.py
```diff
@@ -670,6 +678,7 @@ def min(
             fill_value,
             boundary_discontinuity,
             vector_partner,
+            keep_coords,
         )
```
jbusecke: This is missing `@docstrings.dedent`, could you add it? Sorry for this. That must have gone missing some time ago...

apatlpo: you mean for `max`, correct?

xgcm/grid.py
```diff
@@ -1121,6 +1138,8 @@ def interp(self, da, axis, **kwargs):
         Parameters
         ----------
         %(grid_func.parameters)s
+        keep_coords : boolean, optional
```
jbusecke: Sorry for catching this now. Could you add this in the docstring of `grid_func`, that will get parsed here.

xgcm/grid.py
```diff
@@ -1149,6 +1168,8 @@ def diff(self, da, axis, **kwargs):
...
```
jbusecke: Same

xgcm/grid.py
```diff
@@ -1177,6 +1198,8 @@ def min(self, da, axis, **kwargs):
...
```
jbusecke: Same here.

xgcm/grid.py
```diff
@@ -1206,6 +1229,8 @@ def max(self, da, axis, **kwargs):
...
```
jbusecke: Here

xgcm/grid.py
```diff
@@ -1236,6 +1261,8 @@ def cumsum(self, da, axis, **kwargs):
...
```
jbusecke: Here

xgcm/grid.py
```diff
@@ -1335,6 +1362,8 @@ def derivative(self, da, axis, **kwargs):
         axis : str
```
jbusecke: Could you replace the parameter section with `%(grid_func.parameters)s`. I seem to have forgotten that.

xgcm/grid.py
```diff
@@ -1387,6 +1416,8 @@ def cumint(self, da, axis, **kwargs):
         axis : str, list of str
```
jbusecke: Could you replace this here with `%(grid_func.parameters)s` aswell. Sorry for the trouble.

xgcm/grid.py
jbusecke: Yip, sorry. The unfolding in github is quite annoying...

jbusecke:
Thanks for taking the time to iron out the kinks in the docstrings @apatlpo. I really appreciate the effort.

apatlpo:
no worries

rabernat:
This is fantastic. Thanks @apatlpo for the very useful PR and @jbusecke for the careful review. Sorry I was so slow to review this. I see no problems.

PR #188, opened 2020-03-20
Added conda-forge counter badge

jbusecke:
So that we can see how many folks download xgcm...its already > 7k


jbusecke:
@rabernat Im just going to merge this, since its small! Let me know what you think

PR #197, opened 2020-04-12
WIP: add xarray dataarray_accessor

rabernat:
This is something we have discussed for a while. Today I realized it was pretty simple to implement. (Famous last words.) So I had to give it a try.

Several people (e.g. @JiaweiZhuang) have commented that they would like to be able to use xgcm with an accessor. The challenge is figuring out which Grid object to use with any given xarray.DataArray.

What I did here was copy Dask's mechanism for finding a Client object to use for computation. I used [weakref](https://docs.python.org/3/library/weakref.html) to create a global set of Grid objects. When you access the `.grid` property of a DataArray, the accessor searches for a Grid that can align with the DataArray. The first one it finds is used for `da.grid.interp` and `da.grid.diff` methods.

Here's a simple example:

```python
import numpy as np
import xarray as xr
import xgcm

ds = xr.Dataset(coords={'x_c': (['x_c',], np.arange(1,10)),
                        'x_g': (['x_g',], np.arange(0.5,9))})

grid = xgcm.Grid(ds, coords={'X': {'center': 'x_c', 'left': 'x_g'}})
da = np.cos(ds.x_c)
da.grid.interp('X')
```

There are many ways this could go wrong, particularly if there are multiple Grid objects in existence. However, in my workflows I tend to only ever have one at a time. So perhaps it's a easy, simple solution that will improve our API.

### DIFF
--- a/xgcm/__init__.py
+++ b/xgcm/__init__.py
@@ -5,3 +5,4 @@ del get_versions
 
 from .grid import Grid, Axis
 from .autogenerate import generate_grid_ds
+from .accessor import GridAccessor
--- a/xgcm/accessor.py
+++ b/xgcm/accessor.py
@@ -1,0 +1,39 @@
+import xarray as xr
+
+from .grid import Grid, _get_global_grids
+
+
+class GridError(Exception):
+    pass
+
+
+def _find_compatible_global_grid(da):
+    global_grids = _get_global_grids()
+    for grid in global_grids:
+        try:
+            xr.align(grid._ds, da, join="exact")
+            return grid
+        except ValueError:
+            pass
+    raise GridError("No compatible global grid found.")
+
+
+@xr.register_dataarray_accessor("grid")
+class GridAccessor:
+    def __init__(self, xarray_obj):
+        self._da = xarray_obj
+        self._grid_obj = None
+        self.init()
+
+    def init(self, grid_obj=None):
+        """Initialize xgcm.Grid object for the dataset."""
+        if grid_obj:
+            self._grid_obj = grid_obj
+        else:
+            self._grid_obj = _find_compatible_global_grid(self._da)
+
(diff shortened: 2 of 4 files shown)


jbusecke:
Oh neat! I have definitely worked with many grid objects before, so I think this might be worth thinking about more. But perhaps there could be some input checks that would help sort out the “right” grid object (We discusses this in an earlier release...but the new github app has no autocomplete 😭)

rabernat:
> But perhaps there could be some input checks that would help sort out the “right” grid object

That's what the function `_find_compatible_global_grid` tries to do.

jbusecke:
I see. I think this is really cool (but have to check it out in practice more).

jbusecke:
Just linking #225 here, since it seems relevant to connect the discussions.

rabernat:
Let's keep this alive! Not stale!

lukelbd:
Just wanted to say this is a really cool project and having a `.grid` accessor would be *awesome* --  much more elegant IMO.

Maybe `ds.grid.init()` could be expanded to work as an alternative to the `Grid(ds, ...)` call. Or, the accessor could be made callable, with `ds.grid()` equivalent to `Grid(ds, ...)`. And to handle multiple grid objects, maybe the accessor could be optionally indexed (e.g. `ds.grid[N].interp(...)`), with the "first" / "default" grid automatically selected when no index is provided (e.g. `ds.grid.interp(...)` -- perhaps a warning could be raised when there is more than one option).

hdrake:
Closing as part of v1.0.0 triage (#733) — this has been dormant since 2020, and the internals it patched have since moved to the grid-ufunc architecture, so an accessor would need a fresh design on top of the current `Grid`. Happy to revisit in 1.x; feel free to reopen or file an issue if there's interest.

PR #198, opened 2020-04-16
example notebook for MOM6

raphaeldussin:
getting started with MOM6:

* grid definition
* vorticity example
* plotting


raphaeldussin:
notebook preview at 

https://github.com/raphaeldussin/xgcm/blob/example_nb_mom6/example_notebooks/MOM6_getting_started.ipynb

@rabernat do you want me to add it to the rtd ?

rabernat:
This is really great! Thanks!

Yes, I would prefer to have this in the actual docs. I think the `example_notebooks` folder needs to go away. However, we could merge this now and then handle that moving around as part of a broader refactor. We'd really like to have an example gallery with many different models, and we need to think about how to organize that.

@jbusecke - thoughts?

amit1rrr:
> notebook preview at
> 
> > https://github.com/raphaeldussin/xgcm/blob/example_nb_mom6/example_notebooks/MOM6_getting_started.ipynb

@raphaeldussin @rabernat Sorry to intrude but I noticed you are using notebook previews like above to review notebook PRs. I understand that it's hard to read Jupyter notebook diffs on GitHub.

You might want to checkout [ReviewNB](https://www.reviewnb.com/), code review tool for Jupyter notebooks. It's free for open source repositories. Provides visual diffs for notebooks & ability to comment on notebook cells.

jupyter_notebook_diff

jbusecke:
Thanks @raphaeldussin that looks awesome. I love the additional info on the grid, well done. 
I think it would be helpful to use this as a starting point. I think each model/notebook should have a little description of the 'quirks' of the grid  and then make all the necessary adjustments to create a ds/grid combo that can be run through identical code (e.g. a standard set of computations and plots). I would try to leave the latter part completely untouched to show that we can really work 'grid-agnostic' once all the metrics are set up correctly. The user then only modifies the comments and code up until the grid creation for their model. What do you think?

With regard to organization: I think we should just add another Section in the docs like "Working with xgcm in various models" and add each notebook + a guide on how to contribute a notebook there? @rabernat you have probably more experience in how we could make this cooler/sleeker. But i think this would do the trick?

raphaeldussin:
@jbusecke I kept the compute section small so I'm not redundant with the rest of the documentation. How many computations are you thinking about? I like vorticity since it's a simple yet very useful quantity. I'm not sure how standardized the plotting can be (global vs regional,...) but we can try to stay close to a template.

I like the idea of the examples section. maybe put the MITgcm example first then add new models?

let me know what I can do to help this move forward.

jbusecke:
I think every model for the examples should come with a proper coordinate, but maybe we should skip the cartopy step for this? Xarray can deal with 2d coordinates and that might take some of the complication out of it (although I love how your plot turned out!).

For calculations, I am thinking:

- Vorticity
- Divergence
- Streamfunction
- Someweighted average

This will show most of xgcms functionality and those are all very common operations. 

Thoughts?

raphaeldussin:
@jbusecke @rabernat I have moved the notebook into the doc. I'm ok with adding more calculations in the notebook. RN I don't really have the bandwidth. I was wondering if we could merge as is, and I'll get back to it later. I have a webinar next week and I'd like to tell about xgcm and MOM6.

jbusecke:
I think that is fine. I might have time over the weekend to refactor and create a template based on your notebook. So for that it would be actually helpful. Sorry for the delay on this. I have been very busy over at cmip6_preprocessing.

rabernat:
SGTM!

Issue #193, opened 2020-04-03
apply Grid.average to subset of data


miniufo:
I constructed a Grid object from a lat/lon Dataset `dset`.  When I tried the following codes, I found a huge difference that is totally unexpected:
```python
# ssta is a global 2D field in dset, subregion is the Nino 3.4 region
# unweighted mean
print(ssta.sel(subregion).mean(['yt_ocean', 'xt_ocean']).values)

# weighted mean
print(grid.average(ssta.sel(subregion), ['X', 'Y']).values)
```
The outputs are:
```python
# unweighted mean
[ 1.9443461  1.7437879  1.012453  ... -1.6583643 -1.8293577 -2.0875595]

# weighted mean
[ 0.02257358  0.02026701  0.01267836 ... -0.01836029 -0.02009966
 -0.02342124]
```

I know that these two ways of mean will yield slightly different results but here the difference is quite large.  So I would ask:
1. If the DataArray passed in is only a subset (subregion) of the full domain (as in this example), will the `Grid` object do the average properly?
2. Do other `Grid` functions (e.g., interp, integrate, cumsum etc.) take into account this along any dimension combination?

This is the info for the `dset`:
```python
In [44]: dset
Out[44]: 
<xarray.Dataset>
Dimensions:        (time: 1200, xt_ocean: 360, xt_ocean_left: 360, yt_ocean: 200, yt_ocean_left: 200)
Coordinates:
    geolat_t       (yt_ocean, xt_ocean) float32 -81.5 -81.5 ... 65.21032
    geolon_t       (yt_ocean, xt_ocean) float32 -279.5 -278.5 ... 79.98032
  * yt_ocean       (yt_ocean) float64 -81.5 -80.5 -79.5 -78.5 ... 87.5 88.5 89.5
  * xt_ocean       (xt_ocean) float64 -279.5 -278.5 -277.5 ... 77.5 78.5 79.5
  * time           (time) object 0101-01-16 12:00:00 ... 0200-12-16 12:00:00
  * xt_ocean_left  (xt_ocean_left) float64 -280.0 -279.0 -278.0 ... 78.0 79.0
  * yt_ocean_left  (yt_ocean_left) float64 -82.0 -81.0 -80.0 ... 87.0 88.0 89.0
    dxG            (yt_ocean, xt_ocean_left) float64 1.644e+04 ... 970.4
    dyG            (yt_ocean_left, xt_ocean) float64 nan nan ... 1.112e+05
[...]

Issue #182, opened 2020-02-27
grid.average with MOM6


raphaeldussin:
it works but there is a pitfall, see https://mom6-analysiscookbook.readthedocs.io/en/latest/03_spatial_operations.html#Using-xgcm

I wonder if raising a warning if the number of points summed in the numerator and denominator differ would be a good idea. Or add the option for an optional land/sea mask.

@rabernat  @jbusecke what do you think?

also I have some very basic examples of using xgcm with MOM6. I'd be happy to contribute an example notebook if @jbusecke  don't have one in prep already.

rabernat:
I think the points you raise are very useful and also quite general. Many models (like POP) have similar issues with inconsistent masking of grid variables. This might also be an issue in MIT: the data variables (e.g. `THETA`) are just 0 in the land, and the weights (e.g. `rA`) are defined everywhere. So a naive weighted mean could also give the wrong answer.

There are a couple of paths forward I can think of:
- We could have users define the metrics with the appropriate masks outside of xgcm, i.e. `area = metrics['areacello'] * metrics['wet']` and then pass these the xgcm's metrics. (Perhaps some preprocessing routines for common models could automate this.)
- We could add the notion of masks to xgcm's data model and automatically mask data where necessary. This could also come in handy for other things, like enforcing boundary conditions.

Thoughts?

jbusecke:
I noticed this too in some of my work. I hoped that we could take care of this upstream in [xarray](https://github.com/pydata/xarray/pull/2922), but it seems that might have stalled...

> We could have users define the metrics with the appropriate masks outside of xgcm, i.e. area = metrics['areacello'] * metrics['wet'] and then pass these the xgcm's metrics. (Perhaps some preprocessing routines for common models could automate this.)

I did wright some logic to do weighted averages [here](https://github.com/jbusecke/xarrayutils/blob/003e4f8a12aefc09714f7ecc184ac3c70956dc3c/xarrayutils/weighted_operations.py#L108), which we could modify. This will however make these operations slower. Perhaps it would be good to have it on by default (so a weighted mean over ocean masked properties with a global area is not biased for example), but for the user to be able to turn it off if they have properly preprocessed metrics?

>also I have some very basic examples of using xgcm with MOM6. I'd be happy to contribute an example notebook if @jbusecke don't have one in prep already.

Nothing specific in the pipeline yet (mostly worked with MOM5!), so I would be super happy if you could add some MOM6 examples.

rabernat:
[...]

Issue #111, opened 2018-07-30
 validation of input dimensions sizes


rabernat:
It is currently possible to create `Grid` and `Axis` objects whose sizes are not compatible with xgcm's data model.

This should not be allowed

```
>>> ds = xr.Dataset(coords={'x_c': (['x_c',], np.arange(1,10)),
                            'x_g': (['x_g',], np.arange(0.5,9))})
>>> ds
<xarray.Dataset>
Dimensions:  (x_c: 9, x_g: 9)
Coordinates:
  * x_c      (x_c) int64 1 2 3 4 5 6 7 8 9
  * x_g      (x_g) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5
Data variables:
    *empty*
>>> grid = Grid(ds, coords={'X': {'center': 'x_c', 'outer': 'x_g'}})
>>> grid
<xgcm.Grid>
X Axis (periodic):
  * center   x_c (9) --> outer
  * outer    x_g (9) --> center
```

According to our own [docs](http://xgcm.readthedocs.io/en/latest/grids.html#axes-and-positions), the `outer` position should have a length of 10. So this (and all similarly erroneous Grid / Axis creation operations) should raise errors.

jbusecke:
I think this is relevant to #208. Ill close it and paste the example over there, ok?

jbusecke:
This definitely needs fixing.

PR #204, opened 2020-06-13
implemented xr.weighted, local tests pass

jbusecke:
I have implemented the use of `xr.weighted` in the `grid.average` function. I think this takes care of the issues mentioned in #193 and #182. 

Basically this now takes care of masking out weight values where the data is missing, to avoid the scenario in #182 , where we have a different number of valid values in each. 

Quick example:
```python
import xarray as xr
import numpy as np
from xgcm import Grid

# a minimal example
x = np.arange(10)
data= xr.DataArray(np.ones(10),dims='x', coords={'x':x})
weights = data * 30
ds = xr.Dataset({'data':data})
ds = ds.assign_coords(weights=weights)
ds
```
image

The current version behaves like this:
```python
# create an xgcm grid
grid = Grid(ds, coords={'X':{'center':'dim_0'}}, metrics={'X':['weights']})

# average
print(grid.average(ds.data, 'X'))

# now lets introduce a missing value in the data
data[6:8] = np.nan
# average
print(grid.average(ds.data, 'X'))
```
giving:
```
<xarray.DataArray ()>
array(1.)
<xarray.DataArray ()>
array(0.8)
```
This is obviously not right.

With the proposed changes the same code gives:
```
<xarray.DataArray ()>
array(1.)
<xarray.DataArray ()>
array(1.)
```
which is usually the expected output for such an operation.

It also happens to work with smaller subsets of data:
```python
# average
print(grid.average(ds.data[0:7], 'X'))
```
giving
```
<xarray.DataArray ()>
array(1.)
```
@miniufo, could you check if this addresses your issue in #193 ?

We should still have a broader conversation about how to deal with subsets, but personally I think `grid.average` is often used after the other functions are applied to a global dataset (at least in my personal experience). So having the ability to e.g. calculate the x-gradient and then select a box and average it only over that box, is already a big improvement. 

Thoughts?


miniufo:
>@miniufo, could you check if this addresses your issue in #193 ?

Glad you work this out and it is a good start.  But I'm a little confused.  Do I need to explicitly declare a new DataArray `weights`? Or just convert a field (e.g., SST) into `weights` (0 when SST is nan and 1 otherwise)?  I am expecting that `xgcm` will do this for users (hopeless lazy guy here...)

Do I have to update the whole `xgcm` or just modified the `grid.average` function as in this commit to take effect?  I am not so sure how to test the new codes when it is in a branch that is not merged into master.

jbusecke:
You should not have to do anything else than pulling/installing this pull request. In a conda environment you can just do `pip install git+https://github.com/xgcm/xgcm.git@refs/pull/204/head` (make sure this is a separate env from your main one, so that you can ditch it, if things fail).

Once that is done, execute your example from #193 again, and hopefully it just works correctly. No additional input should be required from your side. Thanks for the the help.

dcherian:
Alternatively if you have an editable install of xgcm (`pip install -e`), you can do `gh pr checkout 204` in the xgcm root directory (https://cli.github.com/)

xgcm/grid.py
```diff
@@ -1431,24 +1431,21 @@ def average(self, da, axis, **kwargs):
...
+            Additional arguments passed to `xarray.DataArray.weighted.mean`
 
 
         Returns
         -------
         da_i : xarray.DataArray
             The averaged data
         """
-
+        da = da.copy()
```
rabernat, line 1442: why do we need to copy?

jbusecke: Hmm probably we dont.

jbusecke:
One thing I noticed just now, is that we probably need at least a warning if the alignment results in an empty array. I was just beating my head against the wall until I realized I was using a grid from another model. Just dropping it here, so I dont forget.

jbusecke:
#111  is relevant to this discussion aswell. 

I think either way we need some validation of the input dimensions.

jbusecke:
I messed this git history up beyond repair (at least for me). Decided to start fresh in #259 and closing this.

Issue #215, opened 2020-09-01
Add test and fix typo in fill value processing in `grid.py`


jbusecke:
The refactoring of the code linting in #214 uncovered a minor bug in the way [fill values are passed to an axis](https://github.com/xgcm/xgcm/pull/214#issuecomment-685112082).

This should be fixed asap.

jbusecke:
Closed via #218

PR #218, opened 2020-09-02
Bugfix: axis fillval parsing from dict input

jbusecke:
Fix for #215


jbusecke:
@dcherian does this look ok to you? I added a very small test to expose the issue and then a fix to `grid.py`.

dcherian (approved):
Thanks for cleaning up my mess.

PR #233, opened 2020-09-14
Add versioneer and setup.py to codecov ignore

jbusecke:
- [ ] Passes `isort . && black . && mypy . && flake8`

I noticed that our code coverage is very low and it seems this is mostly a result of codecov checking both `setup.py` and `versioneer.py`

image

This PR adds both files to the ignore in `codecov.yml`. Any thoughts on this?


dcherian:
Great  :+1:

jbusecke:
If nobody is against this, ill merge this soon.

codecov.yml
```diff
@@ -13,4 +13,8 @@ coverage:
         target: 95
         informational: true
     patch: off
-    changes: off
\ No newline at end of file
+    changes: off
+    
+ignore:
+  - "setup.py"
```
andersy005, line 19: You may also want to exclude `tests` as well:

```suggestion
  - "setup.py"
  - "xgcm/tests/*.py"
```

rabernat (approved):

codecov.yml
andersy005: Never mind :) Just found out that tests are excluded in `.coveragerc`. How about consolidating codecov config files into one file by moving the contents of `.coveragerc` into `codecov.yml`?

jbusecke: Thanks a lot for the suggestion. I have moved the ignore paths over and deleted `.coveragerc`, I did not find equivalents to the other options [in the codecov docs](https://docs.codecov.io/docs/codecovyml-reference), do you have an idea if I need those?

I mean these lines to be specific:
```

[run]
branch = True

[report]
exclude_lines =
    if self.debug:
    pragma: no cover
    raise NotImplementedError
    if __name__ == .__main__.:
ignore_errors = True
```

andersy005: It's my understanding that any line that ends with `# pragma: no cover` is automatically excluded by codecov and as a result you don't need to include it in config file (please correct me if I am wrong or you are getting different behavior).

Regarding `ignore_errors`,  what does it do? I've never had to use it.

jbusecke: As far as I can tell it seems to work on codecov. I would go ahead and merge.

andersy005: Sounds good!

jbusecke:
Thanks for the help @andersy005

Issue #207, opened 2020-07-06
`docrep` usage: refactor and addition to user guide


jbusecke:
We are using `docrep` in several instances of the core code to avoid copy & paste repetition of docstrings, but some other docstrings are still manually pasted. 

I think we should refactor the code to use docrep to the full extent possible. At the same time it would be good to give a 'how-to-docrep' with some specific examples, since the usage is not really trivial (at least not to me). This would help future contributors.

#210 showed further issues with docrep, prompting a quick fix in #211 which involved pinning an old version of docrep. The refactor should be done with the new version and squash those errors.

## TODO:
- Are there options to enforce better doc style within the CI?

jbusecke:
I think a larger docstring refactor should include an effort to make the kwarg handling of Axis/Grid methods more transparent (see #250 for details).

PR #250, opened 2020-09-28
Bugfix 224

### LINKED ISSUE #224: Globally set `fill_value` set on grid object is not propagated.
I just discovered what I think is a bug:

Consider this simple example:
```
from xgcm import Grid
import xarray as xr
import numpy as np

z = np.arange(3)
z_right = z+0.5
ds = xr.DataArray(np.ones_like(z)*1e5, dims=['z'], coords={'z':z}).to_dataset(name='data')
ds = ds.assign_coords({'z_right':z_right})

grid = Grid(ds, coords={'Z':{'center':'z', 'right':'z_right'}},
            boundary='fill', fill_value=np.nan, periodic=False)
grid.interp(ds.data, 'Z')
```
This results in:

image

Which is not what I would expect (the edge value should be padded with nan) and is not equivalent to the 'old-school' manual way:
```
grid.interp(ds.data, 'Z',boundary='fill', fill_value=np.nan)
```
image

I think the issue is that `fill_value=0.0` is hardcoded in all axis methods as default. See [here](https://github.com/xgcm/xgcm/blob/9c83d5368193237d80f7282f94a76561e0711c51/xgcm/grid.py#L614) for example.

We should add a test that checks that results are the same if you specify `boundary` and/or `fill_value` via grid object or in the axis method. 

Pinging @dcherian

jbusecke:
- [x] Closes #224 
 - [x] Tests added
 - [x] Passes `black . `


jbusecke:
Ok so I have patched this behavior, but the handling of kwargs throughout the module is kind of a mess still. 
The 'default' value is set in many spots (like [here](https://github.com/xgcm/xgcm/blob/2780d1ff0facc43138a77f0960d2fbf35c7c7341/xgcm/grid.py#L301-L325) and [here](https://github.com/xgcm/xgcm/blob/2780d1ff0facc43138a77f0960d2fbf35c7c7341/xgcm/grid.py#L456-L476)), where I am wondering if it should be a kwarg (Id suggest making it an argument).
It will be a larger undertaking to refactor all of this, but I think is should be a medium high priority, which will enable new contributors to understand the code much easier.

xgcm/test/test_metrics_ops_single_axis.py
```diff
@@ -71,6 +71,43 @@ def test_weighted_metric_multi_axis(
...
+@pytest.mark.parametrize("fill_value", [0, 10, None])
+def test_boundary_global_input(funcname, boundary, fill_value):
+    """Test that globally defined boundary values result in
+    the same output as when the parameters are defined on either
+    the grid or axis methods
+    """
+    ds, coords, metrics = datasets_grid_metric("C")
+    axis = "X"
+
+    # Test results by globally specifying boundary on grid object
```
dcherian, line 88:
```suggestion
    # Test results by globally specifying fill value on grid object
```

xgcm/test/test_metrics_ops_single_axis.py
```diff
@@ -71,6 +71,43 @@ def test_weighted_metric_multi_axis(
...
+        coords=coords,
+        metrics=metrics,
+        periodic=False,
+        boundary=boundary,
+        fill_value=fill_value,
+    )
+    func_global = getattr(grid_global, funcname)
+    global_result = func_global(ds.tracer, axis)
+
+    # Test results by manually specifying boundary on grid method
```
dcherian, line 100:
```suggestion
    # Test results by manually specifying fill value on grid method
```

dcherian (approved):
LGTM.

Looks like `_get_neighbor_binary_func` is handling the default `boundary` and `fill_value` case which seems low level enough.

jbusecke:
It also gets set in the Axis/Grid `_init_`, so I feel this is confusing in the long term. But yes. Lets add this fix and then discuss further refactoring in another issue (I added some text to #207 to remind us/me).

PR #259, opened 2020-10-06
Fix missing values handling average

jbusecke:
- [x] Replaces #204 
 - [x] Tests added
 - [x] Passes `black . `

I totally messed up #204 and since the changes are minor, I decided to start over fresh. 

This PR replaces our internal logic for the weighted average with xarrays `weighted`. This prevents errors, which arise when the processed data has missing values which are not found in the metric needed for the operation. I have turned my simple example into a [test](https://github.com/jbusecke/xgcm/blob/ccf95f0cda304f9afd486dd9d14718944a6ca273/xgcm/test/test_metrics_ops_single_axis.py#L111-L129), which failed with the current version.

As mentioned in #204 this also allows to compute averages on subsets of the dataset (due to `weighted`'s internal alignment). This addresses #193, but I would like to treat is as a bugfix only, since we still need to discuss a more consistent handling of subsets with all available functions.

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1712,24 +1712,21 @@ class Grid:
         axis : str, list of str
             Name of the axis on which to act
         **kwargs: dict
-            Additional arguments passed to `xarray.DataArray.sum`
-
+            Additional arguments passed to `xarray.DataArray.weighted.mean`
 
         Returns
         -------
         da_i : xarray.DataArray
             The averaged data
         """
-
+        da = da.copy()
         weight = self.get_metric(da, axis)
-        weighted = da * weight
-        # TODO: We should integrate xarray.weighted once available.
+        weighted = da.weighted(weight)
 
         # get dimension(s) corresponding
         # to `da` and `axis` input
         dim = self._get_dims_from_axis(da, axis)
-        # do we need to pass kwargs?
-        return weighted.sum(dim, **kwargs) / weight.sum(dim, **kwargs)
+        return weighted.mean(dim, **kwargs)
 
     def transform(self, da, axis, target, **kwargs):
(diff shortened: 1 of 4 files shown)


jbusecke:
Also here I dont think we need to point to this in the `whats-new.rst`? Thoughts?

xgcm/grid.py
```diff
@@ -1712,24 +1712,21 @@ def average(self, da, axis, **kwargs):
...
-
+            Additional arguments passed to `xarray.DataArray.weighted.mean`
 
         Returns
         -------
         da_i : xarray.DataArray
             The averaged data
         """
-
+        da = da.copy()
```
dcherian, line 1722: should be unnecessary?

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -50,12 +51,12 @@ def test_bgrid(self, funcname, boundary, periodic):
             expected = _expected_result(
                 ds.tracer, ds[metric_name], grid, dim, axis, funcname, **kwargs
             )
-            assert new.equals(expected)
+            xr.testing.assert_allclose(new, expected)
```
dcherian, line 54: shall we do `from xarray.testing import assert_allclose` at the top?

xgcm/grid.py
jbusecke: yes. Thanks. I removed it.

xgcm/test/test_metric_ops_multi_axis.py
jbusecke: Done

dcherian (approved):

rabernat (approved):
Thanks Julius!

Issue #238, opened 2020-09-16
Adding basic benchmarking to CI [FEATURE]


jbusecke:
I think it would be helpful to have an additional testing module, where we benchmark some operations with xgcm. 

I am thinking of having a few test functions that operate on one 'sizeable' synthetic test fixture and apply simple operations on it. I think these should be relatively high level actions like computing a derivative or transforming vertical coordinates. This way we at least have a rough idea of how changes we make impact the performance.

I have found [pytest-benchmark](https://github.com/ionelmc/pytest-benchmark) and some related [github action](https://github.com/rhysd/github-action-benchmark), that seem to give us the functionality to run these relatively elegantly and track results in the CI.

I am completely new to this, but I am sure we need to fiddle around to ensure that these tests are always run on e.g. the same hardware and other factors to ensure comparability...

If anyone here has experience with this sort of testing I would greatly appreciate input on both the framework (pytest-benchmark) and the implementation.

rabernat:
[Airspeed Velocity](https://asv.readthedocs.io/) is the standard tool used in the scientific python ecosystem.

jbusecke:
That looks very cool. From a quick read I have several questions: 

1. We need a webserver to host the results. Where would we put them ideally? I guess we could just make a github-pages in the xgcm orgnization?
2. We need one or several machines with consistent hardware to run these tests. I have been looking at some of the benchmarks of [dask/pandas/xarray and co](https://pandas.pydata.org/speed/), but am pretty clueless how to set this up.

Is there any guide out there how to approach this best? Pinging @hamman, since he seems to have been involved in setting this up for [xarray](https://github.com/pydata/xarray/pull/1457) a while back.

jhamman:
I did do some of the setup on the xarray side but, IIRC, @TomAugspurger holds the keys to the Pandas/Xarray/Dask Airspeed Velocity hardware and did the hardware configuration for us. You can see everything I did in the Xarray repo. As I understand it, there is a dedicated machine that belongs to the Pandas project for this somewhere (maybe in Tom's garage or something).   I would think it possible to set something similar up on a university server somewhere.

jbusecke:
> As I understand it, there is a dedicated machine that belongs to the Pandas project for this somewhere (maybe in Tom's garage or something)

😂 Thats awesome actually! 

But you think this is not feasible to run anywhere in the cloud?

TomAugspurger:
When setting this up a few years ago, we were told that trying to detect
[...]

PR #266, opened 2020-10-09
Adapt the internal transform code to accomodate missing values for target data in interpolation

### LINKED ISSUE #261: Unexpected transform behavior for decreasing target_data with missing values [BUG]
I just ran into an issue with the transform module, when I tried to transform data onto a `target_data` set that has missing values and decreasing values along the axes:

For a simple test dataset with decreasing values along the `Z` axis I get the expected result, which is a value identical to the `target`, since I am interpolating the array onto itself.

```python
ds = xr.DataArray(np.array([6,5,4,3,2,1]), dims=['z'], coords={'z':np.array([1,2,3,4,5,6])}).to_dataset(name='test')
ds
grid = Grid(ds, coords={'Z':{'center':'z'}}, periodic=False)
expected = grid.transform(ds.test, 'Z', [4.0], target_data=ds.test)
expected
```
image

But the results change, once I introduce nans at the beginning and end:
```python
# now introduce nans at the beginning and end
ds['test_nan'] = xr.DataArray(np.array([np.nan,5,4,3,2,np.nan]), dims=['z'], coords={'z':np.array([1,2,3,4,5,6])})
result = grid.transform(ds.test, 'Z', [4.0], target_data=ds.test_nan)
result
```
image

The `numpy.interp()` itself is able to handle nans in the data, but needs the x values (target_data in our context) to be increasing, which is why we implemented a simple internal [`flipping` mechanism](https://github.com/xgcm/xgcm/blob/b2fc97e29c6e262f69dc699d4c31dd2ad3ee53fc/xgcm/transform.py#L21-L24).

I think nans on the edge of the array circumvent the data to be flipped, because it only checks the first/last values. We should either remove missing values from the input in the core algo, or make the flipping decision based on the first/last non-nan value rather then the first/last value of the input array. 

My preferred solution would be the latter along the lines of:
```python
# if last theta value is smaller than first, assume the profile is monotonically decreasing and flip
theta_sign_test = theta[~np.isnan(theta)]
if theta_sign_test[-1] < theta_sign_test[0]:
    theta = np.flip(theta)
    phi = np.flip(phi)
```

Thoughts on whether this will affect performance negatively or if there is a more clever trick to do this @rabernat @dcherian ?

jbusecke:
- [x] Closes #261 
 - [x] Tests added
 - [x] Passes `black


xgcm/transform.py
```diff
@@ -18,8 +18,9 @@
     nopython=True,
 )
 def _interp_1d_linear(phi, theta, target_theta_levels, mask_edges, output):
-    # if last theta value is smaller than first, assume the profile is monotonically decreasing and flip
-    if theta[-1] < theta[0]:
+    # rough check if the data is decreasing with depth. If that is the case, flip.
+    theta_sign_test = theta[~np.isnan(theta)]
+    if theta_sign_test[-1] < theta_sign_test[0]:
```
rabernat, line 23: I don't understand how these checks can possibly be dask friendly. At this point, `theta` is still ND, no? Can you talk me through what you think is happening here?

jbusecke: I might be understanding the way gufuncs work, wrongly. I thought that we can basically treat 'theta' as 1D at this point, since the function is mapped onto 1D columns of the data (I basically thought the numba wrapper implements a very efficient for loop for every 1D column). But it seems that is not the case?
Do you know how to make `all()` work in numba? That way we could use your above suggestion.

rabernat: Ah sorry, I was confused. Here we are inside the gufunc. You are not misunderstanding. Your code makes total sense.

Now my only question is: how expensive is this? What does it cost us to do this check all the time.

dcherian: This should be fine since this is the bit that is vectorized.

re: all in numba: use `np.all`: https://numba.readthedocs.io/en/stable/developer/autogen_numpy_listing.html#listings-for-numpy

rabernat: It's vectorized, but it still takes CPU time. How much? How much slower is the function with / without this check?

The reason I ask is that we might prefer to have a flag like `check=True` that we can turn on / off to bypass.

jbusecke: I like that idea. I'd say we leave it on by default, and add a short section on how to tune performance (e.g. turn this check off if your data doesn't need to be flipped)?

Would be really nice to have #238 to be able to judge how this affects performance.

rabernat: We don't need #238 to judge how this affects performance.

We can just manually time something with / without it and report the answer right here in this thread.

jbusecke: Will do that.

jbusecke: OK I have come up with a simple synthetic test sample:
```python
import xarray as xr
import numpy as np
from xgcm import Grid

# create a dummy array
nx = 1000
ny = 1000
nz = 50
data = xr.DataArray(np.random.rand(nx, ny, nz), dims=['x','y','z'])

target = xr.DataArray(np.arange(nz), dims=['z']).expand_dims(x=nx, y=ny)
# add some noise to target
target = target + 0.2 * xr.DataArray(np.random.rand(nx, ny, nz), dims=['x','y','z'])

ds = xr.Dataset({'data':data, 'target':target})

grid = Grid(ds, coords={'Z':{'center':'z'}}, periodic=False)
```

I have quickly hacked together a bypass switch for the checks and timed the transform
```python
%%timeit 
grid.transform(ds.data, 'Z', np.arange(2,23,1.4), target_data=ds.target)
```
```
1.58 s ± 41.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
%%timeit 
grid.transform(ds.data, 'Z', np.arange(2,23,1.4), target_data=ds.target, bypass_checks=True)
```
```
1.22 s ± 82.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So there is a performance penalty, in this case ~30%. 

I think this is not negligible, so we should probably build this into the actual method? Id still say we should keep the checks on by default and add a 'performance tweaking' section to the docs.

Thoughts?

jbusecke: Just a quick ping to @rabernat. Id go ahead and implement this next week, unless you think it isn't a good idea?

jbusecke:
Why on earth is the `transform` page gone? It shows up in the [RTD build log](https://readthedocs.org/projects/xgcm/builds/12228852/). I don't get this at all...

jbusecke:
I tried to build the docs locally and they work fine. Not sure if we should try to merge and see if this is some sort of quibble with the RTD PR preview?

jbusecke:
It would be lovely to have this as a feature in some of my latest workflows. Just pinging @rabernat + @dcherian  for a review.

rabernat:
Great work Julius!

PR #267, opened 2020-10-09
Better handling of `target_data` input without name in `grid.transform`

### LINKED ISSUE #260: [BUG] `transform` error when `target_data` array has no name
Here is a minimal example

```
import numpy as np
import xarray as xr
from xgcm import Grid
ds = xr.Dataset({"dat1": (["z_c"], np.ones(9)),
        "dat2": (["z_c"], np.linspace(0,20,9))},
                coords={'z_c': (['z_c',], np.arange(1,10)), 'z_l': (['z_l',], np.arange(0.5,9)),'z_o': (['z_o',], np.arange(0.5,10))})
grid = Grid(ds, periodic=False, coords={'Z': {'center': 'z_c', 'outer': 'z_o'}})
grid.transform(ds.dat1, target = np.arange(0,20,2), 
               target_data=ds.dat2.rename(None), method='conservative', mask_edges=True, suffix='_transformed', axis = 'Z')
```

It throws an error: `TypeError: dimension None is not a string`

Obviously, no-one is directly dropping the name of the dataarray, but the name will be dropped if e.g. density is calculated from temp and salinity using jmd95. 

This may not be a super common problem but it was kind of hard to debug so I'd like it if the error was more informative (or if it didn't happen at all).

I'm running on AWS.

jbusecke:
- [x] Closes #260 
 - [x] Tests added
 - [x] Passes `black . `

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -883,7 +883,7 @@ class Axis:
                     " If the additional dimensions are associated witha staggered axis, "
                     "use grid.interp() to move values to other grid position. "
                     "If additional dimensions are not related to the grid (e.g. climate "
-                    "model members or similar, use xr.broadcast() before using transform."
+                    "model ensemble members or similar), use xr.broadcast() before using transform."
                 )
 
         def _parse_target(target, target_dim, target_data_dim, target_data):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -865,7 +865,7 @@ class Axis:
             """Handle target_data input without a name"""
             if target_data.name is None:
                 warnings.warn(
-                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
+                    "Input`target_data` has no name, but we need a name for the transformed dimension. The name `TRANSFORMED_DIMENSION` will be used. To avoid this warning, call `.rename` on `target_data` before calling `transform`."
                 )
                 target_data.name = "_UNNAMED_"
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -877,9 +877,9 @@ class Axis:
             target_da_other_dims = set(target_da.dims) - set(self.coords.values())
             if not target_da_other_dims.issubset(da_other_dims):
                 raise ValueError(
-                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}]" \
-                    "in `target_data` not found in `da`. This could mean that the target " \
-                    "array is not on the same position along other axes." \
+                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}]"
+                    "in `target_data` not found in `da`. This could mean that the target "
+                    "array is not on the same position along other axes."
                     " Use grid.interp() to align the arrays."
                 )
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -869,7 +869,7 @@ class Axis:
                 )
                 da.name = "_UNNAMED_"
 
-        def _check_target_alignment(da, target_da):
+        def _check_target_alignment(target_da):
             # check alignment for all other axes, to avoid broadcasting enourmous arrays
             # to do this on the axis level we will simply check if target_data has some coordinates
             # that are not found in da (excluding all coordinates that belong to the current axis)
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -877,7 +877,10 @@ class Axis:
             target_da_other_dims = set(target_da.dims) - set(self.coords.values())
             if not target_da_other_dims.issubset(da_other_dims):
                 raise ValueError(
-                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}] in `target_data` not found in `da`. This could mean that the target array is not on the same position along other axes. Use grid.interp() to align the arrays."
+                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}]" \
+                    "in `target_data` not found in `da`. This could mean that the target " \
+                    "array is not on the same position along other axes." \
+                    " Use grid.interp() to align the arrays."
                 )
 
         def _parse_target(da, target, target_dim, target_data_dim, target_data):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -861,7 +861,26 @@ class Axis:
                 "`transform` can only be used on axes that are non-periodic. Pass `periodic=False` to `xgcm.Grid`."
             )
 
-        def _parse_target(target, target_dim, target_data_dim, target_data):
+        def _target_data_name_handling(target_data):
+            """Handle target_data input without a name"""
+            if target_data.name is None:
+                warnings.warn(
+                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
+                )
+                da.name = "_UNNAMED_"
+
+        def _check_target_alignment(da, target_da):
+            # check alignment for all other axes, to avoid broadcasting enourmous arrays
+            # to do this on the axis level we will simply check if target_data has some coordinates
+            # that are not found in da (excluding all coordinates that belong to the current axis)
+            da_other_dims = set(da.dims) - set(self.coords.values())
+            target_da_other_dims = set(target_da.dims) - set(self.coords.values())
+            if not target_da_other_dims.issubset(da_other_dims):
+                raise ValueError(
+                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}] in `target_data` not found in `da`. This could mean that the target array is not on the same position along other axes. Use grid.interp() to align the arrays."
+                )
+
+        def _parse_target(da, target, target_dim, target_data_dim, target_data):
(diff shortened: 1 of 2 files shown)


jbusecke:
I dont think this needs an addition to the `whats-new.rst`?

xgcm/grid.py
```diff
@@ -861,7 +861,26 @@ def transform(
...
-        def _parse_target(target, target_dim, target_data_dim, target_data):
+        def _target_data_name_handling(target_data):
+            """Handle target_data input without a name"""
+            if target_data.name is None:
+                warnings.warn(
+                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
+                )
+                da.name = "_UNNAMED_"
+
+        def _check_target_alignment(da, target_da):
```
dcherian, line 872:
```suggestion
        def _check_target_alignment(target_da):
```

to be consistent with the other nested functions.

xgcm/grid.py
```diff
@@ -861,7 +861,26 @@ def transform(
...
+            # check alignment for all other axes, to avoid broadcasting enourmous arrays
+            # to do this on the axis level we will simply check if target_data has some coordinates
+            # that are not found in da (excluding all coordinates that belong to the current axis)
+            da_other_dims = set(da.dims) - set(self.coords.values())
+            target_da_other_dims = set(target_da.dims) - set(self.coords.values())
+            if not target_da_other_dims.issubset(da_other_dims):
+                raise ValueError(
+                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}] in `target_data` not found in `da`. This could mean that the target array is not on the same position along other axes. Use grid.interp() to align the arrays."
```
dcherian, line 880: I ran in to this yesterday. `xr.broadcast` or `DataArray.broadcast_like(target_data)` is a solution for some cases.

Also, lets add some newlines in this string :)

xgcm/grid.py
```diff
@@ -924,11 +934,9 @@ def _check_target_alignment(da, target_da):
                 )
 
             target, target_dim, target_data = _parse_target(
-                target, target_dim, target_data_dim, target_data
+                da, target, target_dim, target_data_dim, target_data
```
dcherian, line 937:
```suggestion
                target, target_dim, target_data_dim, target_data
```

xgcm/grid.py
jbusecke: I dont think this works. We need `da` in `_check_target_alignment`.

xgcm/grid.py
jbusecke: We actually want to avoid broadcasting on purpose. @rabernat ran into an example where the values were not properly aligned and things just blew up.

xgcm/grid.py
```diff
@@ -861,7 +861,29 @@ def transform(
...
+                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
+                )
+                da.name = "_UNNAMED_"
+
+        def _check_target_alignment(da, target_da):
+            # check alignment for all other axes, to avoid broadcasting enourmous arrays
+            # to do this on the axis level we will simply check if target_data has some coordinates
+            # that are not found in da (excluding all coordinates that belong to the current axis)
+            da_other_dims = set(da.dims) - set(self.coords.values())
+            target_da_other_dims = set(target_da.dims) - set(self.coords.values())
```
jbusecke, line 877: Here we need `da`

xgcm/grid.py
```diff
@@ -861,7 +861,29 @@ def transform(
...
+                warnings.warn(
+                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
+                )
+                da.name = "_UNNAMED_"
+
+        def _check_target_alignment(da, target_da):
+            # check alignment for all other axes, to avoid broadcasting enourmous arrays
+            # to do this on the axis level we will simply check if target_data has some coordinates
+            # that are not found in da (excluding all coordinates that belong to the current axis)
+            da_other_dims = set(da.dims) - set(self.coords.values())
```
jbusecke, line 876: Here it is.

xgcm/grid.py
dcherian: nested functions inherit all variables in the parent function's scope AFAIk (also you're using `da` in `_target_data_name_handling` without it being in the signature)

xgcm/grid.py
dcherian: Sorry I didn't express myself clearly enough. We should mention `broadcast` in the error message as a possible solution. 

Also broadcasting and alignment are two different things. We can check alignment with `xr.align(..., join="exact")` to avoid Ryan's problem. Automatically broadcasting should be fine (apply_ufunc should just do this IIUC) and it'll be obvious from the dimensions on the returned object.

xgcm/grid.py
jbusecke: Oh damn. Pardon my ignorance. I'll change

xgcm/grid.py
jbusecke: Ah ok that makes sense.

Assuming some hypothetical 'da' (x, y, z) and 'target_data'(xc,yc,z), strict
aligning beforehand would just give an empty array, no?

I think that neither an empty array or a resulting (x,xc,y,yc,...) is desirable. Thus I'd prefer to only allow arrays that have identical sims (except along the axis we are transforming), and give an informative warning. 

We could check only the dims which are associated with another axis as a compromise?

@rabernat do you have thoughts on this?

dcherian: Oh I see. Ya you are correct.

> Assuming some hypothetical 'da' (x, y, z) and 'target_data'(xc,yc,z), strict aligning beforehand would just give an empty array, no?

`join="exact"` will raise an error if the z coordinates are not exactly equal on both. But I see now why automatic broadcasting would not work.

My use case was `da` (z) and `target_data` (x, y, z) so I needed to broadcast to get past the check. I wonder if we could augement the if condition to also check `da.dims.issubset(target_data.dims)`, in which case broadcasting is fine.

jbusecke:
I just realized the test was ill-posed, since transform only takes the naming from `target_data` if target is not an xarray. Fixing that revealed a small bug, but now everything seems in order. 
I renamed the function to `_check_other_dims` to get rid of the misused 'alignment'. How do you feel about that @dcherian?

xgcm/grid.py
```diff
@@ -861,6 +861,31 @@ def transform(
                 "`transform` can only be used on axes that are non-periodic. Pass `periodic=False` to `xgcm.Grid`."
             )
 
+        def _target_data_name_handling(target_data):
+            """Handle target_data input without a name"""
+            if target_data.name is None:
+                warnings.warn(
+                    "The input dataarray `target_data` has no name. Will default to `_UNNAMED_`."
```
rabernat, line 868:
```suggestion
                    "Input`target_data` has no name, but we need a name for the transformed dimension. The name `TRANSFORMED_DIMENSION` will be used. To avoid this warning, call `.rename` on `target_data` before calling `transform`."
```

xgcm/grid.py
```diff
@@ -861,6 +861,31 @@ def transform(
...
+                )
+                target_data.name = "_UNNAMED_"
```
rabernat, line 870:
```suggestion
                target_data.name = "TRANSFORMED_DIMENSION"
```

xgcm/grid.py
```diff
@@ -861,6 +861,31 @@ def transform(
...
+            target_da_other_dims = set(target_da.dims) - set(self.coords.values())
+            if not target_da_other_dims.issubset(da_other_dims):
+                raise ValueError(
+                    f"Found additional dimensions [{target_da_other_dims-da_other_dims}]"
+                    "in `target_data` not found in `da`. This could mean that the target "
+                    "array is not on the same position along other axes."
+                    " If the additional dimensions are associated witha staggered axis, "
+                    "use grid.interp() to move values to other grid position. "
+                    "If additional dimensions are not related to the grid (e.g. climate "
+                    "model members or similar, use xr.broadcast() before using transform."
```
rabernat, line 886:
```suggestion
                    "model ensemble members or similar), use xr.broadcast() before using transform."
```

xgcm/test/test_transform.py
```diff
@@ -810,6 +808,36 @@ def test_grid_transform_noname(multidim_cases):
...
+    target_data.name = None
+    # the name of target_data is only used if `target` is provided as numpy array
+    target = target.data
+
+    # the high level routines should be able to deal with all cases (no error flag exception like in the mid level)
+    with pytest.warns(UserWarning):
+        transformed = grid.transform(
+            source_da, axis, target, target_data=target_data, **transform_kwargs
+        )
+    "_UNNAMED_" in transformed.dims
```
rabernat, line 838:
```suggestion
    "TRANSFORMED_DIMENSION" in transformed.dims
```

rabernat (approved):
This looks great. I approve modulo the suggested changes.

jbusecke:
Thanks for the helpful changes, @rabernat.

Issue #269, opened 2020-10-12
Link to YouTube video


rabernat:
@paigem mentioned that we should link to this xgcm video from SciPy in the docs

https://www.youtube.com/watch?v=lo6JvrJ-YA8&ab_channel=Enthought

jbusecke:
Yes, that is a great idea. Ill group that with the `quickstart guide`(#268)?

Issue #195, opened 2020-04-08
boundary="periodic"?


dcherian:
I keep finding it funny that xgcm has both `periodic` and `boundary` kwargs. Logically it seems like `periodic` should be replaced by `boundary="periodic"`?

rabernat:
This makes sense to me.

jbusecke:
That’s a great idea, that would simplify the setup.

rabernat:
PR welcome! 🙃

raehik:
Hi, I'm taking a peek at this. There are lots of tests which configure the grid with both `periodic`, and `boundary` options like `fill` and `extend`. As I understand, this change should mean you can't e.g. make an axis `periodic` and `fill` at the same time. Are there any issues expected with changing this? I ask because there I think there are lots of tests this impacts.

Actually, I can see a commented-out case which would assert `boundary=fill` not being used together with `periodic`: https://github.com/xgcm/xgcm/blob/2b28b56eebfca67dbf79530db75b48e385eb6bd3/xgcm/test/test_grid.py#L576-L580

PR #271, opened 2020-10-13
WIP: Quickstart Guide for the docs

### LINKED ISSUE #268: Quick Start Guide for docs main page[DOC]
Over the past days, I have had many great conversations with people who have been using xgcm for the first time. 

From those exchanges I think that we have an urgent need for a 'quickstart' section in our docs, which perhaps should be on the index page itself. 

I am thinking of a short 'recipe' style guide that links to other more detailed parts of the package. 

```markdown
# xgcm docs 

## What is xgcm
A very brief blurb about xgcm and the role of the `Grid` object.

## Quickstart guide

### What do you need to get going with xgcm? (Ingredients)

1. ) A dataset. Any gridded dataset will do.
2. ) Familiarize yourself with the dataset, and figure out the logical `axes`. More info [here]. 
	You don't have multiple dimensions per axis (common for observational gridded datasets)? [Here] is a guide how to reconstruct them.
3. ) Find out what 'metrics' you have available in the dataset (think: distances, areas, volumes), these are optional, but you need to define them if you want to use the more awesome features (like `.integrate`, 'average', etc). 
    - Don't have metrics? [Here] is a guide on how to infer them. 
    - You have some of the metrics but not for all possible grid positions? [Here] is a guide on how to populate the full grid.

### How to set up a grid object (Instructions)

... step by step example 
```

This would address/integrate many issues that came up in the past (e.g. #196 and others).
What do others think about this?

jbusecke:
- [ ] Closes #268 #269 
 - [ ] Add an explanation of the grid object output (See #307)
 - [ ] Explain that metrics represent e.g. the distance around a certain grid position (https://github.com/xgcm/xgcm/issues/306#issuecomment-776221011)


rcaneill:
Nice quickstart guide! Nothing much more than small details to comment on :).
I am not sure of the common practices for comments on PR, I will expect that commenting here is a good option (the other solution I thought about is changing the guide and do a PR on the PR, but it seems like an infinite loop...). Tell me if I should do differently :).

- Item 2.a, there is a typo compared to the image: you speak about SST but made a box around the SSH
- Item 3, maybe add a short sentence on the usage of the `boundary` argument, e.g.
_The boundary condition is used for operations at the boundary points, when an extra point needs to be added. Different options are possible ('fill' this extra values with a certain number, 'extend' to the nearest value or 'interpolate' linearly using the 2 nearest points). Attention, this boundary condition is used to create an extra point before running the operation, it is not used to fill the dataset afterward for points that could not be computed._
Maybe this description would suit more on another part of the doc, but this was also something I struggled with at the beginning.

Apart this quickstart guide, would it be useful if the 1st example was partially model-agnostic? Or maybe easier to do, the simple MITgcm example could describe more the grid and the scale factors / metrics so the grid creation is understandable by more people. This would be a logical continuity of the quickstart guide, where the notions are applied.

jbusecke:
Dear @rcaneill , thanks for reviewing!

To your question: It actually works really nicely (especially for text based files) to make a PR to a PR. You just have to edit the file in question on github (no cloning needed), and it will give you the option of creating a new PR. This will show up as a suggestion in the `discussion` section. If you like please try it for the typo.

I came across the issue with the boundary as well yesterday. We never really mention it. Only in a convoluted way in the setup code examples. **We need to add a more detailed section on `'boundary'` and `'fill'` in 'Simple Grids' asap!**(cc @rabernat , @dcherian ). 

I was thinking of making a completely `synthetic example at the end (e.g. make up some data, create a grid, metrics and then apply some operator). It seems that this would be in your interest too, so let me add that.

rcaneill:
> To your question: It actually works really nicely (especially for text based files) to make a PR to a PR. You just have to edit the file in question on github (no cloning needed), and it will give you the option of creating a new PR. This will show up as a suggestion in the `discussion` section. If you like please try it for the typo.

Amazing! I try it now!

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+    If you see anything else, you are in luck and xgcm detected axes from the metadata. If all your desired axes are present, you can skip the next step.
+2. Investigate the dataset, and determine which :ref:`axis position <axis_position>` is represented by the dimensions of `ds`.
+    Lets take the `CESM MOM6 Ocean Model Example Data <https://catalog.pangeo.io/browse/master/ocean/cesm_mom6_example/>`_ to go through the steps:
+    
+    .. figure:: images/quickstart_example.png
+        :scale: 75 %
+        :alt: CESM Example workflow
+
+    a. For many earth science datasets the convention is to call the position of tracers (temperature, salinity, etc.) `center`. 
+    Find a tracer data variable (we take `SST`; the sea surface temperature). Note the dimensions: `(time, yh, xh)`.
```
rcaneill, line 59:
```suggestion
    Find a tracer data variable (we take `SSH`; the sea surface height). Note the dimensions: `(time, yh, xh)`. 
```

rcaneill: Typo correction

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+    d. Repeat the steps above for each desired axis and pass a nested dictionary to the `Grid` object. E.g. ``Grid(coords={
+    'X':{'center':'xq','left':'xh'}, 'Y':{...}, 'Z':{...}})``
+
+    Find more detailed instructions :ref:`here <grids>`.
+
+    If you have only one dimension per axis, the default is to put it at the `center` location. This works but most of the functionality of xgcm does require 
+    two grid positions. You can find instructions on how to reconstruct additional dimensions `here <autogenerate_examples.ipynb>`_.
+3. Determine if your axes are periodic or define a boundary condition.
+    You can specify this either for all axes ``Grid(ds, periodic=False, boundary='fill')``
+    or per axis, by passing a list/dictionary with seperate axis names ``Grid(ds, periodic=['X'], boundary={'Y':'fill'}``.
```
rcaneill, line 79:
```suggestion
    or per axis, by passing a list/dictionary with separate axis names ``Grid(ds, periodic=['X'], boundary={'Y':'fill'}``.
    The boundary condition is used for operations at the boundary points, when an extra point needs to be added.
```

rcaneill: Typo + adding a sentence on boundary argument.

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+
+    <div style="position: relative; padding-bottom: 10.25%; height: 0; overflow: hidden; max-width: 100%; height: auto;">
+        <iframe width="560" height="315" src="https://www.youtube.com/embed/lo6JvrJ-YA8" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+    </div>
+
+What ingredients do I need to start with xgcm?
+----------------------------------------------
+1. An :ref:`installed <installation>` version of xgcm
+2. Your xarray dataset of choice `ds`.
+3. Information on how the grid of `ds` is layed out. This is often encoded in metadata or you need to infer it from the dataset (see more below).
```
rabernat, line 16:
```suggestion
3. Information on how the grid of `ds` is layed out. This is often encoded in metadata, or else you need to infer it from the dataset (see more below).
```

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+4. (Optional) All 'metrics' you have available in the dataset (think: distances, areas, volumes). You need these if you want to use the more awesome features.
```
rabernat, line 17:
```suggestion
4. (Optional) All "metrics" you have available in the dataset (think: cell distances, areas, volumes). You need these if you want to use the higher-level derivative and integral functions.
```

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+
+
+How to get started - Step by step
+---------------------------------
+The `Grid` object is essential to the way xgcm works. It stores information about the grid setup for you. 
+`Grid` knows about the logical [axes], which dimensions and metrics (represented by xarray [coordinates]) belong to each axis.
```
rabernat, line 23: Why `[coordinates]` in brackets. Is this supposed to be a link?

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+All these informations will only have to be set once and then you can focus on whatever task is at hand without needing to remember the details.
```
rabernat, line 24:
```suggestion
All this information will only have to be set once and then you can focus on whatever task is at hand without needing to remember the details.
```

doc/quickstart.rst
```diff
@@ -0,0 +1,96 @@
...
+    two grid positions. You can find instructions on how to reconstruct additional dimensions `here <autogenerate_examples.ipynb>`_.
+3. Determine if your axes are periodic or define a boundary condition.
+    You can specify this either for all axes ``Grid(ds, periodic=False, boundary='fill')``
+    or per axis, by passing a list/dictionary with seperate axis names ``Grid(ds, periodic=['X'], boundary={'Y':'fill'}``. 
+4. Parse the `metrics <grid_metrics.ipynb>`_.
+    For this you should find out all the variables in your dataset that represent a metric (distance, area, volume),
+    and sort them according to the axis or axes they represent. You can then pass a dictionary with a tuple of the representative axis/axes as key
+    and the names of the corresponding fields in `ds` as a list of strings. xgcm will automatically figure out to which grid position they belong. 
+    An example for two distances along the `X` axis and two areas in the `X`/`Y` plane: ``{('X'):['distance_a', 'distance_b'], ..., ('X', 'Y'):['area_i', 'area_j'],}``. 
+    This dictionary can contain any combination of axes as keys. The less metrics you are missing, the more accurate the results will be.
```
rabernat, line 85:
```suggestion
    This dictionary can contain any combination of axes as keys. The more metrics you have, the more accurate the results will be. 
```

rabernat (commented):
This is a great start.

doc/quickstart.rst
jbusecke: I think I wanted to link to xarray, but maybe that is a bit too much?

jbusecke: I put the link in. Cant have too many of those.

jbusecke:
Thanks a ton for the comments @rcaneill and @rabernat. Ill try to make the example at the end next week and then we can merge this.

rabernat:
Hi @jbusecke. Just pinging on this. Are you ready for another review. Would be great to get this merged.

jbusecke:
Have put aside today for the OMZ paper completely. Ill try to finish this tomorrow or Friday!

tomchor:
@jbusecke Responding here to your comment/request on #291.

The quick start guide seems like a very good start! It definitely makes understanding how `xgcm` works on a high level much much easier. I would suggest that, after those 4 steps are done and you put it all together with pseudo code, you actually do all those steps for a simple model and explain the options and choices. Some things are very straightforward to understand, but I think for others some work is needed to make it clear.

At least for me, here's how it goes:

- The grid/node nomenclature is pretty easy to understand (left, center, outer, etc.; the cartoon you guys use is really all that's needed), as is toggling the periodic topology options. So I think no need to expand too much on that.
- The boundary conditions are a bit harder to understand. I know they should roughly map to Neumann/Dirichlet boundary conditions, but it's not super clear exactly how to translate from my model's BC to what xgcm needs to (almost?) exactly reproduce it. 
   - Also, this may be a separate issue, but I find it weird that I can set a BC for a periodic axis. It makes me a little confused on how those work.
- The metrics for me are definitely the hardest to understand. (In fact I still haven't understood exactly what they are measuring.) So in this part I'm afraid some work has to be done to figure out how to design a cartoon and an example that explain exaclty what they are and how to create them for any model output. Otherwise the user might put something that works but is incorrect and will get wrong answers from then on.

Hope this helps! And keep up the good work

jbusecke:
Thank you for taking the time to go through this @tomchor. This is extremely helpful feedback. 

Some quick notes:

> The boundary conditions are a bit harder to understand.

💯 agree. In fact we barely describe the boundary conditions at all in the docs (see also #273). I am not sure how much detail should go into the quickstart guide vs a separate section. I personally think a lot of these details should go into a more detailed section.

> Also, this may be a separate issue, but I find it weird that I can set a BC for a periodic axis. It makes me a little confused on how those work.

Yeah, this is very confusing, and should definitely be remedied in the future. This will require some rather big changes and a longer deprecation cycle thought (see #195). How about for now I add a cautionary note?

> The metrics for me are definitely the hardest to understand. (In fact I still haven't understood exactly what they are measuring.) So in this part I'm afraid some work has to be done to figure out how to design a cartoon and an example that explain exaclty what they are and how to create them for any model output. Otherwise the user might put something that works but is incorrect and will get wrong answers from then on.

Ok I hear that and will try to spend some time to make this more clear. In short, metrics are quantities that relate the logical rectangular grid (or nd-array) to actual physical space, like distances between cell points, cell volume, and area, etc. How are these kinds of parameters given in the LES world? Perhaps I can draw some inspiration from there to formulate something more general. 

Again many thanks for using xgcm and providing this great feedback!

tomchor:
> Yeah, this is very confusing, and should definitely be remedied in the future. This will require some rather big changes and a longer deprecation cycle thought (see #195). How about for now I add a cautionary note?

Yeah, a cautionary note would definitely help. Also, adding a `boundary="periodic"` is also something I was gonna suggest. Seems like the easy way out.

> Ok I hear that and will try to spend some time to make this more clear. In short, metrics are quantities that relate the logical rectangular grid (or nd-array) to actual physical space, like distances between cell points, cell volume, and area, etc.

I think what metrics _represent_, is pretty well-explained. I think what is confusing is how exactly to represent that. Because there are many different sets of metrics that you can come up with that fully represent the geometry of a grid.

> How are these kinds of parameters given in the LES world?

For finite volume codes like Oceananigans the usual approach is to use a rectilinear C-grid. So it's even simpler than GCMs, that have to account for curvature, etc. Other types of code (the usual ones mix spectral in the horizontal and finite differences in the vertical) have a slightly different grid where the only variable on cell faces is the vertical velocity. Fully spectral codes (like Dedalus) use only cell centers I think. Again, LES grids should be actually easier in general.

rcaneill:
Hello @jbusecke, this is a very nice part of the doc, but that seems stopped for a while... I would be happy to help to finish it, if help is needed for anything!
If so, please let me know with a list of tasks to write :)

jbusecke:
Hey @rcaneill, thanks for the ping on this. I could indeed need some help on this. I am working with @TomNicholas on a larger internal refactor, which will fundamentally change some of the terminology regarding padding and boundary. In short we will rename the current 'boundary' to 'padding' (this is only concerned with the array boundaries which do not mean anything in terms of physical boundaries in the domain! This was always a bit confusing). Once we have merged the new internals (https://github.com/xgcm/xgcm/tree/grid_ufunc_refactor_project), we can actually deal with internal boundary conditions (#324). 

I think this distinction and explanation should be included in the quick start guide, but that might not mean that this should not get finished/merged before that. 

I think the checklist at the top is still quite up to date. The most important thing to include (as @tomchor mentioned) is how to include metrics I think. 

It would be amazing if you could take a shot at adding the missing items. Happy to review and also happy to fill in the boundary discussion (perhaps for now with a warning/note until properly implemented?).

jbusecke:
Yeah lets go via the github route. I propose:
- You fork the repo (probably already done?)

- Define my fork as remote e.g. `git remote add jbusecke ...`

- Then fetch this branch, checkout from the current state under a new name

- Push and create a PR from your changes to this branch (I suggest doing this with some minimal changes, so we can see if this works haha).

jbusecke:
Actually let me try to resolve the issue in this branch first!

jbusecke:
Ok I think that did the trick. Feel free to fetch this state. And thanks a lot for helping here @rcaneill!

rcaneill:
I opened a PR on your fork, so I guess that if you merge, it will appear here.
I can run the tests locally, and tell you whenever it is ready to merge here I guess

jbusecke:
Perfect!

rcaneill:
Hello,
I ping here in case someone has time to review my work on Julius fork:  jbusecke/xgcm#2

PR #285, opened 2020-11-30
Conda builds are currently failing because we are using a deprecated github action.

rabernat:


rabernat:
The `upstream-dev` build is failing on a pip dependency issue. @jbusecke do you have any insight on this?

jbusecke:
If this is the same problem we had before (I thought it went away in between, but might be wrong), this was a problem with the newest upstream numba and its dependencies. I was hoping that this would be solved over time. For the latest PRs I have been ignoring this failure if all the others succeed, but it might be worth digging deeper.

rabernat:
Ok I'll merge this and we will make a new issue to track it.

PR #287, opened 2020-12-01
Drop versioneer

jbusecke:
I saw [rechunker/#65](https://github.com/pangeo-data/rechunker/pull/65/files) and thought we should do the same.


jbusecke:
I pretty much just copied the steps from rechunker, does anyone know how I can test this without triggering a release?

jbusecke:
Ah i guess this is tested during the pip install in the CI, so nevermind.

jbusecke:
Why does the version in the CI jobs appear as `xgcm==0.1.dev1+gcad7f67`. Is this not correctly parsing the version from the git tag?

`Requirement already satisfied: xarray>=0.14.1 in /usr/share/miniconda/envs/test_env_xgcm/lib/python3.8/site-packages (from xgcm==0.1.dev1+gcad7f67) (0.16.1)`

rabernat:
I unfortunately know nothing about setuptools_scm. But thanks for working on this!

jbusecke:
Perhaps @ocefpaf has some input? I would definitely appreciate any pointers with this.

ocefpaf:
> Perhaps @ocefpaf has some input? I would definitely appreciate any pointers with this.

From your PR, locally, I get:

```
python setup.py --version
0.5.2.dev30+g26a310a
```

which looks correct. Let me inspect the CI...

jbusecke:
Thanks for looking into this @ocefpaf. So the convention for non released versions is to bump the bugfix increment and add that scrambled id to that? Just for my understanding.

ocefpaf:
> Thanks for looking into this @ocefpaf. So the convention for non released versions is to bump the bugfix increment and add that scrambled id to that? Just for my understanding.

Not sure. My guess is the lack os setuptools_scm in the CI. Let me try something here...

ocefpaf:
@jbusecke take a look at https://github.com/xgcm/xgcm/pull/288#issuecomment-736827260

It turns out the GHA that check the code out is probably doing a shallow clone and we cannot check the all the tags. Locally it works and it probably will work where it matters, on the publishing GHA. Let me know if you want to port #288 here or if you want to squash and merge that one instead.

jbusecke:
Thanks @ocefpaf for the explanation and working on this. I have tested it locally too and it seems to work. I'd prefer to keep this and #288 together in one PR. Do you want to push to this branch? I can also make the changes.

ocefpaf:
> Do you want to push to this branch?

I tried but I do not have the rights. It seems you did not used a fork and I don't have access to xgcm. Feel free to do whatever is easier for you. I believe I have your commits in #288, so they are already "a single" PR.

jbusecke:
Any objections to merging this @ocefpaf? If you think this is good to go ill merge.

jbusecke:
Again, many thanks for your help!

ocefpaf:
> Any objections to merging this @ocefpaf? If you think this is good to go ill merge.

LGTM!

PR #288, opened 2020-12-01
add setuptools_scm on 3.8 but not others

ocefpaf:
Just want to test if it is the lack of setuptools_scm or the pip install command with the chance of clobbering deps the culprit here.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,6 @@ jobs:
         shell: bash -l {0}
         run: |
           python -m pip install -e . --no-deps
-          python setup.py --version
           conda list
       - name: Run Tests
         shell: bash -l {0}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,8 @@ jobs:
       - name: Set up conda environment
         shell: bash -l {0}
         run: |
-          python -m pip install -e . --no-deps --force-reinstall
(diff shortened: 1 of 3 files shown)


ocefpaf:
Aha! Found the problem, it is related to the shallow clone in GHA:

```
https://github.com/xgcm/xgcm/pull/288/checks?check_run_id=1482589263#step:5:12
```

I guess it is OK as long as a real tag checkout is used to produce the tarball.

.github/workflows/ci.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Set up conda environment
         shell: bash -l {0}
         run: |
-          python -m pip install -e .
+          python -m pip install -e . --no-deps
```
jbusecke, line 41: Do you think it would be good to have a 'pip-only' ci install (without corresponding conda yaml file), to check if our dependencies are up to date?

ocefpaf: With the new pip? Yep. For sure. We don't know yet how the solver will interact with our environments and packages.

ocefpaf:
Ah. Actually, this tries to merge on your branch. xref #287

.github/workflows/ci.yaml
jbusecke: Ill set that up in a separate PR!

jbusecke:
Ah perfect!

PR #299, opened 2021-01-19
Strict conda-forge

jbusecke:
I just saw these options in the [xarray ci](https://github.com/pydata/xarray/blob/7dbbdcafed7f796ab77039ff797bcd31d9185903/.github/workflows/ci.yaml#L67-L68). I think we should also strictly build from conda-forge?

 - [ ] Closes #xxxx
 - [ ] Tests added
 - [ ] Passes `black . `


jbusecke:
I actually think it is better to define channels in the respective environment files, and we will do that once we actually need it. Closing

PR #301, opened 2021-01-19
Clean out unnecessary dependencies

jbusecke:
`pre-commit` takes care of e.g. `black` and we dont need `future` anymore I think...

### DIFF
--- a/ci/environment-core-deps.yml
+++ b/ci/environment-core-deps.yml
@@ -7,9 +7,7 @@ dependencies:
   - dask
   - numpy
   - pytest
-  - future
   - pip
-  - black
   - pip:
     - docrep<=0.2.7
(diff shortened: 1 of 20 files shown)


jbusecke:
As far as I understand `future` was just needed back in the days when we supported python v2? 
Now that we do not support that anymore, I think we can delete it and slim the dependencies and imports down a bit? 

Let me know if there is another reason to keep it that I might be missing here.

jbusecke:
Hmmm I guess this was embedded more deeply in our code than I thought. Do folks here know what is the right way to replace `iteritems()`?

jbusecke:
Ill close this, since it wasnt required by anything and turned out to be more complex thing to do, than I initially thought.

Issue #206, opened 2020-06-29
fix boundary="extrapolate"


dcherian:
There are a number of xfailed tests for when `boundary="extrapolate"`. We should fix these.

PR #325, opened 2021-05-11
adding boundary condition

### LINKED ISSUE #273: We need a section in the docs about `'boundary'` and `'fill'`[DOC]
While working on #271 I noticed that we do not have an actual section describing our boundary treatment in the docs. 
There are some mentions but the user has to read the whole page and I think more detail would still help. 

I propose we add a separate section on 'Boundary Conditions' to `grids.rst`.

rcaneill:
Aims to close #273 

I added a section about the boundary conditions. I also corrected some minor typos.


doc/grids.rst
```diff
@@ -304,6 +305,103 @@ Xgcm can also deal with complex topologies such as cubed-sphere and
...
+``extend`` to the nearest value, ``extrapolate`` linearly using the 2 nearest points,
+or periodic condition if the grid axis is periodic).
+Attention, this boundary condition is used to give the value of X, not to give the value of the 
+boundary T point after the operation.
+
+We can illustrate it by creating some data located at the U point:
+
+.. ipython:: python
+
+    g = xr.ufuncs.sqrt(ds.x_g + 0.5) + xr.ufuncs.sin((ds.x_g - 0.5) * 2 * np.pi / 8)
```
rabernat, line 343:
```suggestion
    g = np.sqrt(ds.x_g + 0.5) + np.sin((ds.x_g - 0.5) * 2 * np.pi / 8)
```

You should never have to explicitly call `xr.ufuncs`.

doc/grids.rst
```diff
@@ -304,6 +305,103 @@ Xgcm can also deal with complex topologies such as cubed-sphere and
...
+    g
+
+We show here the value of the extra added point for 5 cases (extended, extrapolated, filled with 0, filled with 5,
+and periodic). The periodic condition is not an argument of the methods, but is provided
```
rabernat, line 347: An API question: do you think we should change things so that all boundary conditions (including periodic) are handled in the same way?

rabernat (commented):
This is really fantastic @rcaneill! Thanks so much for your work.

Just one suggestion and one question. Otherwise LGTM!

doc/grids.rst
rcaneill: Oh okay I did not know that, thanks!

doc/grids.rst
rcaneill: Good question. I also asked myself about this while writing. Maybe just being able to overwrite the boundary condition at any time while calling the grid methods would be nice. So keep the periodic argument in the grid init, along with the boundary argument, but overwrite them when explicitly asked by user.

jbusecke:
This is really cool @rcaneill! Thank you very much for working on this. 
There are a few issues that might be relevant here: 
- Since we demonstrate the `extrapolate` option, we should really clean up where and when we can apply that. This has been a 'semi-hidden' feature for a long time I think (https://github.com/xgcm/xgcm/issues/206)
- https://github.com/xgcm/xgcm/issues/195 seems relevant for the discussion started by @rabernat above. My two cents: I think this makes sense without loosing generality?

I had some other small nits for the code, but overall this looks amazing!

doc/grids.rst
```diff
@@ -304,6 +305,103 @@ Xgcm can also deal with complex topologies such as cubed-sphere and
...
+as an argument of the ``xgcm.Grid``. We will thus also create 2 grids: one periodic and another one not periodic.
+
+.. ipython:: python
+
+    def plot_bc(ds):
+        plt.plot(ds.x_g, g, marker='o', color='C6', label='g')
```
jbusecke, line 353: This code block (and the following), do not show up nicely in the Readthedocs preview (https://xgcm--325.org.readthedocs.build/en/325/grids.html) because the lines are too long. 

See this example:
image

Could you shorten the lines? Ideally you would use [black](https://github.com/psf/black) to format each code cell and then paste it back? I also wonder if there is a way to do this automatically. Ill open an issue for that in the future.

rcaneill: I just tried to use black (and copy/paste), but it does not seem to solve the problem
xgcm-doc-black

rabernat:
The remaining issues with formatting are problems with our sphinx theme, not this PR.

rcaneill:
So, should I just commit the 2 lines that are changed by black and that's it?

doc/grids.rst
dcherian: Fully in favour. See #195

jbusecke:
I tried to fix the formatting here (since its good to go otherwise), but I am getting:
```
error: cannot format /Users/juliusbusecke/code/xgcm/doc/grids.rst: Cannot parse: 403:0: plt.grid(True)
```

it seems this part is upsetting blackdoc:
```
    @suppress
    plt.grid(True)
    @savefig grid_bc_extra_point.png
    plot_bc(ds)
```

@rcaneill do we necessarily need this? @andersy005 do you have an idea how to fix this?

rcaneill:
Plotting the grid is not essential, it is just that I found it more clear and easier to refer to the value of each condition (as I choose the number to be integer and be located in the grid lines)

doc/grids.rst
```diff
@@ -335,6 +336,103 @@ Xgcm can also deal with complex topologies such as cubed-sphere and
...
+
+.. ipython:: python
+
+    for (i, var) in enumerate([g_extrapolate, g_extend, g_fill_0, g_fill_2, g_perio]):
+        var.plot.line(marker='o', label=var.name)
+
+    @suppress
+    plt.grid(True)
+    @savefig grid_bc_diff.png
+    plt.legend()
```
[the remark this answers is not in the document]
rcaneill: I don't have time before September to do it, but any solution that keeps the figure untouched is fine for me, if anyone wants to commit it.
I'm glad to see motivation around this part of the doc :)

keewis:
there is also a bug in `blackdoc` which is fixed in `v0.3.4`, so you'd need to update the version in `.pre-commit-config.yaml` to get it to pass

rcaneill:
I don't know why it does not pass the code-style / pre-commit check now... The error is:

> `Traceback (most recent call last):
  File "/home/romain/.cache/pre-commit/repobuk3dc8i/py_env-python3/bin/blackdoc", line 5, in <module>
    from blackdoc.__main__ import main
  File "/home/romain/.cache/pre-commit/repobuk3dc8i/py_env-python3/lib/python3.9/site-packages/blackdoc/__main__.py", line 12, in <module>
    from .blackcompat import (
  File "/home/romain/.cache/pre-commit/repobuk3dc8i/py_env-python3/lib/python3.9/site-packages/blackdoc/blackcompat.py", line 9, in <module>
    import toml
ModuleNotFoundError: No module named 'toml'`

keewis:
yes, that's the bug I was talking about in https://github.com/xgcm/xgcm/pull/325#issuecomment-912647306. The fix is to upgrade the version of `blackdoc`.

rcaneill:
I changed the code blocks so it passes all the tests and I am happy with the way it looks. I think it's ready for merging :)

jbusecke (approved):
So sorry for the long wait again @rcaneill. Really appreciate your tenacity here. Good news is this will now be included in v0.6.0 (which we are releasing today). 

Thanks again for this important contribution

PR #329, opened 2021-05-20
Improve input checks for grid.transform

### LINKED ISSUE #328: [BUG] xgcm.transport should check that the input is datarray
Currently you can do something like: 
```python
...
grid.transform(ds, ...)
```

It then fails with some obscure dtype failure. We should have a check that the input is a dataarray. 

Just noting this here since that came up during the siparcs meeting 

cc @andersy005 @jdldeauna

jbusecke:
I tried to improve the error handling for the transform routines. The decorator `input_handling` now raises a ValueError if *any* of the data inputs is not a datarray.
The error message is however still not super useful because the variable naming differs for the grid method `grid.transform` and the lower level functions like `linear_interpolation` etc. 

I see a couple of options: 
- Catch the errors in `grid.transform` and raise new errors with proper naming (seems cumbersome).
- Implement the check at the highest level and remove it from the lower level functions (would leave those more error prone, but is anybody actually using these directly?)

cc @andersy005 

 - [x] Closes #328
 - [x] Tests added

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -874,10 +874,7 @@ class Axis:
             ("target", target, [xr.DataArray, np.ndarray]),
             ("target_data", target_data, [xr.DataArray]),
         ]:
-            if not (
-                any([isinstance(variable, ty) for ty in allowed_types])
-                or variable is None
-            ):
+            if not (isinstance(variable, tuple(allowed_types)) or variable is None):
                 raise ValueError(
                     f"`{var_name}` needs to be a {' or '.join([str(a) for a in allowed_types])}. Found {type(variable)}"
                 )
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -869,14 +869,17 @@ class Axis:
             )
 
         # complain if the target values are not provided as xr.dataarray
-        for var_name, variable in [
-            ("da", da),
-            ("target", target),
-            ("target_data", target_data),
+        for var_name, variable, allowed_types in [
+            ("da", da, [xr.DataArray]),
+            ("target", target, [xr.DataArray, np.ndarray]),
+            ("target_data", target_data, [xr.DataArray]),
         ]:
-            if not (isinstance(variable, xr.DataArray) or variable is None):
+            if not (
+                any([isinstance(variable, ty) for ty in allowed_types])
+                or variable is None
+            ):
                 raise ValueError(
-                    f"`{var_name}` needs to be a xr.DataArray. Found {type(variable)}"
+                    f"`{var_name}` needs to be a {' or '.join([str(a) for a in allowed_types])}. Found {type(variable)}"
                 )
 
         def _target_data_name_handling(target_data):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -868,6 +868,17 @@ class Axis:
                 "`transform` can only be used on axes that are non-periodic. Pass `periodic=False` to `xgcm.Grid`."
             )
 
+        # complain if the target values are not provided as xr.dataarray
+        for var_name, variable in [
+            ("da", da),
+            ("target", target),
+            ("target_data", target_data),
+        ]:
+            if not (isinstance(variable, xr.DataArray) or variable is None):
+                raise ValueError(
+                    f"`{var_name}` needs to be a xr.DataArray. Found {type(variable)}"
+                )
+
         def _target_data_name_handling(target_data):
             """Handle target_data input without a name"""
             if target_data.name is None:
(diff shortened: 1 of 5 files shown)


andersy005 (approved):
Looks good to me 👍

xgcm/test/test_transform.py
```diff
@@ -1045,3 +1045,44 @@ def test_chunking_dim_error():
...
+        )
+    assert str(e_info.value) == err_msg
+
+    transform_kwargs["target_data"] = transform_kwargs["target_data"].to_dataset(
+        name="dummy"
+    )
+
+    with pytest.raises(ValueError) as e_info:
+        transformed = grid.transform(source.data, axis, target, **transform_kwargs)
+    assert str(e_info.value) == err_msg
```
andersy005, line 1088:
```suggestion
    with pytest.raises(ValueError, match=r"should be passed as xr.DataArray"):
        transformed = grid.transform(source, axis, target, **transform_kwargs)
    with pytest.raises(ValueError, match=r"should be passed as xr.DataArray"):
        transformed = grid.transform(
            source.data, axis, target.to_dataset(name="dummy"), **transform_kwargs
        )
    transform_kwargs["target_data"] = transform_kwargs["target_data"].to_dataset(
        name="dummy"
    )
    with pytest.raises(ValueError, match=r"should be passed as xr.DataArray"):
        transformed = grid.transform(source.data, axis, target, **transform_kwargs)
    
```

andersy005 (commented):
@jbusecke, I left a few inline suggestions

jbusecke:
Thanks @andersy005. I decided to rip out the type checking and put it into the grid methods (which already does a bunch of checking). I think these error messages are sufficiently clear.

@jdldeauna Do you want to take this for a spin in your notebook? Happy to help with the steps needed for that.

jbusecke:
oops this causes some issues with target (which we allow to be a numpy array). Let me revise this.

xgcm/grid.py
```diff
@@ -868,6 +868,20 @@ def transform(
...
+            ("da", da, [xr.DataArray]),
+            ("target", target, [xr.DataArray, np.ndarray]),
+            ("target_data", target_data, [xr.DataArray]),
+        ]:
+            if not (
+                any([isinstance(variable, ty) for ty in allowed_types])
+                or variable is None
+            ):
+                raise ValueError(
+                    f"`{var_name}` needs to be a {' or '.join([str(a) for a in allowed_types])}. Found {type(variable)}"
```
jbusecke, line 882: This will result in somewhat long error messages like 
```
`da` needs to be a <class 'xarray.core.dataarray.DataArray'>. Found <class 'xarray.core.dataset.Dataset'>"
```
@andersy005 is there a nice way to shorten these strings? Might not be necessary though...

xgcm/grid.py
```diff
@@ -868,6 +868,20 @@ def transform(
...
             )
 
+        # complain if the target values are not provided as xr.dataarray
+        for var_name, variable, allowed_types in [
+            ("da", da, [xr.DataArray]),
+            ("target", target, [xr.DataArray, np.ndarray]),
+            ("target_data", target_data, [xr.DataArray]),
+        ]:
+            if not (
+                any([isinstance(variable, ty) for ty in allowed_types])
```
andersy005, line 878: `isinstance` allows checking against a tuple of types. So, the second for loop could be removed: 


```suggestion
            ("da", da, (xr.DataArray,)),
            ("target", target, (xr.DataArray, np.ndarray)),
            ("target_data", target_data, (xr.DataArray,)),
        ]:
            if not isinstance(variable, allowed_types)
```

xgcm/grid.py
andersy005: Try `variable.__class__.__name__` instead of `type(variable)`

```python
In [8]: ds.__class__
Out[8]: xarray.core.dataset.Dataset

In [9]: ds.__class__.__name__
Out[9]: 'Dataset'

In [10]: type(ds)
Out[10]: xarray.core.dataset.Dataset

```

xgcm/grid.py
jbusecke: oh sorry, didnt see that you already put in a suggestion. I did implement a similar fix below.

andersy005: No worries. The fix looks 👍🏽

andersy005:
Is this still in draft mode?

jbusecke:
Not anymore. Haha.

jbusecke:
Should we merge this? Or do you need another look? Perhaps we should also release a bugfix version so that this is updated in the pangeo cloud?

andersy005:
@jbusecke, feel free to merge this. Regarding the new release, I noticed that there hasn't been a release since 10/2020. 👍🏽 for a new release.

jbusecke:
Cool. Ill trigger one right now. Generally what do you think is 'too often' for a release cycle? Clearly we are not releasing enough!

Issue #199, opened 2020-04-16
Make _assign_metrics public


dcherian:
https://github.com/xgcm/xgcm/issues/108#issuecomment-592647947  suggests that `_assign_metrics` should be public.

That way you can use `grid.interp` and `grid.diff` to construct metrics when necessary and then just assign them to an existing object.

jbusecke:
I think this is resonable. @rabernat, do you have thoughts?

jbusecke:
I think if we do this, we could also implement a `.get_metrics_dict`. I have been messing around with subsets of data and it would be very helpful to get a complete dict like `{'X':['dx1', 'dx2', 'dx3',...], 'Y':[...]}` to pass it to a new grid object (we currently only store the actual dataarrays, not the coordinate names). This would probably only require small effort to refactor.

dcherian:
How about `.metrics` with property setters and getters?

jbusecke:
I would be in favor of that!

rabernat:
Sure, sounds good to me.

jbusecke:
Something we can integrate in this feature is a way to return the original `metrics_dict` so it could be passed to another grid object. This might be helpful for subsets of data (#193 )

jbusecke:
Just started to tinker with a PR for #103 and I think it would be helpful to address this here first. A few additional questions: 

I would create a subclass `metrics` on the grid object? which would store a dictionary like this:
```
metric_dict = {axis_combo:[(metric_name1, metric_da1), (metric_name2, metric_da2), ...]}
```
should the `set` method accept only str (names of variables in the `grid._ds` dataset? Or should we expand this functionality to accept an actual dataarray?

Should we be able to `update` a single metric once the grid object is created? Or require the user to completely 'reset' the metric_dict?

What checks should we perform in the `set` method? 

1. Make sure that each grid position is unique (e.g. if two x distances are passed and are located on the same position, we should error out)
2. ... anything else?

dcherian:
> I would create a subclass metrics on the grid object? which would store a dictionary like this:

How about something like 

``` python
class Metrics(dict):
    def __setitem__(self, key, value):
        # do validation / alignment checks here
        super().__setitem__(key, value)

    def validate(self):
        # could be useful
        pass

    def infer_remaining(self):
        # maybe?
        pass
    
class Grid:
    
    metrics: Metrics = Metrics()
    
inst = Grid()
print(inst.metrics)  # {}
inst.metrics["a"] = 10
[...]

PR #336, opened 2021-05-28
adding metrics method

jdldeauna:
- [ ] metrics added as property to grid.py
 - [ ] metrics added as setter to grid.py
 - [ ] test_grid_metrics added to test_grid.py

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -16,7 +16,7 @@ try:
 
     from .transform import conservative_interpolation, linear_interpolation
 except ImportError:
-    numba = None
+m    numba = None
 
 
 docstrings = docrep.DocstringProcessor(doc_key="My doc string")
@@ -1255,16 +1255,16 @@ class Grid:
             self.axes[axis]._facedim = facedim
             self.axes[axis]._connections = axis_links
 
-    def _assign_metrics(self, metrics):
-        """
-        metrics should look like
-           {('X', 'Y'): ['rAC']}
-        check to make sure everything is a valid dimension
-        """
+    _metrics = {}
+    
+    @property
+    def metrics(self):
+        return self._metrics
 
-        self._metrics = {}
-
-        for key, metric_vars in metrics.items():
+    @metrics.setter
+    def metrics(self, key, value):
+        
+        for key, metric_vars in key, value:
             metric_axes = frozenset(_maybe_promote_str_to_list(key))
             if not all([ma in self.axes for ma in metric_axes]):
                 raise KeyError(
@@ -1283,6 +1283,37 @@ class Grid:
                 # TODO: check for consistency of metric_var dims with axis dims
                 # check for duplicate dimensions among each axis metric
                 self._metrics[metric_axes].append(metric_var)
+        
+        self._metrics[key] = value
+    
+    def _assign_metrics(self, metrics):
+        """
+        metrics should look like
+           {('X', 'Y'): ['rAC']}
+        check to make sure everything is a valid dimension
+        """
+
+        self._metrics = {}
+
+        # for key, metric_vars in metrics.items():
+        #     metric_axes = frozenset(_maybe_promote_str_to_list(key))
+        #     if not all([ma in self.axes for ma in metric_axes]):
+        #         raise KeyError(
+        #             "Metric axes %r not compatible with grid axes %r"
+        #             % (metric_axes, tuple(self.axes))
+        #         )
+        #     # initialize empty list
+        #     self._metrics[metric_axes] = []
+        #     for metric_varname in _maybe_promote_str_to_list(metric_vars):
+        #         if metric_varname not in self._ds:
+        #             raise KeyError(
+        #                 "Metric variable %s not found in dataset." % metric_varname
+        #             )
+        #         # resetting coords avoids potential broadcasting / alignment issues
+        #         metric_var = self._ds[metric_varname].reset_coords(drop=True)
+        #         # TODO: check for consistency of metric_var dims with axis dims
+        #         # check for duplicate dimensions among each axis metric
+        #         self._metrics[metric_axes].append(metric_var)
 
     def _get_dims_from_axis(self, da, axis):
         dim = []


xgcm/grid.py
```diff
@@ -16,7 +16,7 @@
 
     from .transform import conservative_interpolation, linear_interpolation
 except ImportError:
-    numba = None
+m    numba = None
```
dcherian, line 19: tiny typo.
```suggestion
    numba = None
```

xgcm/grid.py
```diff
@@ -1255,16 +1255,16 @@ def check_neighbor(link, position):
...
 
-    def _assign_metrics(self, metrics):
-        """
-        metrics should look like
-           {('X', 'Y'): ['rAC']}
-        check to make sure everything is a valid dimension
-        """
-
-        self._metrics = {}
+    _metrics = {}
```
dcherian, line 1258: really minor comment: it's convention to move these kinds of variables up above `def __init__`

xgcm/grid.py
```diff
@@ -1255,16 +1255,16 @@ def check_neighbor(link, position):
...
+    
+    @property
+    def metrics(self):
+        return self._metrics
 
-        for key, metric_vars in metrics.items():
+    @metrics.setter
+    def metrics(self, key, value):
```
dcherian, line 1265: As you seem to have realized :) we can actually remove the `for` loop below since `key` and `value` are scalars or single values.

What we will want to do is use the `for` loop in the `__init__` method to loop over every element in the `metrics` dictionary passed to `__init__`, and assign appropriately `self.metrics[key] = value`

dcherian (commented):
Nice start @jdldeauna :partying_face:

xgcm/grid.py
jdldeauna: So I should replace `metrics=None`?

```python
class Grid:
    """
    An object with multiple :class:`xgcm.Axis` objects representing different
    independent axes.
    """

    def __init__(
        self,
        ds,
        check_dims=True,
        periodic=True,
        default_shifts={},
        face_connections=None,
        coords=None,
         _metrics = {},
        boundary=None,
        fill_value=None,
    ):
```

dcherian: sorry for the confusion, I meant
```
class Grid:
	_metrics = {}

	def __init__(
		...,
		metrics=None,
		...,
	):
		...
```

jdldeauna: Oh I see, thanks!

jdldeauna: So I should retain `metrics=None` under def __init__?

dcherian: yes that way users can opt-in to providing metrics...

andersy005: @dcherian, we might want to be careful here so as to avoid race conditions and other weird outcomes. Using `_metrics` as a class attribute instead of an instance attribute results into unintended outcomes:


```python
n [1]: class Grid:
   ...:     _metrics = {}
   ...: 

In [2]: a = Grid()

In [3]: a._metrics['key'] = 'value'

In [5]: a._metrics
Out[5]: {'key': 'value'}

In [6]: b = Grid()

In [7]: b._metrics
Out[7]: {'key': 'value'}
```

Should we make `_metrics` an instance attribute instead?

dcherian: Oh oops! thanks Anderson

``` python
class Grid:

	def __init__(
		...,
		metrics=None,
		...,
	):
		self._metrics = {}

		# parse metrics dict here
		...
```

andersy005: I just saw that `_metrics` is an instance attribute in the existing version. Are there any benefits to making this a class attribute?

dcherian: No, I was typing without thinking.

xgcm/grid.py
jdldeauna: Do you mean changing:
```python
class Grid
...
if metrics is not None:
    self._assign_metrics(metrics)
```

into:

```python
    self._metrics(metrics)
```

dcherian: Something like

``` python
def __init__(..., metrics=None):
	self._metrics = {}
	if metrics is not None:
		# loop over metrics.items() and assign to self._metrics
```

jdldeauna: I'm trying to understand whether the `for loop` previously under _assign_metrics should be under `__init__` or the `@metrics.setter`. If the `__init__` for loop already assigns `self._metrics[key] = value`, what would be the purpose of the `@metrics.setter` method?

dcherian: It should be under `__init__`.

> If the __init__ for loop already assigns self._metrics[key] = value

It should assign to `self.metrics`, *not* `self._metrics` (so it will use the `setter` method, and make appropriate checks)

xgcm/grid.py
```diff
@@ -1255,45 +1258,42 @@ def check_neighbor(link, position):
...
-        check to make sure everything is a valid dimension
-        """
 
-        self._metrics = {}
+        metric_axes = frozenset(_maybe_promote_str_to_list(key))
+        if not all([ma in self.axes for ma in metric_axes]):
+            raise KeyError(
+                "Metric axes %r not compatible with grid axes %r"
+                % (metric_axes, tuple(self.axes))
+            )
```
andersy005, line 1273: You could us an f-string here:

```suggestion
            raise KeyError(
                f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
            )
```

xgcm/grid.py
```diff
@@ -1255,45 +1258,42 @@ def check_neighbor(link, position):
...
+
+        # initialize empty list
+        self._metrics[metric_axes] = []
+
+        metric_varname = frozenset(_maybe_promote_str_to_list(value))
+        if metric_varname not in self._ds:
+            raise KeyError("Metric variable %s not found in dataset." % metric_varname)
```
andersy005, line 1280:
```suggestion
            raise KeyError(f"Metric variable {metric_varname} not found in dataset.")
```

jdldeauna:
Using this:

```python
def __init__ ()
…
if metrics is not None:
    for key,value in metrics.items()
        self.metrics[key] = value
```
actually bypasses the metrics.setter method, and `value` is just determined by 
```python
@property
def metrics(self):
    return self._metrics
```

So for now, it seems using a `set_metrics(key, value)` method allows for the metric axes to be assigned as variables from the dataset. I’m not entirely sure how to move forward with the setter/getter methods, but using this configuration for now allows for exploring other functionality to `set_metrics` (e.g., how to handle updating previously assigned metrics with new values).

xgcm/grid.py
```diff
@@ -1255,34 +1258,62 @@ def check_neighbor(link, position):
...
+    def set_metrics(self, key, value):
 
-        for key, metric_vars in metrics.items():
-            metric_axes = frozenset(_maybe_promote_str_to_list(key))
-            if not all([ma in self.axes for ma in metric_axes]):
+        metric_axes = frozenset(_maybe_promote_str_to_list(key))
+        if not all([ma in self.axes for ma in metric_axes]):
+            raise KeyError(
+                f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
+            )
```
andersy005, line 1272: Nitpick: We could invert any/all to simplify comparisons 

```suggestion
        if any(ma not in self.axes for ma in metric_axes):
            raise KeyError(
                f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
            )
```

xgcm/grid.py
```diff
@@ -1255,34 +1258,62 @@ def check_neighbor(link, position):
...
+        #     for metric_varname in _maybe_promote_str_to_list(metric_vars):
+        #         if metric_varname not in self._ds:
+        #             raise KeyError(
+        #                 "Metric variable %s not found in dataset." % metric_varname
+        #             )
+        #         # resetting coords avoids potential broadcasting / alignment issues
+        #         metric_var = self._ds[metric_varname].reset_coords(drop=True)
+        #         # TODO: check for consistency of metric_var dims with axis dims
+        #         # check for duplicate dimensions among each axis metric
+        #         self._metrics[metric_axes].append(metric_var)
```
andersy005, line 1316: Is this still needed?

xgcm/grid.py
jbusecke: I like it, but maybe we want to make it even more explicit and loop over the axes? Then we can show which axes are not compatible?
Something like:
```
axes_not_found = []
for ma in metric_axes:
    if ma not in self.axes:
        axes_not_found.append(ma)

if len(axes_not_found) > 0:
    f"Metric axes {axes_not_found!r} not compatible with grid axes {tuple(self.axes)!r}"
```

Just an idea.

andersy005: I'm in favor of the explicit `for loop`. It's easy to understand (for me at least :). We could use a list comprehension if we want to:

```python
axes_not_found = [ma for ma in metric_axes if ma not in self.axes]

if len(axes_not_found) > 0:
    f"Metric axes {axes_not_found!r} not compatible with grid axes {tuple(self.axes)!r}"
```

jbusecke: I would leave this up to @jdldeauna. Equally fine with both.

dcherian:
>  I’m not entirely sure how to move forward with the setter/getter methods, but using this configuration for now allows for exploring other functionality to set_metrics

Sounds good. We can get this in and then explore the property approach later.

.gitignore
```diff
@@ -72,3 +72,7 @@ dask-worker-space/dask-worker-space/purge.lock
 
 # CI/Linting
 .mypy_cache
+xgcm/.vscode/settings.json
+02_Testing xgcm metrics.ipynb
+.ipynb_checkpoints/02_Testing xgcm metrics-checkpoint.ipynb
+xgcm/test/sample.py
```
dcherian, line 78:
```suggestion
xgcm/.vscode
*.ipynb
.ipynb_checkpoints
```

delete `sample.py`?

xgcm/grid.py
jdldeauna: I find this for loop easier to understand, too:
```python
axes_not_found = [ma for ma in metric_axes if ma not in self.axes]
```
Thanks!

xgcm/grid.py
jdldeauna: You're right, I think set_metrics already works, so I have removed this section.

dcherian:
> We can get this in and then explore the property approach later.


After some experimentation, you cannot set a dictionary as a property. That's probably why I typed out the more complicated subclassing approach here: https://github.com/xgcm/xgcm/issues/199#issuecomment-708644470

@jdldeauna I apologize for not having experimented with this more thoroughly earlier.

I can open a PR with the dict-subclass approach once you're done with implementing all this functionality.

PR #338, opened 2021-06-03
Implementing axis directionality

### LINKED ISSUE #337: [FEATURE] Specify directionality of axes
I believe it is currently the case that `xgcm` does not intrinsically know the _direction_ in which an axis is defined and thus the direction in which to take differences and/or derivatives.

This can be problematic to the naive user (e.g. me), with the most common example would be a model whose index in the vertical dimension ascends with depth, i.e. the index counts from 1 to nz, whereas the depth descends from, say, -5m to -5000m.

I encountered this when naively performing a vertical derivative with ECCO output, and recognizing that it gave an answer that was of the wrong sign. The manual hack is to multiply the output of `.diff` by -1, or to multiply the appropriate metric by -1 before using `.derivative`.

Would it be straightforward (and desired) to allow for the direction of axes to be specified at the point of defining the Grid object, e.g.
```
grid = Grid(ds, axis_direction={'X':'increasing', 'Y':'increasing', 'Z':'decreasing'})
```
and use this information in `.diff`?

Happy to work on a PR if it is thought to be a useful addition.

gmacgilchrist:
This is a first pass at implementing direction-aware axes (#337), including a preliminary test, with thanks to @jbusecke. The main behavior is to define a sign for the `Grid.diff` function.

There was some discussion in #337 about naming conventions. I have implemented just a simple version here wherein a direction (increasing or decreasing) is specified at the point of defining the grid. Happy to iterate on this logic.  

I was not sure what would be the correct way to define the default behavior. Here I have specified that `None` defaults to assuming that all axes are `increasing`. This leads to some redundancy, but keeps the default conventions consistent with the API for other variables. 

 - [x] Closes #337 
 - [x] Tests added

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -118,8 +118,6 @@ class Axis:
               latitude increases with increasing y-index.
             * 'decreasing': coordinate decreases with increasing index, e.g. 
               depth decreases with increasing z-index.
-            Optionally a dict mapping axis name to separate directions for each
-            axis can be passed.
             
 
         REFERENCES
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -74,6 +74,7 @@ class Axis:
         coords=None,
         boundary=None,
         fill_value=None,
+        axis_direction="increasing",
     ):
         """
         Create a new Axis object from an input dataset.
@@ -109,6 +110,8 @@ class Axis:
             boundary kwarg when calling specific methods.
         fill_value : {float}, optional
             The value to use in the boundary condition when `boundary='fill'`.
+        axis_direction : str, optional,
+            Directionality of index for axis.
 
         REFERENCES
         ----------
@@ -133,6 +136,13 @@ class Axis:
         else:
             # fall back on comodo conventions
             self.coords = comodo.get_axis_positions_and_coords(ds, axis_name)
+
+        if axis_direction=="increasing":
+            self.direction_sign=1
+        elif axis_direction=="decreasing":
+            self.direction_sign=-1
+        else:
+            raise ValueError(f"Axis direction not recognized. Expecting `increasing` or `decreasing`, got `{axis_direction}`.")
 
         # self.coords is a dictionary with the following structure
         #   key: position_name {'center' ,'left' ,'right', 'outer', 'inner'}
@@ -645,7 +655,7 @@ class Axis:
             boundary_discontinuity,
             vector_partner,
             keep_coords=keep_coords,
-        )
+        )*self.direction_sign
 
     @docstrings.dedent
     def cumsum(self, da, to=None, boundary=None, fill_value=0.0, keep_coords=False):
@@ -1057,6 +1067,7 @@ class Grid:
         metrics=None,
         boundary=None,
         fill_value=None,
+        axis_direction=None,
     ):
         """
         Create a new Grid object from an input dataset.
@@ -1105,6 +1116,8 @@ class Grid:
             can be passed.
         keep_coords : boolean, optional
             Preserves compatible coordinates. False by default.
+        axis_direction : {str, dict} optional
+
 
         REFERENCES
         ----------
@@ -1150,6 +1163,16 @@ class Grid:
                     "mapping axis name to a boundary option; a number or None."
                 )
 
+            if isinstance(axis_direction, dict):
+                axis_axis_direction = axis_direction.get(axis_name, None)
+            elif isinstance(axis_direction, str) or axis_direction is None:
+                axis_axis_direction = axis_direction
+            else:
+                raise ValueError(
+                    f"axis_direction={axis_direction} is invalid. Please specify a dictionary "
+                    "mapping axis name to a direction option; a string or None."
+                )
+
             self.axes[axis_name] = Axis(
                 ds,
                 axis_name,
@@ -1158,6 +1181,7 @@ class Grid:
                 coords=coords.get(axis_name),
                 boundary=axis_boundary,
                 fill_value=axis_fillvalue,
+                axis_direction=axis_direction,
             )
 
         if face_connections is not None:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -74,7 +74,7 @@ class Axis:
         coords=None,
         boundary=None,
         fill_value=None,
-        axis_direction="increasing",
+        direction=None,
     ):
         """
         Create a new Axis object from an input dataset.
@@ -110,8 +110,17 @@ class Axis:
             boundary kwarg when calling specific methods.
         fill_value : {float}, optional
             The value to use in the boundary condition when `boundary='fill'`.
-        axis_direction : str, optional,
-            Directionality of index for axis.
+        direction : str, optional
+            The direction in which axes are defined:
+            * None: Do not define a specific direction; defaults to same
+              behaviour as 'increasing'
+            * 'increasing': coordinate increases with increasing index, e.g. 
+              latitude increases with increasing y-index.
+            * 'decreasing': coordinate decreases with increasing index, e.g. 
+              depth decreases with increasing z-index.
+            Optionally a dict mapping axis name to separate directions for each
+            axis can be passed.
+            
 
         REFERENCES
         ----------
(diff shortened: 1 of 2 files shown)


xgcm/grid.py
```diff
@@ -1150,6 +1179,16 @@ def __init__(
                     "mapping axis name to a boundary option; a number or None."
                 )
 
+            if isinstance(direction, dict):
```
jbusecke, line 1182: We could add some logic to infer the values of direction using cf-xarray, and then overwrite values from there with the ones manually provided.

jbusecke (changes requested):
Super nice start @gmacgilchrist! I would love to hear what the other maintainers have to say. 

A few more suggestions to add after this: 
1) I think it would be super nice to have the directionality indicated in the print representation (Modifying the text in [here](https://github.com/xgcm/xgcm/blob/3b809933ce7758aa9357a1d13cdb910147d875c6/xgcm/grid.py#L211-L219) should do the trick.
2) We definitely need some short example in the docs, showing this off!
3) I think we could implement the CF-based autodetection in a follow up PR, but if you are keen, it could also be included here!

jbusecke:
You should be able to get the `pre-commit` CI to pass by following [these steps](https://xgcm.readthedocs.io/en/latest/contributor_guide.html#code-formatting). You can also install the black formatter for vscode 😜

xgcm/test/test_grid.py
```diff
@@ -782,3 +782,14 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
     actual2 = grid2.interp(ds.data_g, ("X", "Y"))
 
     xr.testing.assert_identical(actual1, actual2)
+
+def test_axisdirectionality():
```
jbusecke, line 786: I think you can probably test this over more of the examples by using:
```python
def test_axisdirectionality(all_datasets):    
    ds, _,_ = all_datasets
    ...
```

This might expose some more edge cases.

gmacgilchrist:
Thanks for these thoughts and suggestions @jbusecke, I will get to them next week.

jbusecke:
Awesome. I am very excited for this addition.

xgcm/grid.py
dcherian: We only need to look for the `positive` attribute.
``` python
if ....attrs["positive"] == "up":
	axis_direction = "increasing"
elif ....attrs["positive"] == "down":
    axis_direction = "decreasing"
else:
	raise ValueError
```

xgcm/grid.py
```diff
@@ -109,6 +110,15 @@ def __init__(
...
         fill_value : {float}, optional
             The value to use in the boundary condition when `boundary='fill'`.
+        direction : str, optional
+            The direction in which axes are defined:
+            * None: Do not define a specific direction; defaults to same
+              behaviour as 'increasing'
+            * 'increasing': coordinate increases with increasing index, e.g. 
+              latitude increases with increasing y-index.
+            * 'decreasing': coordinate decreases with increasing index, e.g. 
+              depth decreases with increasing z-index.
```
rabernat, line 120: I know the naming has already been discussed a bit. I'll give my opinion.

This seems a little complicated and verbose for something that is essentially a binary value. Either the axis is increasing upward or not. So I would weakly favor the `positive_upward=True` convention.

xgcm/grid.py
```diff
@@ -134,6 +144,13 @@ def __init__(
             # fall back on comodo conventions
             self.coords = comodo.get_axis_positions_and_coords(ds, axis_name)
 
+        if direction=="increasing" or direction is None:
+            self.direction_sign=1
+        elif direction=="decreasing":
+            self.direction_sign=-1
+        else:
+            raise ValueError(f"Axis direction not recognized. Expecting `increasing` or `decreasing` or None, got `{direction}`.")
```
rabernat, line 152: Similar to my above comment, using a boolean value here would obviate a lot of this logic.

xgcm/grid.py
```diff
@@ -645,7 +662,7 @@ def diff(
             boundary_discontinuity,
             vector_partner,
             keep_coords=keep_coords,
-        )
+        )*self.direction_sign
```
rabernat, line 665: IMO, I think we do want it in `diff`. Otherwise, what's the point of having the parameter at all? If you don't want to use it, just set `positive_upward=True` (the default).

I would also prefer to implement this in the following way.
```python
result = self._neighbor_binary_func(...)
if not self.positive_upward:
    result *= -1
return result
```

Although it doesn nothing, needlessly multiplying by 1 increases the computational cost of the dask graph.

rabernat (commented):
Thanks so much @gmacgilchrist for working on this important feature! Really appreciate your contribution. 🙌 

A few comments below.

xgcm/grid.py
dcherian: >  If you don't want to use it, just set positive_upward=True (the default).

I favor opt-in. But that's the opposite interpretation of `positive_upward`, that means positive values in the result indicate field values increasing upward in physical space. Maybe we need something else then.

rabernat: Agree `positive_upward` may not be the right term. For the sake of discussion, let's just call it, super verbosely,`logical_ordering_is_the_same_as_physical_ordering=True`. The value of `True` represents the current default, where we make no distinction between the logical and physical ordering.

I'm unclear on what you mean by opt-out.  With `logical_ordering_is_the_same_as_physical_ordering=True`, the default, nothing would change from the current behavior. With `numerical_ordering_is_the_same_as_physical_ordering=False`, the `diff` and `derivative` calls would get an extra factor of -1. By setting this parameter to a non-default value, isn't the user already "opting in"? Or do you imagine they would specify numerical_ordering_is_the_same_as_physical_ordering=False` when creating a `Grid` and then again when calling `diff`? That feels a bit redundant.

dcherian: >  Or do you imagine they would specify numerical_ordering_is_the_same_as_physical_ordering=Falsewhen creating aGridand then again when callingdiff`? That feels a bit redundant.

No. I am saying the same thing as you but with inverted perspective =) . `use_physical_ordering` seems like a good name. I'll switch the cf_xarray flag to match whatever we decide here.

cc @malmans2 who may have ideas.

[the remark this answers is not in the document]
jbusecke: I disagree with this, TBH. If not set explicitly in the call to `Grid(...)` nothing changes here. This would only be a slight 'black-box' situation if we detect directionality from CF-attributes (which I think is not *yet* done here). In that case we should just raise a proper warning IMO.

Implementing this on the lowest level makes sure it 'propagates' up to whatever more complex functions we come up with. This might be personal preference, but I would find it way more confusing (and it would require tons of doc changes) if `grid.diff(da, axis) / da_metric != grid.derivative(da,axis)`

jbusecke: I think @rabernat s implementation is good. We don't want to grow any dask graphs hahaha.

I am not super convinced of using `physical_ordering` as a kwarg. I think it very nicely describes what is happening to us devs, but might be very confusing to potential users. I did not dislike `direction` to much TBH, but maybe the arguments could be `forward`/`backward` (`forward = logical ordering represents increasing coordinates`). Meh, not in love with that. Lets discuss this further...

gmacgilchrist: Thank you for all of these helpful comments. I likewise agree with @rabernat 's implementation, and will make those changes.

In terms of naming, it seems tricky to find something that is both logically correct and intuitive to the user. I like the use of "ordering". I was thinking about `ordered=True` or `consistent_ordering=True`? Not entirely logically correct perhaps, but close enough. Together with a careful description in the docstring this might be enough?

rabernat: Regarding naming, MITgcm uses [rkSign](https://github.com/MITgcm/MITgcm/blob/b7589390f219db12d2062055ad7f822bd7b80a0e/model/inc/GRID.h#L308-L309). Maybe we could use `axis_sign`?

jbusecke: I like that!

gmacgilchrist: With the options then being `positive` for the case that the logical ordering matches the physical ordering, and `negative` for the opposite? So in the common case for the vertical axis the sign would be negative? I can see a user getting confused by that implementation, just in the same sense that `positive_upward` is not completely foolproof.  That also gets away from the boolean implementation that @dcherian favored.

dcherian: Isn't `axis_sign` (or any other kwarg) useless if it isn't boolean? If you're going to specify `axis_sign=-1` you could just do `-1 * grid.diff(...)`.

jbusecke: I wanted to revive this thread here, because I think this is some nice functionality that can be implemented relatively easily. 

After rereading this I think I am in favor of 'axis_sign` after all, mostly because it is succinct. I think the most important thing to do is adding good docstrings and examples to make it really clear what is happening.

jbusecke:
I would like to pick up this discussion here, now that #344 is implemented. I believe we still need to agree on the naming, but I was also curious if there is a way to implement the sign swapping inside the grid_ufuncs (avoiding additional dask tasks)? 

cc @TomNicholas

dcherian:
> if there is a way to implement the sign swapping inside the grid_ufuncs (avoiding additional dask tasks)

multiplying works blockwise, so it'll just get fused no? 

image

jbusecke:
> multiplying works blockwise, so it'll just get fused no?

Ah yes good point!

dhruvbalwada:
Wondering what the status of this PR is?

jbusecke:
To avoid further confusion here I will close this PR for now.

PR #343, opened 2021-06-09
add interp metric method

### LINKED ISSUE #234: grid.interp_like()
I propose a small addition to xgcm Axis/Grid. 

`grid.interp_like(da, da_target)` (adopting naming from xarray) would determine which axes are found in both `da` and `da_target` and if da is not on the same axis position as `da_target` use interpolation along the axis to 'move' `da` to the adjacent coordinate.

I found this would be very helpful in many contexts that I have encountered in the past. 

Happy to put in a PR soon, but wanted to discuss  first.

jdldeauna:
- [x] Closes #234 
 - [x] Tests added (test_metrics.py/test_interp_metrics)
 - [x] Tests added (test_grid.py/test_interp_like)

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -11,7 +11,7 @@ v0.6.0 (unreleased)
 New Features
 ~~~~~~~~~~~~
 - :py:meth:`~xgcm.grid.Grid.set_metrics` now enables adding metrics to a grid object (:pull:`336`, :issue:`199`).
-  By `Dianne Deauna <https://github.com/jdldeauna>`_.
+  By `Dianne Deauna <https://github.com/jdldeauna>`_ under the `SIParCS internship <https://www2.cisl.ucar.edu/siparcs-2021-projects#8>`_.
 
 
 v0.5.2 (2021/5/27)
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1354,49 +1354,48 @@ class Grid:
                 pass
         if metric_vars is None:
             raise KeyError(
-                "Unable to find any combinations of metrics for "
-                "array dims %r and axes %r" % (array_dims, axes)
+                f"Unable to find any combinations of metrics for array dims {array_dims!r} and axes {axes!r}"
             )
 
         # return the product of the metrics
         return functools.reduce(operator.mul, metric_vars, 1)
 
     @docstrings.dedent
-    def interp_like(self, da_target, da_source):
-        """Compares positions between two data arrays and interpolates missing values
+    def interp_like(self, array, like):
+        """Compares positions between two data arrays and interpolates da_source to the position of da_target if necessary
 
         Parameters
         ----------
-        da_target : DatArray
-            Specifies which positions and coordinates should be present in the source array
-        da_source : DataArray
-            Array with initial position and coordinate values for comparison with target array
+        array : DataArray
+            DataArray to interpolate to the position of like
+        like : DataArray
+            DataArray with desired grid positions for source array
 
         Returns
         -------
-        da_source : DataArray
+        array : DataArray
             Source data array with updated positions along axes matching with target array
         """
 
         for axname, axis in self.axes.items():
-            # This will raise a KeyError since this for-loop goes through all axes contained in self,
-            # but the method is applied for only 1 axis at a time
             try:
-                position_da_target, _ = axis._get_axis_coord(da_target)
-                position_da_source, _ = axis._get_axis_coord(da_source)
+                position_array, _ = axis._get_axis_coord(array)
+                position_like, _ = axis._get_axis_coord(like)
+            # This will raise a KeyError if you have multiple axes contained in self,
+            # since the for-loop will go through all axes, but the method is applied for only 1 axis at a time
             except KeyError:
                 continue
-            if position_da_target != position_da_source:
-                da_source = self.interp(da_source, axname)
+            if position_like != position_array:
+                array = self.interp(array, axname)
 
-        return da_source
+        return array
 
     def _interp_metric(self, da, axes):
         metric_available = self._metrics.get(frozenset(axes), None)
         if metric_available is not None:
             # this function works with only one metric at a time
-            m = metric_available[0]
-            metric_interp = self.interp_like(da, m)
+            metric = metric_available[0]
+            metric_interp = self.interp_like(metric, da)
         return metric_interp
 
     def __repr__(self):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1362,7 +1362,7 @@ class Grid:
 
     @docstrings.dedent
     def interp_like(self, array, like):
-        """Compares positions between two data arrays and interpolates da_source to the position of da_target if necessary
+        """Compares positions between two data arrays and interpolates array to the position of like if necessary
 
         Parameters
         ----------
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -37,7 +37,7 @@ class Axis:
     """
     An object that represents a group of coordinates that all lie along the same
     physical dimension but at different positions with respect to a grid cell.
-    There are four possible positions::
+    There are four possible positions:
 
          Center
          |------o-------|------o-------|------o-------|------o-------|
@@ -1261,7 +1261,9 @@ class Grid:
     def set_metrics(self, key, value):
 
         metric_axes = frozenset(_maybe_promote_str_to_list(key))
-        if not all([ma in self.axes for ma in metric_axes]):
+        axes_not_found = [ma for ma in metric_axes if ma not in self.axes]
+
+        if len(axes_not_found) > 0:
             raise KeyError(
                 f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
             )
@@ -1358,6 +1360,26 @@ class Grid:
 
         # return the product of the metrics
         return functools.reduce(operator.mul, metric_vars, 1)
+
+    def interp_like(self, da_source, metric):
+        for axname, axis in self.axes.items():
+            try:
+                position_da, _ = axis._get_axis_coord(da_source)
+                position_metric, _ = axis._get_axis_coord(metric)
+            except KeyError:
+                continue
+            if not position_da == position_metric:
+                metric = self.interp(metric, axname)
+
+        return metric
+
+    def interp_metric(self, da, axes):
+        metric_available = self._metrics.get(frozenset(axes), None)
+        if metric_available is not None:
+            m = metric_available[0]
+            metric_interp = self.interp_like(da, m)
+
+        return metric_interp
 
     def __repr__(self):
         summary = ["<xgcm.Grid>"]
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1368,19 +1368,19 @@ class Grid:
         Parameters
         ----------
         da_target : DatArray
-            Original array for comparison with target array
+            Specifies which positions and coordinates should be present in the source array
         da_source : DataArray
-            Specifies which positions and coordinates should be present in the target array
+            Array with initial position and coordinate values for comparison with target array
 
         Returns
         -------
         da_source : DataArray
-            Source data array with updated positions along axes compared with the target array
+            Source data array with updated positions along axes matching with target array
         """
 
         for axname, axis in self.axes.items():
             # This will raise a KeyError since this for-loop goes through all axes contained in self,
-            # but it is possible to apply the method for only 1 axis at a time
+            # but the method is applied for only 1 axis at a time
             try:
                 position_da_target, _ = axis._get_axis_coord(da_target)
                 position_da_source, _ = axis._get_axis_coord(da_source)
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1265,7 +1265,7 @@ class Grid:
 
         if len(axes_not_found) > 0:
             raise KeyError(
-                f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
+                f"Metric axes {axes_not_found!r} not compatible with grid axes {tuple(self.axes)!r}"
             )
 
         # initialize empty list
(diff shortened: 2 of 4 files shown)


xgcm/grid.py
```diff
@@ -1261,7 +1261,9 @@ def check_neighbor(link, position):
     def set_metrics(self, key, value):
 
         metric_axes = frozenset(_maybe_promote_str_to_list(key))
-        if not all([ma in self.axes for ma in metric_axes]):
+        axes_not_found = [ma for ma in metric_axes if ma not in self.axes]
```
jdldeauna, line 1264: I forgot to change this from PR #336

xgcm/grid.py
```diff
@@ -1261,7 +1261,9 @@ def check_neighbor(link, position):
...
+
+        if len(axes_not_found) > 0:
             raise KeyError(
                 f"Metric axes {metric_axes!r} not compatible with grid axes {tuple(self.axes)!r}"
```
dcherian, line 1268: Should we also say `f"the following axes {axes_not_found} were not found"`.

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
         # return the product of the metrics
         return functools.reduce(operator.mul, metric_vars, 1)
 
+    def interp_like(self, da_source, metric):
+        for axname, axis in self.axes.items():
+            try:
+                position_da, _ = axis._get_axis_coord(da_source)
```
dcherian, line 1367: minor comment: Maybe next we can rename `_get_axis_coord` to `_get_position_name`. Thats literally what the docstring says
> """Return the position and name of the axis coordiante in a DataArray."""

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
...
+                position_metric, _ = axis._get_axis_coord(metric)
+            except KeyError:
+                continue
```
dcherian, line 1370: Could be nice to add a comment explaining when a KeyError might be raised

xgcm/grid.py
jbusecke, line 1268: I think `metric_axes` should be replaced by `axes not found`?

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
...
+            if not position_da == position_metric:
```
dcherian, line 1371:
```suggestion
            if position_da != position_metric:
```

minor comment: slightly more conventional syntax

xgcm/test/test_metrics.py
```diff
@@ -163,3 +163,16 @@ def test_set_metric():
...
+
+    ds, coords, _ = datasets_grid_metric("C")
+
+    grid = Grid(ds, coords=coords)
+    grid.set_metrics("X", "dx_n")
+    interp_metric = grid.interp_metric(ds.u, "X")
+
+    test_metric = grid.interp(ds.dx_t, "X")
+
+    assert interp_metric.dims == test_metric.dims
```
dcherian, line 178: we could go further and use `xr.testing.assert_equal` on `interp_metric` and `test_metric`

dcherian (commented):
nice stuff diane

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
         # return the product of the metrics
         return functools.reduce(operator.mul, metric_vars, 1)
 
+    def interp_like(self, da_source, metric):
```
jbusecke, line 1364: I find the naming a bit confusing. `da_source` to me implies 'the dataarray that will be interpolated' onto a 'target'.
I would replace in the following way:
`da_source --> da_target`
`metric --> da_source`

The reason why I would not use anything 'metric' here is because this function has much broader use cases which do not necessarily involve metrics.

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
...
+        for axname, axis in self.axes.items():
```
jbusecke, line 1365: This function is part of the public API and should get a [docstring](https://www.python.org/dev/peps/pep-0257/)

xgcm/grid.py
```diff
@@ -1359,6 +1361,26 @@ def iterate_axis_combinations(items):
...
+                position_metric, _ = axis._get_axis_coord(metric)
+            except KeyError:
+                continue
+            if not position_da == position_metric:
+                metric = self.interp(metric, axname)
+
+        return metric
+
+    def interp_metric(self, da, axes):
+        metric_available = self._metrics.get(frozenset(axes), None)
```
jbusecke, line 1377: This one should probably get one too?

xgcm/test/test_metrics.py
```diff
@@ -163,3 +163,16 @@ def test_set_metric():
 
         for metric_expected, metric in zip(v, grid._metrics[k]):
             xr.testing.assert_equal(metric_expected.reset_coords(drop=True), metric)
+
+
```
jbusecke, line 167: We could parametrize this test, so that we can check this for more than one predefined metric. Happy to help with that.

xgcm/test/test_metrics.py
jbusecke, line 178: You can also check that the values are the same (not just the dimensions) with `xr.testing.assert_allclose(test_metric, interp_metric)`

xgcm/test/test_metrics.py
```diff
@@ -163,3 +163,16 @@ def test_set_metric():
 
         for metric_expected, metric in zip(v, grid._metrics[k]):
             xr.testing.assert_equal(metric_expected.reset_coords(drop=True), metric)
+
```
jbusecke, line 166: I think we should also have a test for `interp_like`, but I think that should live in `test_grid.py`.

jbusecke (changes requested):
Awesome progress @jdldeauna. I think the core logic is good to go. 

I had a couple of comments regarding the naming, docstrings etc. 

I also think you should probably add a small example to the documentation? We could do that in a separate PR too though.

jbusecke:
LOL @dcherian that was a dupe review on my end!

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
         # return the product of the metrics
         return functools.reduce(operator.mul, metric_vars, 1)
 
-    def interp_like(self, da_source, metric):
+    @docstrings.dedent
+    def interp_like(self, da_target, da_source):
+        """Compares positions between two data arrays and interpolates missing values
```
jbusecke, line 1366: I find this description a bit misleading. To me it sounds like this interpolates gaps in the the data. How about: "Compares positions between two dataarrays and interpolates `da_source` to the position of `da_target` if necessary."

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
+
+        Parameters
+        ----------
+        da_target : DatArray
+            Original array for comparison with target array
```
jbusecke: How about: 'dataarray with desired grid positions`

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
+        da_source : DataArray
+            Specifies which positions and coordinates should be present in the target array
```
jbusecke: And here: 'dataarray to interpolate to the position of `da_target`'

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
         return functools.reduce(operator.mul, metric_vars, 1)
 
-    def interp_like(self, da_source, metric):
+    @docstrings.dedent
+    def interp_like(self, da_target, da_source):
+        """Compares positions between two data arrays and interpolates missing values
+
+        Parameters
+        ----------
+        da_target : DatArray
```
jbusecke, line 1370: Typo

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
+
         for axname, axis in self.axes.items():
+            # This will raise a KeyError since this for-loop goes through all axes contained in self,
+            # but it is possible to apply the method for only 1 axis at a time
             try:
-                position_da, _ = axis._get_axis_coord(da_source)
-                position_metric, _ = axis._get_axis_coord(metric)
+                position_da_target, _ = axis._get_axis_coord(da_target)
+                position_da_source, _ = axis._get_axis_coord(da_source)
             except KeyError:
```
jbusecke, line 1387: Can you add a little comment for other contributors, why you expect a key error?

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
-        return metric
+        return da_source
 
-    def interp_metric(self, da, axes):
+    def _interp_metric(self, da, axes):
         metric_available = self._metrics.get(frozenset(axes), None)
         if metric_available is not None:
+            # this function works with only one metric at a time
             m = metric_available[0]
             metric_interp = self.interp_like(da, m)
```
jbusecke, line 1399: for a second I was confused if here you are interpolating `da` onto `m`, but then I realized that its the signature of `interp_like` that threw me off. 
I would suggest to change `def interp_like(self, da_target, da_source)` to `def interp_like(self, da_source, da_target)`. This is just a convention thing, where most tools I use emply a left to right (`da_source`->`da_target`) convention. @dcherian @andersy005, do you have an opinion either way?

xgcm/test/test_grid.py
```diff
@@ -782,3 +782,26 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
     actual2 = grid2.interp(ds.data_g, ("X", "Y"))
 
     xr.testing.assert_identical(actual1, actual2)
+
+
+@pytest.mark.parametrize(
+    "metric_axes,metric_name",
+    [
+        ("X", "dx_t"),
+        ("Y", "dy_ne"),
```
jbusecke, line 791: You should be able to add `("['Y', 'X']", "dy_n"), and this should still work. If that does not work, just ignore the comment.

xgcm/test/test_grid.py
```diff
@@ -782,3 +782,26 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
+    ],
+)
+def test_interp_like(metric_axes, metric_name):
+
+    ds, coords, _ = datasets_grid_metric("C")
+
+    grid = Grid(ds, coords=coords)
+    grid.set_metrics(metric_axes, metric_name)
+    metric_available = grid._metrics.get(frozenset(metric_axes), None)
```
jbusecke, line 800: Can you use one of the data variables from the test dataset ('u', 'v', or 'tracer') here instead of the coordinates? I think that makes the code smaller and easier to understand. Basically I would try to get rid of any 'metric' mention in this test. I can help with this.

jbusecke (changes requested):
This is a really usefull addition to xgcm @jdldeauna! I am very excited about this PR and think it is in good shape. I just had a few small typo/formulation suggestions and a bit more extensive renaming for `test_interp_all` (which I would consider useful but not necessary). Almost there!

xgcm/grid.py
dcherian: yeahI like your suggestion. I think we need better names in the signature.

```
def interp_like(self, array, like):
	pass
```
 so this statement could become `self.interp_like(m, like=da)` making it very clear

xgcm/grid.py
```diff
@@ -1361,24 +1361,42 @@ def iterate_axis_combinations(items):
...
 
-        return metric
+        return da_source
 
-    def interp_metric(self, da, axes):
+    def _interp_metric(self, da, axes):
         metric_available = self._metrics.get(frozenset(axes), None)
         if metric_available is not None:
+            # this function works with only one metric at a time
             m = metric_available[0]
```
dcherian, line 1398:
```suggestion
            metric = metric_available[0]
```

v. minor comment: I personally prefer avoiding single letter variable names in most cases. It's more readable this way.

xgcm/test/test_grid.py
jdldeauna: Unfortunately it returned this error: `Fail: KeyError: 'Metric axes ["[\'Y\', \'X\']"] not compatible with grid axes (\'X\', \'Y\', \'Z\')'`

jbusecke: Sorry that was my fault. It should be `['Y', 'X']`, not `"['Y', 'X']"`

xgcm/test/test_grid.py
jdldeauna: Do you mean something like:

```python
interp = grid.interp_like(ds.v, ds.u)
```

where ds.v is the "source" and ds.u is the "target"? I'm not sure how to construct the next statement for comparison since grid.interp works with an axis specification:

```python
test_metric = grid.interp(ds[metric_name], metric_axes)
```

jbusecke: If you use the dictionary style notation (`ds['u']`) you can just define var_name='u' and then you just need to figure out the manual steps required to go from source to target. This will be a single or two axes that need to be interpolated, very similar to the other test.

xgcm/test/test_grid.py
```diff
@@ -786,22 +786,34 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
-
+    interp_metric = grid.interp_like(metric_available, ds.u)
     test_metric = grid.interp(ds[metric_name], metric_axes)
 
     xr.testing.assert_equal(interp_metric, test_metric)
     xr.testing.assert_allclose(interp_metric, test_metric)
+
+
+@pytest.mark.parametrize(
+    "var_name,like_name,var_axes",
```
jbusecke, line 806: This is awesome!

jbusecke (approved):
I really like the new naming. Great job @jdldeauna. This is ready to merge in my eyes.

andersy005 (approved):
Looks pretty good 👍🏽

xgcm/test/test_grid.py
jbusecke: If I had any nit, you could add one or two more combinations here.

dcherian:
Nice work Diane!

doc/whats-new.rst
```diff
@@ -11,7 +11,7 @@ v0.6.0 (unreleased)
 New Features
 ~~~~~~~~~~~~
 - :py:meth:`~xgcm.grid.Grid.set_metrics` now enables adding metrics to a grid object (:pull:`336`, :issue:`199`).
-  By `Dianne Deauna <https://github.com/jdldeauna>`_.
+  By `Dianne Deauna <https://github.com/jdldeauna>`_ under the `SIParCS internship <https://www2.cisl.ucar.edu/siparcs-2021-projects#8>`_.
```
jbusecke, line 15: You do have to give yourself credit for this one for sure!

rabernat:
So fantastic to have a new xgcm contributor. Thanks so much @jdldeauna!

PR #345, opened 2021-06-22
Update get_metric with conditions

jdldeauna:
- [ ] Tests added to test_metrics.py


xgcm/test/test_metrics.py
```diff
@@ -183,3 +183,53 @@ def test_interp_metrics(metric_axes, metric_name):
...
+    ds, coords, _ = datasets_grid_metric("C")
+    grid = Grid(ds, coords=coords)
+    grid.set_metrics(("X", "Y"), "area_e")
+    get_metric = grid.get_metric(ds.v, metric_axes)
+
+    grid = Grid(ds, coords=coords)
+    grid.set_metrics(("X", "Y"), "area_n")
+    test_metric = grid.get_metric(ds.v, metric_axes)
+
+    # The metric is interpolated to the same dimensions, but has different values
```
jdldeauna, line 218: Metric is interpolated to desired dimension, but actual values differ

xgcm/test/test_metrics.py
```diff
@@ -183,3 +183,53 @@ def test_interp_metrics(metric_axes, metric_name):
...
+    grid.set_metrics(("X"), "dx_n")
+    grid.set_metrics(("Y"), "dy_n")
+    get_metric = grid.get_metric(ds.v, metric_axes)
+
+    grid = Grid(ds, coords=coords)
+    grid.set_metrics(("X", "Y"), "area_n")
+    test_metric = grid.get_metric(ds.v, metric_axes)
+
+    # Area calculated by multiplying metrics has different value than given area
+    # xr.testing.assert_equal(get_metric, test_metric)
```
jdldeauna, line 234: Area calculated by multiplying separate metrics has different value than actual given area

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -196,24 +196,24 @@ def test_missingaxis(self, axis, funcname, periodic, boundary):
...
-        match_message = "Unable to find any combinations of metrics for array dims.*X.*"
-        with pytest.raises(KeyError, match=match_message):
-            func(ds.tracer.mean("xt"), "X", **kwargs)
-
-        match_message = (
-            "Unable to find any combinations of metrics for array dims.*X.*Y.*Z.*"
-        )
-        with pytest.raises(KeyError, match=match_message):
-            func(ds.tracer.mean("xt"), ["X", "Y", "Z"], **kwargs)
+    # def test_missingdim(self, funcname, periodic, boundary):
```
jdldeauna, line 199: I'm still working on figuring out how to make this test work, so I've commented it out for now before pushing changes.

xgcm/test/test_metrics.py
```diff
@@ -138,6 +136,105 @@ def test_get_metric(axes, data_var, drop_vars, metric_expected_list, expected_er
         assert metric.equals(expected)
 
 
+@pytest.mark.parametrize(
+    "set_metrics,metric_axes,ds_array,ds_expected,set_metric_vars,expected_metric",
```
dcherian, line 140: This is very impressive use of `parametrize` but it's a hard to follow because your parameters and the actual test code are far away from each other. 

```
if expected_metric == "cond1":
```
I think we should break this test up in to 4 tests where each test tests a single condition. There's not that much code shared between the 4 actual tests so there's not much gained by sticking it in one function.

xgcm/test/test_metrics.py
```diff
@@ -138,6 +136,105 @@ def test_get_metric(axes, data_var, drop_vars, metric_expected_list, expected_er
...
+        metric_interp = grid.interp(metric_values[0], "Y")
+        expected_metric = (metric_interp * metric_values[1]).reset_coords(drop=True)
+    elif expected_metric == "cond4b":
+        metric_values = [ds[v] for k, v in set_metric_vars.items()]
+        metric_interp_1 = grid.interp(metric_values[0], "Y")
+        metric_interp_2 = grid.interp(metric_values[1], "Y")
+        expected_metric = (metric_interp_1 * metric_interp_2).reset_coords(drop=True)
+
+    xr.testing.assert_equal(get_metric, expected_metric)
+    xr.testing.assert_allclose(get_metric, expected_metric)
```
dcherian, line 235: Only one of these is needed: the only difference is whether they use `np.testing.array_equal` or `np.testing.array_allclose` under the hood. The latter allows you to ignore minor floating point differences. If it passes with `equal` use that, if not use `allclose`

jbusecke (approved):
This looks very nice! Thanks @jdldeauna. Lets just remember to add a test for `iterate_axis_combinations`!

PR #350, opened 2021-06-28
Add periodic and boundary conditions to interp_like

jdldeauna:
- [X] Periodic condition added to test_grid/test_interp_like
 - [x] Implement changes to interp_like (include boundary conditions as input)
 - [x] Boundary conditions  and fill value added to test_grid/test_interp_like
 - [x] Include a test for metics.py/iterate_axis_combinations


xgcm/test/test_grid.py
```diff
@@ -786,18 +786,19 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
     grid.set_metrics(metric_axes, metric_name)
     metric_available = grid._metrics.get(frozenset(metric_axes), None)
     metric_available = metric_available[0]
     interp_metric = grid.interp_like(metric_available, ds.u)
-    test_metric = grid.interp(ds[metric_name], metric_axes)
+    expected_metric = grid.interp(ds[metric_name], metric_axes)
 
-    xr.testing.assert_equal(interp_metric, test_metric)
-    xr.testing.assert_allclose(interp_metric, test_metric)
+    if xr.testing.assert_equal(interp_metric, expected_metric) is False:
```
dcherian, line 800: assert works by raising an error so this won't work.

I would just keep the `allclose` version; it's the right one for floating point comparisons

dcherian: If you really wanted to do that comparison, do it this way:
```
if interp_metric.equals(expected_metric) is False:
```

but for here, `assert_allclose` should be good enough

xgcm/grid.py
```diff
@@ -1388,6 +1387,9 @@ def interp_like(self, array, like):
             Source data array with updated positions along axes matching with target array
         """
 
+        boundary = self._parse_axes_kwargs(boundary)
+        fill_value = self._parse_axes_kwargs(fill_value)
+
```
jdldeauna, line 1392: Using `interp_axes` and `interp_axes.append(axname)` results in a type of double interpolation, where array gets interpolated on to the first axis, and then as the loop continues to the second axis, because axname is appended to interp_axes, the array gets interpolated on to the first axis a second time. The correct output (from the first interpolation) is overwritten as result.

xgcm/test/test_grid.py
```diff
@@ -784,20 +784,62 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
+        (
+            {"X": "extrapolate", "Y": "extrapolate"},
+            {"X": "extrapolate", "Y": "extrapolate"},
+        ),
+        ("fill", {"X": "fill", "Y": "fill"}),
+        ("extend", {"X": "extend", "Y": "extend"}),
+        ("extrapolate", {"X": "extrapolate", "Y": "extrapolate"}),
+        ({"X": "extend", "Y": "fill"}, {"X": "extend", "Y": "fill"}),
+        ({"X": "extrapolate", "Y": "fill"}, {"X": "extrapolate", "Y": "fill"}),
+        pytest.param(
```
jdldeauna, line 804: This set of parameters causes the test to fail under `periodic=True` (therefore pytest.mark.xfail works) but passes under `periodic=False`

jdldeauna: Maybe create a separate test for those expected to fail

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,29 @@ def test_assign_metric(key, metric_vars):
...
+    axes = ["X", "Y"]
+    iterate_metric = []
+    for axis_combinations in iterate_axis_combinations(axes):
+        try:
+            append_axis = list(axis_combinations)
+            iterate_metric.append(append_axis)
+        except KeyError:
+            pass
+
+    expected_metric = [
```
jdldeauna, line 111: Need to try out different inputs to discover fail cases

xgcm/test/test_metrics.py
```diff
@@ -178,7 +202,7 @@ def test_get_metric_with_conditions_03():
     metric_var_2 = ds.dz_t
```
jbusecke, line 202: I wonder if you should assign all metrics like this `grid = Grid(ds, coords=coords, metrics=metrics)`. This makes the code smaller and also ensures that no wrong metrics are picked even if they are available. I believe this should be changed in all of the tests. Pretty sure that wont break anything, either.

xgcm/test/test_metrics.py
```diff
@@ -284,7 +308,7 @@ def test_interp_metrics(metric_axes, metric_name):
     grid.set_metrics(metric_axes, metric_name)
     interp_metric = grid._interp_metric(ds.u, metric_axes)
```
jbusecke, line 309: This one needs to be challenged with the boundary conditions similarly to `interp_like`.

xgcm/grid.py
```diff
@@ -1388,6 +1387,9 @@ def interp_like(self, array, like):
...
         for axname, axis in self.axes.items():
             # This will raise a KeyError since this for-loop goes through all axes contained in self,
```
jbusecke, line 1394: Seems like a duplicate comment?

xgcm/grid.py
```diff
@@ -1399,7 +1401,12 @@ def interp_like(self, array, like):
             except KeyError:
```
jbusecke, line 1401: Why would we continue here if the `_get_axis_coord` raises an error?

xgcm/grid.py
```diff
@@ -1399,7 +1401,12 @@ def interp_like(self, array, like):
...
-                array = self.interp(array, axname)
+                array = self.interp(
+                    array,
+                    axname,
+                    fill_value=fill_value.get(axname, None),
+                    boundary=boundary.get(axname),
+                )
         return array
 
     def _interp_metric(self, da, axes):
```
jbusecke, line 1412: How do we handle boundaries here?

jbusecke (changes requested):
Hey @jdldeauna this is some great progress. I think the `_interp_metric` logic needs some more work though. 
I would suggest that you split this work into two PRs:
1) This one here should only contain new tests and logic pertinent to `interp_like` and boundaries (which looks almost ready to me)
2) The second PR would build on that and perhaps add some clever defaults for `boundary`/`fill` within `_interp_metric`.

xgcm/grid.py
```diff
@@ -1399,7 +1401,12 @@ def interp_like(self, array, like):
             except KeyError:
                 continue
             if position_like != position_array:
-                array = self.interp(array, axname)
+                array = self.interp(
+                    array,
+                    axname,
+                    fill_value=fill_value.get(axname, None),
```
dcherian, line 1407:
```suggestion
                    fill_value=fill_value.get(axname),
```

`None` is the default value returned

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,29 @@ def test_assign_metric(key, metric_vars):
...
+    expected_metric = [
+        [frozenset({"X", "Y"})],
+        [frozenset({"X"}), frozenset({"Y"})],
+        [frozenset({"Y"}), frozenset({"X"})],
+    ]
+
+    iterate_metric.sort()
+    expected_metric.sort()
+
+    assert iterate_metric == expected_metric
```
dcherian, line 120:
```suggestion
    assert set(iterate_metric) == set(expected_metric)
```

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,29 @@ def test_assign_metric(key, metric_vars):
     _ = Grid(ds, coords=coords, metrics={key: metric_vars})
 
 
+def test_iterate_axis_combinations():
+
+    axes = ["X", "Y"]
+    iterate_metric = []
```
dcherian, line 103: use `set(iterate_axis_combinations(...))` instead

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,29 @@ def test_assign_metric(key, metric_vars):
...
+    for axis_combinations in iterate_axis_combinations(axes):
+        try:
+            append_axis = list(axis_combinations)
+            iterate_metric.append(append_axis)
+        except KeyError:
+            pass
+
+    expected_metric = [
+        [frozenset({"X", "Y"})],
+        [frozenset({"X"}), frozenset({"Y"})],
```
dcherian, line 113: need tuples instead of list

xgcm/test/test_metrics.py
dcherian, line 111: needs ot be set instead of list

xgcm/test/test_grid.py
```diff
@@ -784,20 +784,62 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
 
 @pytest.mark.parametrize(
     "metric_axes,metric_name",
-    [("X", "dx_t"), ("Y", "dy_ne"), (["Y", "X"], "dy_n")],
+    [(["Y", "X"], "area_n")],
```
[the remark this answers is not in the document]
jdldeauna: Hi, so I just wanted to get one set of metric axes and metric name going with all the parameters before I put in other inputs because testing takes so long for me X_____x So basically I will be trying out one pair of metric axes and name at a time with all of the parameters, and then for the final PR, I'll include all metric axes and name pairs.

xgcm/grid.py
jdldeauna: There could be an axis present in `grid` that is not available for either `array` or `like`. But usually for the axis you are interested in, you would make sure that axis is available for all three.

xgcm/grid.py
jdldeauna: Boundary can either be a string or dictionary, and the line at the beginning of the loop:
```python
boundary = self._parse_axes_kwargs(boundary)
```
will assign boundary values and then you can select per axis name inside the loop `boundary=boundary.get(axname)`

xgcm/test/test_grid.py
```diff
@@ -784,20 +784,62 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
+@pytest.mark.parametrize("fill_value", [None, 0.1])
+def test_interp_like(
+    metric_axes, metric_name, periodic, boundary, boundary_expected, fill_value
+):
 
     ds, coords, _ = datasets_grid_metric("C")
-    grid = Grid(ds, coords=coords)
+    grid = Grid(ds, coords=coords, periodic=periodic)
     grid.set_metrics(metric_axes, metric_name)
     metric_available = grid._metrics.get(frozenset(metric_axes), None)
```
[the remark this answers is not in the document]
jdldeauna: Hmm I think we do apply the test to `ds.u` (see lines 834 to 835). Do you mean assigning grid as

```python
grid = Grid(ds.u, coords=coords, periodic=periodic)
```

then doing the test as:
```python
interp_metric = grid.interp_like(metric_available, ds, boundary=boundary, fill_value=fill_value)
```

xgcm/test/test_grid.py
```diff
@@ -811,7 +853,7 @@ def test_interp_like_var(var_name, like_name, var_axes):
     ds, coords, metrics = datasets_grid_metric("C")
```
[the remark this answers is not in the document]
jdldeauna: Oh okay I see, I can consolidate that. I'm not sure how it would change with the boundary and fill_value inputs, I'll let you know.

xgcm/test/test_metrics.py
```diff
@@ -161,7 +185,7 @@ def test_get_metric_with_conditions_02():
     metric_var_2 = ds.dy_n
     expected_metric = (metric_var_1 * metric_var_2).reset_coords(drop=True)
 
-    if xr.testing.assert_equal(get_metric, expected_metric) is False:
+    if get_metric.equals(expected_metric) is False:
```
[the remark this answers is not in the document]
jdldeauna: Yes Deepak has also said it's confusing, I'll just be removing all of the get_metric.equals() statements and using `xr.testing.assert_allclose` for all the tests.

xgcm/grid.py
dcherian: I think you could just move the `interp` line outside the for loop? Or provide `axname` instead of `interp_axes` so you interp one axis at a time as you loop

jdldeauna: Right, so moving the interp line outside the for loop worked, thanks! 😅

xgcm/test/test_grid.py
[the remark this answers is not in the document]
jdldeauna: It worked! 🎉

xgcm/grid.py
jbusecke: Thanks for the explanation, that makes sense. Could you add this to the comment?

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,38 @@ def test_assign_metric(key, metric_vars):
...
+            ("X", "Y"),
+            (
+                (frozenset({"X", "Y"}),),
+                (frozenset({"X"}), frozenset({"Y"})),
+                (frozenset({"Y"}), frozenset({"X"})),
+            ),
+        ),
+        (
+            ("X", "Y", "Z"),
+            (
```
jdldeauna, line 113: I'm having difficulty with parameterizing this test, because when I use three axes `("X","Y","Z")` each time I run `iterate_axis_combinations` the output can be slightly different. For example:
```python
(frozenset({"Y","X"}), frozenset({"Z}))
```
can become:
```python
(frozenset({"X","Y"}), frozenset({"Z}))
```
It becomes difficult to set `expected_axes` because I can't really predict the exact output from `iterate_axis_combinations`. I might have to rethink how the test is being done.

dcherian:
```
a = {
    (frozenset({"Y", "X"}), frozenset({"Z"})), (frozenset({"X", "Z"}), frozenset({"Y"}))
}
b = {
    (frozenset({"X", "Y"}), frozenset({"Z"})), (frozenset({"Z", "X"}), frozenset({"Y"}))
}

a == b
```

is `True` so I'm confused about what you're running in to.

jdldeauna: When I use
```python
assert iterate_axes == expected_axes
```
I get an `AssertionError: Use -v to get the full diff`, so I tried using
```python
set_diff = iterate_axes.difference(expected_axes)
assert set_diff == set()
```
which works for `axis = ("X","Y")` but not for `axis = ("X","Y","Z")`, and I noticed that the order of the frozensets would change sometimes so I thought that might be it, but maybe something else is causing the error.

jdldeauna: I tried adding these elements to `expected_axes` and the test passed. I think I was too preoccupied with the debug function and trying to get the sets to match exactly each time. X_________x

andersy005 (approved):
Looks good to me 👍🏽

xgcm/grid.py
```diff
@@ -1381,28 +1380,52 @@ def interp_like(self, array, like):
...
+            * 'fill':  Set values outside the array boundary to fill_value
+              (i.e. a Dirichlet boundary condition.)
+            * 'extend': Set values outside the array to the nearest array
+              value. (i.e. a limited form of Neumann boundary condition where
+              the difference at the boundary will be zero.)
+            * 'extrapolate': Set values by extrapolating linearly from the two
+              points nearest to the edge
+            This sets the default value. It can be overriden by specifying the
+            boundary kwarg when calling specific methods.
+        fill_value : {float}, optional
```
dcherian, line 1396:
```suggestion
        fill_value : float, optional
```

xgcm/test/test_grid.py
```diff
@@ -784,34 +784,52 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
-    test_metric = grid.interp(ds[metric_name], metric_axes)
-
-    xr.testing.assert_equal(interp_metric, test_metric)
-    xr.testing.assert_allclose(interp_metric, test_metric)
-
-
+@pytest.mark.parametrize("periodic", [True, False])
 @pytest.mark.parametrize(
-    "var_name,like_name,var_axes",
+    "boundary, boundary_expected",
```
dcherian, line 797: I don't think we need `boundary_expected` at all? You can just pass `boundary` to `interp`?

xgcm/test/test_grid.py
```diff
@@ -784,34 +784,52 @@ def test_boundary_kwarg_same_as_grid_constructor_kwarg():
...
+    ds, coords, _ = datasets_grid_metric("C")
+    grid = Grid(ds, coords=coords, periodic=periodic)
+    grid.set_metrics(metric_axes, metric_name)
+    metric_available = grid._metrics.get(frozenset(metric_axes), None)
+    metric_available = metric_available[0]
+    interp_metric = grid.interp_like(
+        metric_available, ds.u, boundary=boundary, fill_value=fill_value
+    )
+    expected_metric = grid.interp(
+        ds[metric_name], metric_axes, boundary=boundary_expected, fill_value=fill_value
```
dcherian, line 832:
```suggestion
        ds[metric_name], metric_axes, boundary=boundary, fill_value=fill_value
```

xgcm/test/test_metrics.py
```diff
@@ -96,6 +97,42 @@ def test_assign_metric(key, metric_vars):
...
+                (frozenset({"X", "Y"}), frozenset({"Z"})),
+            ),
+        ),
+    ],
+)
+def test_iterate_axis_combinations(axes, expected_axes):
+
+    iterate_axes = set(iterate_axis_combinations(axes))
+    set_diff = iterate_axes.difference(expected_axes)
+    assert set_diff == set()
```
dcherian, line 133: We should be able to directly check equality?

```suggestion
    assert iterate_axes == expected_axes
```

dcherian (approved):
Just some minor suggestions. Nice work!

xgcm/test/test_grid.py
jdldeauna: There's a test where boundary = 'fill' and the boundary_expected is {"X": "fill", "Y": "fill"}, and we want to make sure that the boundaries are still assigned correctly. There's also the pytest.mark.xfail case where `boundary` and `boundary_expected` are not the same, so for these cases I think it's worth it to have them as separate variables.

xgcm/test/test_metrics.py
jdldeauna: When I use `assert iterate_axes == expected_axes`, I get the ff error:
```python
>       assert iterate_axes == expected_axes
E       AssertionError: assert {(frozenset({...{'X', 'Y'}),)} == ((frozenset({...enset({'X'})))
E         Use -v to get the full diff
```
But when I use set difference the test passes. Does it pass for you if you directly check equality?

dcherian: Ah two things.

1. `expected_axes` is a tuple so `assert iterate_axes == set(expected_axes)` works for the first case.
2. The set difference is not symmetric! `iterate_axes.difference(expected_axes)` will be empty if there are no elements in `iterate_axes` that are not in `expected_axes`. But `expected_axes` could have extra elements and that difference would still be an empty set. You could use `symmetric_difference`. But it should fail just as direct equality fails for the second set. 

So the second set of parameters needs to be fixed.

jdldeauna: I see, thanks Deepak!

jdldeauna: I get AssertionErrors because when I run iterate_axis_combinations it will produce only one of this type of combination, and `assert iterate_axes == set(expected_axes)` doesn't count these as equal:
```python
(frozenset({'Y'}), frozenset({'Z'}), frozenset({'X'}))
(frozenset({'Y'}), frozenset({'X'}), frozenset({'Z'}))
```
and you can't predict which one iterate_axis_combination will produce so setting up `expected_axes` feels like whack-a-mole as I try to get the left and right set to match. 😅 I'll try to see what can be a more efficient way to test this.

dcherian: I think that means you want to convert things to a set one level down. So that a single element is `set(frozenset({'Y'}), frozenset({'Z'}), frozenset({'X'}))`

jdldeauna: If I change an element into a set: `set((frozenset({"Z"}), frozenset({"Y"}), frozenset({"X"})))`
There's a TypeError: 
```python 
E       TypeError: unhashable type: 'set'
```
if I do: `set(frozenset({"Z"}), frozenset({"Y"}), frozenset({"X"}))`
I also get a TypeError:
```python
E   TypeError: set expected at most 1 argument, got 3
```

jdldeauna: I think the main issue is that `iterate_axis_combinations` will always produce a set of tuples. So an element of the result (of type tuple) can either be:
```python
(frozenset({"Z"}), frozenset({"X"}), frozenset({"Y"}))
(frozenset({"Z"}), frozenset({"Y"}), frozenset({"X"}))
```
And since tuples are compared position by position, `expected_axes` can only contain one of the above so that
```python
assert iterate_axes == set(expected_axes)
```
will result in `True`. Unfortunately, there's no real way to predict what `iterate_axis_combinations` will produce, especially if you factor in the other combinations such as `(frozenset({"Y"}), frozenset({"X"}), frozenset({"Z"})),` and `(frozenset({"X"}), frozenset({"Y"}), frozenset({"Z"})),`. We can discuss this further, but I think the solution will be to keep the test limited to `axes = ("X","Y")`.

dcherian: Alright I fixed it a hacky way. Now we can move on to bigger and better things =)

jbusecke (approved):
This looks great. Thanks for the hard work on this @jdldeauna!

PR #351, opened 2021-07-07
Allow for overwriting metrics in set_metrics method

jdldeauna:
- [X] Include overwrite kwarg in set_metrics
 - [X] Tests added in test_metrics/test_set_metric_overwrite_true and test_set_metric_overwrite_false


xgcm/grid.py
```diff
@@ -1424,15 +1452,6 @@ def interp_like(self, array, like, boundary=None, fill_value=None):
         )
         return array
 
-    def _interp_metric(self, da, axes):
```
jdldeauna, line 1427: I think this functionality will be implemented once interp_like is combined with get_metric.

xgcm/grid.py
```diff
@@ -1268,19 +1268,49 @@ def set_metrics(self, key, value):
...
+                    )
+                else:
+                    value_new = self._ds[metric_varname].reset_coords(drop=True)
+                    # check for duplicate values
+                    if any(value_new.equals(ve) for ve in value_exist):
+                        # go through each existing value until data array with matching dimensions is selected
+                        for idx, ve in enumerate(value_exist):
+                            # double check if dimensions match
+                            if set(value_new.dims) == set(ve.dims):
+                                warnings.warn(
```
jbusecke, line 1289: This should only be triggered if overwrite is true? I think you just need to move it a line down under `if overwrite:`. Could you add this to the tests,  and check if this is triggered only if the value is actually overwritten?

xgcm/grid.py
jbusecke: yes lets keep this in another PR.

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,94 @@ def test_set_metric():
...
 
 
-@pytest.mark.skip(reason="For the next PR")
 @pytest.mark.parametrize(
-    "metric_axes,metric_name",
+    "metric_axes, metric_varname, overwrite_",
     [
-        ("X", "dx_t"),
-        ("Y", "dy_ne"),
+        ("X", ["dx_n", "dx_t", "dx_ne", "dx_t"], True),
```
jbusecke, line 319: I might split this into two tests: One for overwrite=True and another for overwrite=False. Each of these could test that the appropriate warning/error is raised each time (see my comment above). We can discuss this later too.

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,94 @@ def test_set_metric():
...
-    # need to test with boundary and fill_value conditions
+def test_set_metric_overwrite(metric_axes, metric_varname, overwrite_):
     ds, coords, metrics = datasets_grid_metric("C")
     grid = Grid(ds, coords=coords)
-    grid.set_metrics(metric_axes, metric_name)
-    interp_metric = grid._interp_metric(ds.u, metric_axes)
 
-    expected_metric = grid.interp(ds[metric_name], metric_axes)
+    for mv in metric_varname:
+        grid.set_metrics(metric_axes, mv, overwrite=overwrite_)
```
jbusecke, line 337: Is this actually overwriting anything? I dont think you are setting the metric before?

To really test that a new dataarray is set I would do the following in the test:
1) Set all metrics that are in the dataset with `grid=Grid(ds, ..., metrics=metrics)`
2) Modify a metric and assign it under a new name. For example:
```python
metric_varname_modified = f'{metric_varname}_overwrite'
ds = ds.assign_coords(metric_varname_modified, ds[metric_varname]*10)
```
3) Then assign this modified metric to the dataset with `grid.set_metric(metric_axes, metric_varname_modified)`. You will have to write this in a loop, this is just an example. 
4) Now you can check explicitly that the new metric is actually 10 times as large. Without that you cant be really sure that the metric has been replaced!

jbusecke (changes requested):
Thanks for this @jdldeauna. As far as I can see the code changes look good, but I think the tests need some modifications to ensure the behavior we want. Happy to help further with this if anything is unclear.

xgcm/grid.py
jdldeauna: So right now the flow is: 
1) check if there is input which matches an existing metric (Line 1284)
2) cycle through existing metrics to find the match 
3) check if the dimensions between input and existing metric match 
4) replace if the value of overwrite = True
Just to clarify would you want to switch steps 3 and 4?

jbusecke: Oh no. I am just concerned about the warning. 

Currently you have somthing like this:
```python
if set(value_new.dims) == set(ve.dims):
    warn(... replacing...).# this will be triggered regardless of whether overwrite is true or false
    if overwrite:
        # replace metric
    else:
       raise error
```

What I suggest is simply to move the warning into the if statement. Otherwise you will get a very confusing output for overwrite false. It will basically say: "I am overwriting" (which it isnt) and then fail.

jbusecke: What I would do here is again start with the test: Make a test that checks that this warning is raised when overwrite=True, and more importanty: Make sure it is **not** raised when overwrite=False. I believe this will fail with the current code, and then pass when you move the warning.

xgcm/test/test_metrics.py
jdldeauna: Hi, I was trying to incorporate this into the tests, and I wanted to clarify. Is the purpose of the test to:
1. Add a new metric that is 10x the value of the old one (e.g., you have both `dx_t` and `dx_t_overwrite` as metrics)
2. Replace an existing metric with something named similarly but 10x in value (e.g., you end up with one `dx_t` metric that's 10x in value as the original one from metrics=metrics)

jbusecke: Yes that way you can be sure that what you are checking is not just the metric set in the first place.

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,86 @@ def test_set_metric():
...
+
+    ds = ds.assign_coords({"dx_t_overwrite": ds["dx_t"] * 10})
+
+    grid = Grid(ds, coords=coords, metrics=metrics)
+
+    with pytest.raises(ValueError, match="setting overwrite=True."):
+        grid.set_metrics("X", "dx_t_overwrite", overwrite=False)
+
+
+def test_set_metric_replace():
```
jdldeauna, line 367: I couldn't integrate this test with `test_set_metric_replace` because if metrics are set as `metrics=metrics` from the beginning, I can't append a new metric when its dimensions already matches an existing metric (can only overwrite).

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,86 @@ def test_set_metric():
...
+    ds, coords, metrics = datasets_grid_metric("C")
+
+    mv_modified = f"{metric_varname}_overwrite"
+    ds = ds.assign_coords({mv_modified: ds[metric_varname] * 10})
+
+    grid = Grid(ds, coords=coords, metrics=metrics)
+
+    grid.set_metrics(metric_axes, mv_modified, overwrite=True)
+
+    if len(metric_axes) > 1:
```
jdldeauna, line 337: Hacky approach to using `metric_axes = ("X","Y")`😅  
Feel free to suggest an alternative!

xgcm/test/test_metrics.py
jbusecke: You dont have to set all the metrics in the beginning. How about you modify the `metrics` input dict according to your input parameters:
```python
metric_names = ['dx_t', 'dx_ne'] #not sure this is the right variable name, just a sketch
ds, coords, metrics = ....
metrics = {k:[m for m in v if m in metric_names] for k,v in metrics.items()} # this will then only have the above names in it.
grid = Grid(..., metrics=metrics, ...)
```

xgcm/test/test_metrics.py
jbusecke: does `frozenset(list(metric_axes))` work in both cases?

jdldeauna: It worked!

xgcm/grid.py
```diff
@@ -1268,19 +1268,46 @@ def set_metrics(self, key, value):
...
+                                    f" Overwrite {metric_varname} by setting overwrite=True."
+                                )
+                    # if no existing value matches new value dimension-wise, just append new value
+                    if not did_overwrite:
+                        self._metrics[metric_axes].append(value_new)
+        else:
+            # initialize empty list
+            self._metrics[metric_axes] = []
+            for metric_varname in metric_value:
+                if metric_varname not in self._ds:
```
jbusecke, line 1304: These few lines of code are repeated above. Can we factor them out into a small helper method? Something like:
```
def _maybe_get_var(self, varname):
     if varname not in self._ds:
         raise KeyError(f"Variable {varname} not found in dataset.")
    else:
        return self._ds[varname].reset_coords(drop=True)
```
and then call it as `metric_var = self._maybe_get_var(metric_varname)` in both cases?

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,71 @@ def test_set_metric():
...
+            ["dx_n_overwrite", "dx_ne"],
+            ["dx_t", "dx_n_overwrite", "dx_e", "dx_ne"],
+        ),
     ],
 )
-def test_interp_metrics(metric_axes, metric_name):
-    # need to test with boundary and fill_value conditions
+def test_set_metric_overwrite_true(
+    metric_axes, exist_metric_varname, add_metric_varname, expected_varname
+):
```
jbusecke, line 342: I ❤️ this test! Great job.

jbusecke (approved):
Hey @jdldeauna, this is in great shape! I had one (optional) suggestion to make the code even cleaner, but I would also be happy to merge this as is!

xgcm/grid.py
```diff
@@ -1268,19 +1268,46 @@ def set_metrics(self, key, value):
...
-            # check for duplicate dimensions among each axis metric
-            self._metrics[metric_axes].append(metric_var)
+        if metric_axes in existing_metric_axes:
+            value_exist = self._metrics.get(metric_axes)
+            for metric_varname in metric_value:
+                if metric_varname not in self._ds:
+                    raise KeyError(
+                        f"Metric variable {metric_varname} not found in dataset."
+                    )
+                else:
```
dcherian, line 1281: could remove this else and avoid one level of indentation. It will only run if the `if` condition is not True.

xgcm/grid.py
```diff
@@ -1268,19 +1268,46 @@ def set_metrics(self, key, value):
...
+                        if set(value_new.dims) == set(ve.dims):
+                            if overwrite:
+                                # replace existing data array with new data array input
+                                self._metrics[metric_axes][idx] = value_new
+                                did_overwrite = True
+                            else:
+                                raise ValueError(
+                                    f"Metric variable {metric_varname} with dimensions {ve.dims} already assigned in metrics."
+                                    f" Overwrite {metric_varname} by setting overwrite=True."
+                                )
```
dcherian, line 1296: `metric_varname` is the "new" metric the user is trying to assign. I think you want to print `ve.name` instead.


```suggestion
                                    f"Metric variable {ve.name} with dimensions {ve.dims} already assigned in metrics."
                                    f" Overwrite {ve.name} with {metric_varname} by setting overwrite=True."
                                )
```

xgcm/grid.py
```diff
@@ -1268,19 +1268,46 @@ def set_metrics(self, key, value):
...
-        self._metrics[metric_axes] = []
-        for metric_varname in _maybe_promote_str_to_list(value):
-            if metric_varname not in self._ds:
-                raise KeyError(
-                    f"Metric variable {metric_varname} not found in dataset."
-                )
-            # resetting coords avoids potential broadcasting / alignment issues
-            metric_var = self._ds[metric_varname].reset_coords(drop=True)
+        metric_value = _maybe_promote_str_to_list(value)
+        existing_metric_axes = set(k for k in self._metrics.keys())
```
dcherian, line 1272:
```suggestion
        existing_metric_axes = set(self._metrics.keys())
```

should also work? `dict.keys()` and `dict.values()` are "iterables" so you can cast them to `tuple` `set` and `list` directly.

xgcm/grid.py
```diff
@@ -1268,19 +1268,46 @@ def set_metrics(self, key, value):
...
+                            else:
+                                raise ValueError(
+                                    f"Metric variable {metric_varname} with dimensions {ve.dims} already assigned in metrics."
+                                    f" Overwrite {metric_varname} by setting overwrite=True."
+                                )
+                    # if no existing value matches new value dimension-wise, just append new value
+                    if not did_overwrite:
+                        self._metrics[metric_axes].append(value_new)
+        else:
+            # initialize empty list
```
dcherian, line 1301:
```suggestion
            # no metrics for metric_axes yet; initialize empty list
```

I was confused about when this else is being triggered. Another reason to reduce indentation above,

xgcm/grid.py
dcherian: Alternatively we can just make sure that  all `metric_varname` entries are `in self._ds.variables` right at the top before we loop through so we only ever write this code once.

`.variables` includes both `.data_vars` and `.coords`

xgcm/test/test_metrics.py
dcherian: agreed! :heart:

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,71 @@ def test_set_metric():
...
     ds, coords, metrics = datasets_grid_metric("C")
-    grid = Grid(ds, coords=coords)
-    grid.set_metrics(metric_axes, metric_name)
-    interp_metric = grid._interp_metric(ds.u, metric_axes)
 
-    expected_metric = grid.interp(ds[metric_name], metric_axes)
+    ds = ds.assign_coords({add_metric_varname[0]: ds[exist_metric_varname[1]] * 10})
+
+    metrics = {
+        k: [m for m in v if m in exist_metric_varname] for k, v in metrics.items()
```
dcherian, line 348: haha list comprehension in a dict comprehension. dang!

xgcm/test/test_metrics.py
```diff
@@ -311,28 +307,71 @@ def test_set_metric():
...
+        metric_var = ds[ev].reset_coords(drop=True)
+        expected_metric.append(metric_var)
+
+    assert len(set_metric) == len(expected_metric)
+
+    for i in range(len(set_metric)):
+        assert set_metric[i].equals(expected_metric[i])
+
+
+def test_set_metric_overwrite_false():
```
dcherian, line 369:
```suggestion
def test_set_metric_errors():
```

We should also test `grid.set_metrics("X", "foo")` raises an error.

What should `grid.set_metrics("X", "dx_e", overwrite=False)` do? It's identical to an already set metric.

dcherian (commented):
looking great.

xgcm/test/test_metrics.py
jbusecke: 👾

xgcm/test/test_metrics.py
jdldeauna: Yehey thank you! ❤️

xgcm/test/test_metrics.py
jdldeauna: 👹

xgcm/grid.py
```diff
@@ -1262,50 +1262,46 @@ def check_neighbor(link, position):
...
             raise KeyError(
                 f"Metric axes {axes_not_found!r} not compatible with grid axes {tuple(self.axes)!r}"
             )
 
         metric_value = _maybe_promote_str_to_list(value)
-        existing_metric_axes = set(self._metrics.keys())
+        for metric_varname in metric_value:
+            if (metric_varname not in self._ds.variables) and (
+                metric_varname not in self._ds.coords
+            ):
```
dcherian, line 1274:
```suggestion
            if metric_varname not in self._ds.variables:
```

that should check all variables including those in `.coords`

dcherian (approved):

dcherian:
Great nice job Diane!

go ahead and merge since you have two approvals

jdldeauna:
> Great nice job Diane!
> 
> go ahead and merge since you have two approvals

Did a small tweak to a test before merging, thanks Deepak!

Issue #145, opened 2019-07-23
Interpolation (not centred)


rcaneill:
Hello,
I am trying to figure how to do interpolation that are not centred, e.g. when a "center" point is not at the middle point of 2 "right" points. I tried different ways but I did not manage to find a satisfying way that would fit in the xgcm.Axis class.
My idea was to implement a new function for interpolation between data_left and data_right, taking also a weight representing the distance to the new position for each data, and use it instead of `raw_interp_function`:

```python
def interp_function(data_left, data_right, weight_left, weight_right):
    return (data_left * weight_left + data_right * weight_right) / (weight_left + weight_right)
```

For an interpolation from "right" to "center", weights would be (C: "center, R: "right", wl: "weight_left", wr: "weight_right"):

```
    |-----------------|----------------|----------------|
            C         R           C    R           C    R
                  data_left        data_right           
                       <---------> <--> 
                            wr      wl 
```

I did not really find how to implement this solution without changing too much code, as I am not yet familiar with xgcm. Maybe it is necessary to reshape some parts.
Does someone has an idea of the rough lines? Or another solution?

rabernat:
Hi @rcaneill - thanks for this useful idea.

So far, xgcm only supports grids in which the variables are assumed to be centered. But I know this is not the case for some models, like MOM5 / MOM6. Could you let us know what is your specific use case here?

As you have already seen, the interpolation function is exceedingly simple:

https://github.com/xgcm/xgcm/blob/cac66afeb2426546b142aef07b7752bd02ea0ebb/xgcm/grid.py#L1249-L1252

I would definitely be open to extending this to the more general case where the interpolation weights vary spatially. There would be some design questions to answer here, such as
- Are there any existing metadata standards for providing weights in netCDF files?
- Where in xgcm's data model should the weights be stored?
- How should the user-facing API specify these weights?

rcaneill:
Hi,
I am using NEMO in s coordinates, where the vertical variable are not necessarily centered.
I'll think about the design questions and keep you in touch if I find a clean solution.

jbusecke:
Can this be achieved in principal with #344 ?

jbusecke:
@andersy005 This now seems too early? Shouldnt my post count as activity?

andersy005:
> @andersy005 This now seems too early? Shouldnt my post count as activity?

You're right... My hunch is that this is related to this bug: https://github.com/actions/stale/issues/509. I will look into it sometime next week.

[...]

PR #352, opened 2021-07-08
Update stale workflow

andersy005:
Addresses https://github.com/xgcm/xgcm/issues/145#issuecomment-873505715. We should wait for a fix for https://github.com/actions/stale/issues/509 before merging this... 


jbusecke:
Thanks @andersy005

jbusecke (approved):
Thanks @andersy005

PR #353, opened 2021-07-09
Fix Arakawa grids Wikipedia page link

yz3062:
A simple link fix for a Wikipedia URL.


jbusecke (approved):
Thanks a lot for the fix @yz3062

PR #354, opened 2021-07-12
Integrate interp_like with get_metric

jdldeauna:
- [x] Enable interpolation for metrics using get_metric 
- [x] Tests added to test_metrics.py with different inputs (test_get_metric_with_conditions)
- [x] Updated test_metric_ops_multi_axis
- [x] Rename _get_axis_coord to _get_position_name

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1369,6 +1369,9 @@ class Grid:
                     break
             if metric_vars is None:
                 # Condition 2: interpolate metric with matching axis to desired dimensions
+                warnings.warn(
+                    "Metric outputs are interpolated from initial metrics input"
+                )
                 metric_vars = self.interp_like(mv, array, "extend", None)
         else:
             for axis_combinations in iterate_axis_combinations(axes):
@@ -1389,6 +1392,9 @@ class Grid:
                             break
                         else:
                             # Condition 4: metrics in the wrong position (must interpolate before multiplying)
+                            warnings.warn(
+                                "Metric outputs are interpolated from initial metrics input"
+                            )
                             metric_vars = tuple(
                                 self.interp_like(pc, array, "extend", None)
                                 for pc in possible_combinations
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1305,22 +1305,19 @@ class Grid:
 
     def _get_dims_from_axis(self, da, axis):
         dim = []
+        axis = _maybe_promote_str_to_list(axis)
         for ax in axis:
-            set_ax = frozenset(_maybe_promote_str_to_list(ax))
-            if set_ax.issubset(set(self.axes.keys())):
+            if ax in self.axes:
                 all_dim = self.axes[ax].coords.values()
                 matching_dim = [di for di in all_dim if di in da.dims]
                 if len(matching_dim) == 1:
                     dim.append(matching_dim[0])
                 else:
                     raise ValueError(
-                        "Did not find single matching dimension %s from %s corresponding to axis %s. Got (%s)"
-                        % (da.dims, da.name, ax, matching_dim)
+                        f"Did not find single matching dimension {da.dims} from {da.name} corresponding to axis {ax}, got {matching_dim}."
                     )
             else:
-                raise KeyError(
-                    "Did not find axis %s from data array %s" % (ax, da.name)
-                )
+                raise KeyError(f"Did not find axis {ax} from data array {da.name}")
         return dim
 
     def get_metric(self, array, axes):
@@ -1331,8 +1328,7 @@ class Grid:
         Parameters
         ----------
         array : xarray.DataArray
-            The array for which we are looking for a metric. Only its
-            dimensions are considered.
+            The array for which we are looking for a metric. Only its dimensions are considered.
         axes : iterable
             A list of axes for which to find the metric.
 
@@ -1387,11 +1383,9 @@ class Grid:
                             break
                         else:
                             # Condition 4: metrics in the wrong position (must interpolate before multiplying)
-                            possible_combinations_dims = [
-                                pc.dims for pc in possible_combinations
-                            ]
+                            possible_dims = [pc.dims for pc in possible_combinations]
                             warnings.warn(
-                                f"Metric at {array.dims} being interpolated from metrics at dimensions {possible_combinations_dims}. Boundary value set to 'extend'."
+                                f"Metric at {array.dims} being interpolated from metrics at dimensions {possible_dims}. Boundary value set to 'extend'."
                             )
                             metric_vars = tuple(
                                 self.interp_like(pc, array, "extend", None)
@@ -1824,7 +1818,7 @@ class Grid:
         da_i : xarray.DataArray
             The cumulatively integrated data
         """
-        # first check if axis is present in da
+
         weight = self.get_metric(da, axis)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1400,11 +1400,18 @@ class Grid:
                         break
                 except KeyError:
                     pass
-        if metric_vars is None:
+        if metric_vars is not None:
+            # final check for metric_vars, where dimensions should match with input array
+            if set(metric_vars.dims).issubset(array_dims):
+                return metric_vars
+            else:
+                raise KeyError(
+                    f"Unable to find any combinations of metrics for array dims {array_dims!r} and axes {axes!r}"
+                )
+        else:
             raise KeyError(
                 f"Unable to find any combinations of metrics for array dims {array_dims!r} and axes {axes!r}"
             )
-        return metric_vars
 
     @docstrings.dedent
     def interp_like(self, array, like, boundary=None, fill_value=None):
@@ -1777,7 +1784,7 @@ class Grid:
         return diff / dx
 
     @docstrings.dedent
-    def integrate(self, da, axis, **kwargs):
+    def integrate(self, da, axis, boundary=None, fill_value=None, **kwargs):
         """
         Perform finite volume integration along specified axis or axes,
         accounting for grid metrics. (e.g. cell length, area, volume)
@@ -1795,7 +1802,7 @@ class Grid:
             The integrated data
         """
 
-        weight = self.get_metric(da, axis)
+        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
@@ -1822,15 +1829,17 @@ class Grid:
         da_i : xarray.DataArray
             The cumulatively integrated data
         """
+        boundary = kwargs.get("boundary")
+        fill_value = kwargs.get("fill_value")
 
-        weight = self.get_metric(da, axis)
+        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
         return self.cumsum(weighted, axis, **kwargs)
 
     @docstrings.dedent
-    def average(self, da, axis, **kwargs):
+    def average(self, da, axis, boundary=None, fill_value=None, **kwargs):
         """
         Perform weighted mean reduction along specified axis or axes,
         accounting for grid metrics. (e.g. cell length, area, volume)
@@ -1847,7 +1856,7 @@ class Grid:
         da_i : xarray.DataArray
             The averaged data
         """
-        weight = self.get_metric(da, axis)
+        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
         weighted = da.weighted(weight)
 
         # get dimension(s) corresponding
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1317,7 +1317,7 @@ class Grid:
                 )
         return dim
 
-    def get_metric(self, array, axes, boundary=None, fill_value=None):
+    def get_metric(self, array, axes):
         """
         Find the metric variable associated with a set of axes for a particular
         array.
@@ -1369,9 +1369,8 @@ class Grid:
                     break
             if metric_vars is None:
                 # Condition 2: interpolate metric with matching axis to desired dimensions
-                metric_vars = self.interp_like(mv, array, boundary, fill_value)
+                metric_vars = self.interp_like(mv, array, "extend", None)
         else:
-            # Condition 3: use provided metrics to calculate for required metric
             for axis_combinations in iterate_axis_combinations(axes):
                 try:
                     # will raise KeyError if the axis combination is not in metrics
@@ -1385,13 +1384,13 @@ class Grid:
                             [d for mv in possible_combinations for d in mv.dims]
                         )
                         if metric_dims.issubset(array_dims):
-                            # we found a set of metrics with dimensions compatible with the array
+                            # Condition 3: use provided metrics with matching dimensions to calculate for required metric
                             metric_vars = possible_combinations
                             break
                         else:
                             # Condition 4: metrics in the wrong position (must interpolate before multiplying)
                             metric_vars = tuple(
-                                self.interp_like(pc, array, boundary, fill_value)
+                                self.interp_like(pc, array, "extend", None)
                                 for pc in possible_combinations
                             )
                     if metric_vars is not None:
@@ -1401,7 +1400,7 @@ class Grid:
                 except KeyError:
                     pass
         if metric_vars is not None:
-            # final check for metric_vars, where dimensions should match with input array
+            # final check for metric_vars, where dimensions should match with input array.dims
             if set(metric_vars.dims).issubset(array_dims):
                 return metric_vars
             else:
@@ -1784,7 +1783,7 @@ class Grid:
         return diff / dx
 
     @docstrings.dedent
-    def integrate(self, da, axis, boundary=None, fill_value=None, **kwargs):
+    def integrate(self, da, axis, **kwargs):
         """
         Perform finite volume integration along specified axis or axes,
         accounting for grid metrics. (e.g. cell length, area, volume)
@@ -1802,7 +1801,7 @@ class Grid:
             The integrated data
         """
 
-        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
+        weight = self.get_metric(da, axis)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
@@ -1829,17 +1828,15 @@ class Grid:
         da_i : xarray.DataArray
             The cumulatively integrated data
         """
-        boundary = kwargs.get("boundary")
-        fill_value = kwargs.get("fill_value")
 
-        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
+        weight = self.get_metric(da, axis)
         weighted = da * weight
         # TODO: We should integrate xarray.weighted once available.
 
         return self.cumsum(weighted, axis, **kwargs)
 
     @docstrings.dedent
-    def average(self, da, axis, boundary=None, fill_value=None, **kwargs):
+    def average(self, da, axis, **kwargs):
         """
         Perform weighted mean reduction along specified axis or axes,
         accounting for grid metrics. (e.g. cell length, area, volume)
@@ -1856,7 +1853,8 @@ class Grid:
         da_i : xarray.DataArray
             The averaged data
         """
-        weight = self.get_metric(da, axis, boundary=boundary, fill_value=fill_value)
+
+        weight = self.get_metric(da, axis)
         weighted = da.weighted(weight)
 
         # get dimension(s) corresponding
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -274,7 +274,7 @@ class Axis:
         da_i : xarray.DataArray
             The differenced data
         """
-        position_from, dim = self._get_axis_coord(da)
+        position_from, dim = self._get_position_name(da)
         if to is None:
             to = self._default_shifts[position_from]
 
@@ -340,7 +340,7 @@ class Axis:
         boundary conditions.
         """
 
-        position, this_dim = self._get_axis_coord(da)
+        position, this_dim = self._get_position_name(da)
         this_axis_num = da.get_axis_num(this_dim)
 
         def face_edge_data(fnum, face_axis, count=1):
@@ -508,7 +508,7 @@ class Axis:
         position_check=True,
     ):
 
-        position_from, dim = self._get_axis_coord(da)
+        position_from, dim = self._get_position_name(da)
         axis_num = da.get_axis_num(dim)
 
         boundary_kwargs = dict(
@@ -664,7 +664,7 @@ class Axis:
             The cumsummed data
         """
 
-        pos, dim = self._get_axis_coord(da)
+        pos, dim = self._get_position_name(da)
 
         if to is None:
             to = self._default_shifts[pos]
@@ -930,7 +930,7 @@ class Axis:
             _check_other_dims(target_data)
             return target, target_dim, target_data
 
-        _, dim = self._get_axis_coord(da)
+        _, dim = self._get_position_name(da)
         if method == "linear":
             target, target_dim, target_data = _parse_target(
                 target, target_dim, dim, target_data
@@ -972,7 +972,7 @@ class Axis:
                 # Rechunk to keep xr.apply_func from complaining.
                 # TODO: This should be made obsolete, when the internals are refactored using numba
                 target_data = target_data.chunk(
-                    {self._get_axis_coord(target_data)[1]: -1}
+                    {self._get_position_name(target_data)[1]: -1}
                 )
 
             out = conservative_interpolation(
@@ -991,7 +991,7 @@ class Axis:
         Take the base coords from da, the data from data_new, and return
         a new DataArray with a coordinate on position_to.
         """
-        position_from, old_dim = self._get_axis_coord(da)
+        position_from, old_dim = self._get_position_name(da)
         try:
             new_dim = self.coords[position_to]
         except KeyError:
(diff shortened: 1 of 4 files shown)


xgcm/grid.py
```diff
@@ -1400,11 +1400,18 @@ def get_metric(self, array, axes, boundary=None, fill_value=None):
                         break
                 except KeyError:
                     pass
-        if metric_vars is None:
+        if metric_vars is not None:
+            # final check for metric_vars, where dimensions should match with input array
+            if set(metric_vars.dims).issubset(array_dims):
```
dcherian, line 1405: does this ever get triggered or are you just making sure while you develop this feature?

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -46,15 +44,30 @@ def test_bgrid(self, funcname, boundary, periodic):
             ["dx_t", "dy_t", "dz_t", "area_t", "volume_t"],
             ["xt", "yt", "zt", ["xt", "yt"], ["xt", "yt", "zt"]],
         ):
-            new = func(ds.tracer, axis, **kwargs)
+            if funcname == "cumint":
+                kwargs = dict(boundary=boundary, fill_value=fill_value)
+                new = func(ds.tracer, axis, **kwargs)
+            else:
+                new = func(ds.tracer, axis, boundary, fill_value)
```
dcherian, line 51: this could be combined to only one line with no if conditions

xgcm/grid.py
jdldeauna: Under test_metric_ops_multi_axis/test_missingdim, tracer is averaged over `xt` and so array.dims are `zt` and `yt`, and then it tries to get the average over `"X"`. As it goes through get_metric, the resulting metric has dimensions `xu` and `yt`, and this happens because it goes to Condition 2. If I don't add this final check, that's the final metric that gets passed on to the test and the KeyError message under pytest fails.

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: I rewrote `cumint` so it expects `kwargs` input, because it passes that on to `cumsum`, which has other `kwargs` such as `keep_coords`. I rewrote `average` and `integrate` to accept `boundary` and `fill_value` so they can pass on the values to `get_metric`. These are set to `None` by default. Both `average` and `integrate` have kwargs as possible  inputs but these seem to be passed on to xarray operations rather than within `grid.py`.

xgcm/grid.py
jdldeauna: To clarify, using get_metric goes to Condition 2 because there are existing X-axis metrics specified. So basically, it finds metrics under the X-axis (`len overlap_metrics > 0`), then doesn't find a metric with a matching dimension (because `array.dims = zt and yt`)so it goes on to Condition 2 (interpolate metric with matching axis to desired dimensions).

xgcm/grid.py
```diff
@@ -1352,11 +1367,11 @@ def get_metric(self, array, axes):
                 if metric_dims.issubset(array_dims):
                     metric_vars = mv
                     break
-            # if metric_vars is None:
-            #     # Condition 4: interpolate metric with matching axis to desired dimensions
-            #     metric_vars = self.interp_like(mv, array)
+            if metric_vars is None:
+                # Condition 2: interpolate metric with matching axis to desired dimensions
+                metric_vars = self.interp_like(mv, array, boundary, fill_value)
```
jbusecke, line 1372: So what happens when the user does not set boundary or fill_value? Shoud we set 'reasonable defaults'?

xgcm/grid.py
```diff
@@ -1373,23 +1388,30 @@ def get_metric(self, array, axes):
                             # we found a set of metrics with dimensions compatible with the array
                             metric_vars = possible_combinations
```
jbusecke, line 1389: So you prefer to combine the metric before interpolating it? Should the user have an option to switch this order?

xgcm/grid.py
jbusecke: I am not sure I totally understand what is going on here. Do you have a short code snippet that illustrates the problem (e.g. a notebook cell I can paste)?

xgcm/grid.py
```diff
@@ -1687,7 +1709,7 @@ def _apply_vector_function(self, function, vector, **kwargs):
             )
         for axis_name, component in vector.items():
             axis = self.axes[axis_name]
-            position, coord = axis._get_axis_coord(component)
+            position, coord = axis._get_position_name(component)
```
jbusecke, line 1712: thank you for renaming! much clearer.

xgcm/grid.py
```diff
@@ -1762,7 +1784,7 @@ def derivative(self, da, axis, **kwargs):
         return diff / dx
 
     @docstrings.dedent
-    def integrate(self, da, axis, **kwargs):
+    def integrate(self, da, axis, boundary=None, fill_value=None, **kwargs):
```
jbusecke, line 1787: Just reminding myself here: Is this new option tested?

xgcm/grid.py
```diff
@@ -1807,15 +1829,17 @@ def cumint(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The cumulatively integrated data
         """
+        boundary = kwargs.get("boundary")
```
jbusecke, line 1832: This is inconsistent with the way you parse boundary and fill_value in `integrate`. They should all have the same functional form - either get them from kwargs or define them as explicit keyword arguments. Any thoughts which is preferrable @dcherian @andersy005 ?

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -5,14 +5,18 @@
 from xgcm.test.datasets import datasets_grid_metric
 
 
-def _expected_result(da, metric, grid, dim, axes, funcname, boundary=None):
+def _expected_result(
+    da, metric, grid, dim, axes, funcname, boundary=None, fill_value=None
```
jbusecke, line 9: This does not actually test any of the boundary conditions? I would parametrize over some options (at least for `boundary`).

xgcm/test/test_metric_ops_multi_axis.py
jbusecke: See my comment above: I believe boundary and fill_value should be explicit kwargs in all of those functions. Then you can also simplify this here.

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -46,15 +44,30 @@ def test_bgrid(self, funcname, boundary, periodic):
...
             assert_allclose(new, expected)
 
             # test with tuple input if list is provided
             if isinstance(axis, list):
-                new = func(ds.tracer, tuple(axis), **kwargs)
+                if funcname == "cumint":
+                    kwargs = dict(boundary=boundary, fill_value=fill_value)
+                    new = func(ds.tracer, tuple(axis), **kwargs)
+                else:
+                    new = func(ds.tracer, tuple(axis), boundary, fill_value)
```
jbusecke, line 70: Similar as above

jbusecke (changes requested):
This is a great piece of work! Very excited for this to be merged. I think there are some design things to iron out and there is one piece of logic that warrants a more extensive discussion:

Should we favor combination of metrics over interpolation. Ill post an example in a minute

jbusecke:
Ok so I was actually curious myself whether interpolating or combining lower order metrics is 'better'. So I made this little [notebook](https://nbviewer.jupyter.org/gist/jbusecke/f8bad77e7957ab011d4c83fac37dc8e2), where I check how close we are getting when reconstructing the area on the t cell using various methods. This is only one model, and really just a quick look, but the take home message is here: 
image
This shows the global distribution of relative error for a few ways to combine the metric. It turns out that combining from existing dx/dy on the tracer point performs the best (which is your current first choice in the code). Interpolation is close and first interpolating distances from the u/v points and then combining leads to larger errors (maybe not surprising). The good take home is that all these seem to have a relative error well below 1%!!!

xgcm/grid.py
jdldeauna: Line 1320: `boundary=None, fill_value=None`

dcherian: > boundary or fill_value? Shoud we set 'reasonable defaults'?

I prefer making the user specify defaults in the Grid constructor. We already do this with boundary.

I think xgcm should encourage the pattern of "setup your grid object for all possible cases, then do analysis".

xgcm/grid.py
dcherian: I always prefer explicit keyword arguments. Nice for autocomplete.

dcherian:
Another high-level question is whether we should make the user opt-in to this metric interpolation. For e.g. `grid.autoguess_metrics = True`. If `False` then we could raise an error instead of interpolating.

xgcm/grid.py
jdldeauna:
```python
ds, coords, metrics = datasets_grid_metric("C)
grid = Grid(ds, coords=coords, metrics=metrics, periodic=periodic)
da = ds.tracer.mean("X)
axis = ("X")
weight = grid.get_metric(da, axis) # use current version of grid_metric with interp_like included
```

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: `kwargs` in `cumint` is passed on to `cumsum` which can accept boundary and fill_value
`kwargs` in `average` and `integrate` are passed on to `xarray.mean` and `xarray.sum`
if I just include boundary and fill_value as part of `kwargs` for average and integrate I get the following error statements:
```python
TypeError: nansum() got an unexpected keyword argument 'boundary'
TypeError: mean() got an unexpected keyword argument 'boundary'
```

jdldeauna: See lines 1813 and 1865

xgcm/grid.py
jdldeauna: Suggestion from Deepak: https://github.com/xgcm/xgcm/pull/343#discussion_r648695848

xgcm/grid.py
jdldeauna: `kwargs` in `cumint` is passed on to `cumsum` which can accept boundary and fill_value
`kwargs` in `average` and `integrate` are passed on to `xarray.mean` and `xarray.sum`
if I just include boundary and fill_value as part of `kwargs` for average and integrate I get the following error statements:
```python
TypeError: nansum() got an unexpected keyword argument 'boundary'
TypeError: mean() got an unexpected keyword argument 'boundary'
```
See lines 1813 and 1865

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: `kwargs` in `cumint` is passed on to `cumsum` which can accept boundary and fill_value
`kwargs` in `average` and `integrate` are passed on to `xarray.mean` and `xarray.sum`
if I just include boundary and fill_value as part of `kwargs` for average and integrate I get the following error statements:
```python
TypeError: nansum() got an unexpected keyword argument 'boundary'
TypeError: mean() got an unexpected keyword argument 'boundary'
```
See lines 1813 and 1865

xgcm/test/test_metric_ops_multi_axis.py
dcherian: Yes instead of `kwargs` lets add `boundary` and `fill_value` to the signature of `average` and `integrate`. That way it gets intercepted, is not passed on to xarray, and users know that they can specify it using autocomplete.

xgcm/grid.py
jdldeauna: So because boundary and fill_value are already specified with Grid:
```python
class Grid:
    """
    An object with multiple :class:`xgcm.Axis` objects representing different
    independent axes.
    """

    def __init__(
        self,
        ds,
        check_dims=True,
        periodic=True,
        default_shifts={},
        face_connections=None,
        coords=None,
        metrics=None,
        boundary=None,
        fill_value=None,
    ):
```

Those values will pass on to `get_metric` and I can just use:
```python
def get_metric(self, array, axes, boundary, fill_value):
```
instead of:
```python
def get_metric(self, array, axes, boundary=None, fill_value=None):
```

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: because the function `_expected_result` is used inside other functions:
```python
def test_bgrid(self, funcname, boundary, periodic, fill_value):
        grid = Grid(ds, coords=coords, metrics=metrics, periodic=periodic)

        func = getattr(grid, funcname)
        kwargs = dict(boundary=boundary, fill_value=fill_value)

        # test tracer position
        for axis, metric_name, dim in zip(
            ["X", "Y", "Z", ["X", "Y"], ["X", "Y", "Z"]],
            ["dx_t", "dy_t", "dz_t", "area_t", "volume_t"],
            ["xt", "yt", "zt", ["xt", "yt"], ["xt", "yt", "zt"]],
        ):
            new = func(ds.tracer, axis, **kwargs)
            expected = _expected_result(
                ds.tracer,
                ds[metric_name],
                grid,
                dim,
                axis,
                funcname,
                boundary,
                fill_value,
            )
            assert_allclose(new, expected)
```
the `boundary` value will differ depending on the parameters used per iteration, and `boundary` is parameterized fro `test_bgrid`.

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: > Yes instead of `kwargs` lets add `boundary` and `fill_value` to the signature of `average` and `integrate`. 

Is this what you mean?

```python
def average(self, da, axis, boundary, fill_value, **kwargs):
        """
        Perform weighted mean reduction along specified axis or axes,
        accounting for grid metrics. (e.g. cell length, area, volume)

        Parameters
        ----------
        axis : str, list of str
            Name of the axis on which to act
        boundary : str or dict, optional,
            boundary can either be one of {None, 'fill', 'extend', 'extrapolate'}
            * None:  Do not apply any boundary conditions. Raise an error if
              boundary conditions are required for the operation.
            * 'fill':  Set values outside the array boundary to fill_value
              (i.e. a Dirichlet boundary condition.)
            * 'extend': Set values outside the array to the nearest array
              value. (i.e. a limited form of Neumann boundary condition where
              the difference at the boundary will be zero.)
            * 'extrapolate': Set values by extrapolating linearly from the two
              points nearest to the edge
            This sets the default value. It can be overriden by specifying the
            boundary kwarg when calling specific methods.
        fill_value : float, optional
            The value to use in the boundary condition when `boundary='fill'`.
        **kwargs: dict
            Additional arguments passed to `xarray.DataArray.weighted.mean`

        Returns
        -------
        da_i : xarray.DataArray
            The averaged data
        """
```

jdldeauna: Then cumint will still use kwargs?

```python
def cumint(self, da, axis, **kwargs):
        """
        Perform cumulative integral along specified axis or axes,
        accounting for grid metrics. (e.g. cell length, area, volume)

        Parameters
        ----------
        axis : str, list of str
            Name of the axis on which to act
        %(grid_func.parameters)s

        Returns
        -------
        da_i : xarray.DataArray
            The cumulatively integrated data
        """
```

jdldeauna:
> The good take home is that all these seem to have a relative error well below 1%!!!

Great!

jdldeauna:
> Another high-level question is whether we should make the user opt-in to this metric interpolation. For e.g. `grid.autoguess_metrics = True`. If `False` then we could raise an error instead of interpolating.

Is`autoguess_metrics` a new keyword in `class Grid` or `get_metric`?
If I understand correctly, the error will appear before any metric interpolation steps, so after `Line 1370` and `Line 1392`?

xgcm/grid.py
jdldeauna: The line `weight = grid.get_metric(da,axis)` is from average:

```python
def average(self, da, axis, **kwargs):
        
        weight = self.get_metric(da, axis)
        weighted = da.weighted(weight)

        # get dimension(s) corresponding
        # to `da` and `axis` input
        dim = self._get_dims_from_axis(da, axis)
        return weighted.mean(dim, **kwargs)
```
so in `test_metric_ops_multi_axis/test_missingdim`, we shouldn't be able to do `cumint, average or integrate` for cases where you don't have a metric that matches the exact dimensions of your DataArray `da`.

xgcm/grid.py
dcherian: No you want
``` python
def get_metric(self, array, axes, boundary=None, fill_value=None):	
```
The first version (with no default values) requires the user to specify `boundary` and `fill_value` always even if default `boundary` and `fill_value` is set on the `Axis` object.

We want the second version because that allows the user to override at the method-level if necessary but when `boundary is None` (the default) `_get_neighbour_binary_func` (or whatever) will use `ax.boundary`
https://github.com/xgcm/xgcm/blob/f61edc455cc4a1c5d35e7f9ffaf88f408f333f65/xgcm/grid.py#L281-L285

xgcm/test/test_metric_ops_multi_axis.py
dcherian: Yes but with default `boundary=None, fill_value=None`. In this case, it should pull the default from `Axis.boundary` and `Axis.fill_value` automatically ( but this should be tested )

xgcm/grid.py
jdldeauna: Alright!

> I prefer making the user specify defaults in the Grid constructor. We already do this with boundary.
> I think xgcm should encourage the pattern of "setup your grid object for all possible cases, then do analysis".

I do wonder what you meant by the statement above in relation to get_metric.

jbusecke: I suggest to just hardcode `self.interp_like(mv, array, 'extend', None)` here for simplicity. What do you think about that? Would eliminate all the fiddling with boundary and fill_val in the other methods.

xgcm/grid.py
```diff
@@ -1352,11 +1367,13 @@ def get_metric(self, array, axes):
                 if metric_dims.issubset(array_dims):
                     metric_vars = mv
                     break
-            # if metric_vars is None:
-            #     # Condition 4: interpolate metric with matching axis to desired dimensions
-            #     metric_vars = self.interp_like(mv, array)
+            if metric_vars is None:
+                # Condition 2: interpolate metric with matching axis to desired dimensions
+                warnings.warn(
+                    "Metric outputs are interpolated from initial metrics input"
```
dcherian, line 1373:
```suggestion
                    f"Metric at {array.dims} being interpolated from metrics at {mv.dims}"
```

xgcm/grid.py
```diff
@@ -1354,6 +1345,10 @@ def get_metric(self, array, axes):
         metric_vars = None
         array_dims = set(array.dims)
 
+        # Will raise a Value Error if array doesn't have a dimension corresponding to metric axes specified
+        # See _get_dims_from_axis
+        self._get_dims_from_axis(array, frozenset(axes))
```
jbusecke, line 1350: Nice!

xgcm/grid.py
```diff
@@ -1834,7 +1824,7 @@ def cumint(self, da, axis, **kwargs):
         da_i : xarray.DataArray
             The cumulatively integrated data
         """
-
+        # first check if axis is present in da
```
jbusecke, line 1827: This seems a bit confusing over here?

xgcm/test/test_metrics.py
```diff
@@ -179,6 +179,16 @@ def test_get_metric_orig(
         assert metric.equals(expected)
 
 
+def test_get_metric_with_no_matching_axis():
+    # the metric axis should be present in array
+    ds, coords, metrics = datasets_grid_metric("C")
+    grid = Grid(ds, coords=coords, metrics=metrics)
+
+    match_message = "Did not find single matching dimension"
```
jbusecke, line 187: Isn't this the same logic as in `test_metric_axes_missing_from_array`?

xgcm/test/test_metrics.py
```diff
@@ -190,18 +200,6 @@ def test_get_metric_with_conditions_01():
     xr.testing.assert_allclose(get_metric, expected_metric)
 
 
-# @pytest.mark.parametrize("periodic", [True, False])
```
jbusecke: Should we leave the periodic parametrization in here?

xgcm/grid.py
jdldeauna: Yeah, I forget to delete that, it was supposed to just remind me to do that step 😅

xgcm/test/test_metrics.py
jdldeauna: Yep, I wrote this initially to test including the line `self._get_dims_from_axis(array, frozenset(axes))` in get_metric. I'll take it out, because `test_metric_axes_missing_from_array` does the same thing.

xgcm/test/test_metrics.py
jdldeauna: Hmm wouldn't a `False` condition require setting boundary values?

jbusecke: Yeah you could just set `boundary='extend`, which will be ignored when periodic is true. Just an idea

xgcm/grid.py
```diff
@@ -1306,14 +1306,20 @@ def set_metrics(self, key, value, overwrite=False):
     def _get_dims_from_axis(self, da, axis):
         dim = []
         for ax in axis:
-            all_dim = self.axes[ax].coords.values()
-            matching_dim = [di for di in all_dim if di in da.dims]
-            if len(matching_dim) == 1:
-                dim.append(matching_dim[0])
+            set_ax = frozenset(_maybe_promote_str_to_list(ax))
```
dcherian, line 1309: I think the promotion should be outside the loop since you're assuming that `axis` is `("X",)` and not `"X"` for example

so 
``` python
axis = _maybe_promote_str_to_list(ax)
for ax in axis:
    if ax in self.axes:
        ...
```

you don't always have to use set operations to check for membership.

xgcm/grid.py
```diff
@@ -1306,14 +1306,20 @@ def set_metrics(self, key, value, overwrite=False):
...
+            if set_ax.issubset(set(self.axes.keys())):
+                all_dim = self.axes[ax].coords.values()
+                matching_dim = [di for di in all_dim if di in da.dims]
+                if len(matching_dim) == 1:
+                    dim.append(matching_dim[0])
+                else:
+                    raise ValueError(
+                        "Did not find single matching dimension %s from %s corresponding to axis %s. Got (%s)"
+                        % (da.dims, da.name, ax, matching_dim)
```
dcherian, line 1318: Lets f-string these. They're easier to read.

Feel free to ignore this suggestion

xgcm/grid.py
```diff
@@ -1370,15 +1382,21 @@ def get_metric(self, array, axes):
...
                             break
-                        # else:
-                        #     # Condition 5: metrics in the wrong position (must interpolate before multiplying)
-                        #     metric_vars = tuple(
-                        #         self.interp_like(pc, array)
-                        #         for pc in possible_combinations
-                        #     )
+                        else:
+                            # Condition 4: metrics in the wrong position (must interpolate before multiplying)
+                            possible_combinations_dims = [
```
dcherian, line 1390: really minor comments: can we use `possible_dims` instead so it's a single line?

xgcm/grid.py
dcherian: Wait I don't see what the improvement is over the old version.

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -179,41 +173,35 @@ def test_missingaxis(self, axis, funcname, periodic, boundary):
...
+            kwargs = dict(boundary="fill")
         else:
             kwargs = dict()
 
-        match_message = "Unable to find any combinations of metrics for array dims.*X.*"
-        with pytest.raises(KeyError, match=match_message):
+        func = getattr(grid, funcname)
+
+        match_message = "Did not find single matching dimension"
+        with pytest.raises(ValueError, match=match_message):
```
dcherian, line 202:
```suggestion
        with pytest.raises(ValueError, match="Did not find single matching dimension"):
```

xgcm/test/test_metric_ops_multi_axis.py
```diff
@@ -179,41 +173,35 @@ def test_missingaxis(self, axis, funcname, periodic, boundary):
...
             func(ds.tracer.mean("xt"), "X", **kwargs)
 
-        match_message = (
-            "Unable to find any combinations of metrics for array dims.*X.*Y.*Z.*"
-        )
-        with pytest.raises(KeyError, match=match_message):
+        match_message = "Did not find single matching dimension"
+        with pytest.raises(ValueError, match=match_message):
```
dcherian, line 206:
```suggestion
        with pytest.raises(ValueError, match="Did not find single matching dimension"):
```

dcherian (approved):
v. minor comments. Nice work Dianne!

xgcm/test/test_metric_ops_multi_axis.py
dcherian: This post is relevant: http://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

xgcm/grid.py
jdldeauna: In the old version, if `ax` wasn't present in `self.axes`:
```python
for ax in axis:
    all_dim = self.axes[ax].coords.values()
```
This line would just produce an error: `KeyError: []`
I included an if statement:
```python
set_ax = frozenset(_maybe_promote_str_to_list(ax))
if set_ax.issubset(set(self.axes.keys())):
```
so that the error would have more explanation: `KeyError("Did not find axis %s from data array %s" % (ax, da.name))`

xgcm/test/test_metric_ops_multi_axis.py
jdldeauna: Thanks for this Deepak! I'll do my best to keep it in mind.

xgcm/grid.py
dcherian: ah nice!

Issue #110, opened 2018-07-30
spurious coordinates added when input datasets have dimensions without coordinates 


rabernat:
In https://github.com/xgcm/xgcm/issues/108#issuecomment-406433489, @vrx- suggested that xgcm might not be properly handling xarray dimensions without coordinates. This is plausible, since, when I started working on xgcm, coordinates were required for all dimensions.

I couldn't actually find an example where this causes xgcm to break. However, some unexpected things do happen. 

Create a dataset with no coords:
```python
>>> ds = xr.Dataset({'f_c': ('x_c', np.ones(10)),'f_g': ('x_g', np.ones(11))})
>>> ds
<xarray.Dataset>
Dimensions:  (x_c: 10, x_g: 11)
Dimensions without coordinates: x_c, x_g
Data variables:
    f_c      (x_c) float64 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0
    f_g      (x_g) float64 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0
```

Create a grid object from it:
```python
>>> grid = Grid(ds, coords={'X': {'center': 'x_c', 'outer': 'x_g'}})
>>> grid
<xgcm.Grid>
X Axis (periodic):
  * center   x_c (10) --> outer
  * outer    x_g (11) --> center
```

Use it for operations:
```python
>>> grid.diff(ds.f_g, 'X')
<xarray.DataArray (x_c: 10)>
array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])
Coordinates:
  * x_c      (x_c) int64 0 1 2 3 4 5 6 7 8 9
>>> grid.interp(ds.f_c, 'X', boundary='fill')
<xarray.DataArray (x_g: 11)>
array([ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.])
Coordinates:
  * x_g      (x_g) int64 0 1 2 3 4 5 6 7 8 9 10
```

**The outputs have coordinates, although the parent dataset has none!** This is becase `ds['x_c']` still returns a valid data array, which is used to create the coordinates of the output DataArray.

Instead, we probably want the outputs to have no coordinates either.

PR #358, opened 2021-07-20
Update constructor docstring and input checks

### LINKED ISSUE #208: Check if values passed in `coords={..}` are valid dimensions.
Currently xgcm does not check if the values passed in `Grid(..., coords={...})` are valid dimensions upon the creation of the grid object. 

I recently came across a situation like this:
```python
import xarray as xr
import numpy as np
data = np.random.rand(4,5)
coord = np.random.rand(4,5)
ds = xr.DataArray(data, dims=['x','y'], coords={'c':(['x', 'y'], coord)}).to_dataset(name='data')
```

Should it be possible to assign a coordinate instead of a dimension like this:
```python
from xgcm import Grid
grid = Grid(ds, coords={'X':{'center':'c'}})
```
this currently does not raise an error/warning.

Only when trying to compute something we get:
```python
grid.diff(ds.data,'X')
```

```
...
KeyError: "None of the DataArray's dims ('x', 'y') were found in axis coords."
```

I think it might be easier for new users to get a warning like ` c is not a dataset dimension. xgcm can only accept 1D dimensions in coords` or similar. Should be very cheap test too, so not weigh the grid object creation down too much.

cc @abigalew

jbusecke:
- [x] Closes #208
 - [x] Tests added

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1087,7 +1087,12 @@ class Grid:
             axes in ``ds``. If the values are not present in ``ds`` or are not dimensions,
             an error will be raised.
         metrics : dict, optional
-            Specification of grid metrics
+            Specification of grid metrics mapping axis names (X, Y, Z) to corresponding
+            metric variable names in the dataset 
+            (e.g. {('X',):['dx_t'], ('X', 'Y'):['area_tracer', 'area_u']}
+            for the cell distance in the x-direction ``dx_t`` and the 
+            horizontal cell areas ``area_tracer`` and ``area_u``, located at 
+            different grid positions).
         boundary : {None, 'fill', 'extend', 'extrapolate', dict}, optional
             A flag indicating how to handle boundaries:
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1082,9 +1082,9 @@ class Grid:
         coords : dict, optional
             Specifies positions of dimension names along axes X, Y, Z, e.g
             ``{'X': {'center': 'XC', 'left: 'XG'}}``.
-            Each key should be an axis name (e.g., `X`, `Y`, or `Z`) and map 
+            Each key should be an axis name (e.g., `X`, `Y`, or `Z`) and map
             to a dictionary which maps positions (`center`, `left`, `right`,
-            `outer`, `inner`) to dimension names in the dataset 
+            `outer`, `inner`) to dimension names in the dataset
             (in the example above, `XC` is at the `center` position and `XG`
             at the `left` position along the `X` axis).
             If the values are not present in ``ds`` or are not dimensions,
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1080,11 +1080,12 @@ class Grid:
         face_connections : dict
             Grid topology
         coords : dict, optional
-            Explicit specification of axis coordinates, e.g
+            Explicit specification of axis dimensions, e.g
             ``{'X': {'center': 'XC', 'left: 'XG'}}``.
             Each key should be the name of an axis. The value should be
             a dictionary mapping positions (e.g. ``'left'``) to names of
-            coordinates in ``ds``.
+            axes in ``ds``. If the values are not present in ``ds`` or are not dimensions,
+            an error will be raised.
         metrics : dict, optional
             Specification of grid metrics
         boundary : {None, 'fill', 'extend', 'extrapolate', dict}, optional
@@ -1119,6 +1120,18 @@ class Grid:
         else:
             all_axes = comodo.get_all_axes(ds)
             coords = {}
+
+        # check coords input validity
+        for axis, positions in coords.items():
+            for pos, dim in positions.items():
+                if dim not in ds.variables:
+                    raise ValueError(
+                        f"Could not find dimension `{dim}` (for the `{pos}` position on axis `{axis}`) in input dataset."
+                    )
+                if dim not in ds.dims:
+                    raise ValueError(
+                        f"Input `{dim}` (for the `{pos}` position on axis `{axis}`) is not a dimension in the input datasets `ds`."
+                    )
 
         self.axes = OrderedDict()
         for axis_name in all_axes:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1082,9 +1082,11 @@ class Grid:
         coords : dict, optional
             Specifies positions of dimension names along axes X, Y, Z, e.g
             ``{'X': {'center': 'XC', 'left: 'XG'}}``.
-            Keys should be one of `X`, `Y`, or `Z` and map to a dictionary
-            mapping positions (`center`, `left`, `right`, `outer`, `inner`)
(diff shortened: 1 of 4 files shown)


jbusecke:
Ok I think I fixed all tests except [`test_grid_no_coords`](https://github.com/jbusecke/xgcm/blob/2bb9324e65a8a898490683bc92c0160b4d22a1ec/xgcm/test/test_grid.py#L625-L636). I am not sure we need this test anymore? @rabernat could you confirm this?

rabernat:
> Ok I think I fixed all tests except [`test_grid_no_coords`](https://github.com/jbusecke/xgcm/blob/2bb9324e65a8a898490683bc92c0160b4d22a1ec/xgcm/test/test_grid.py#L625-L636). I am not sure we need this test anymore?

This test is a regression test for #110. The point of the test is to ensure that you can use xgcm with Xarray datasets that don't have dimension coordinates. The failure of the test indicates that this PR would reintroduce the bug covered by the test. So it needs to be fixed.

xgcm/grid.py
```diff
@@ -1120,6 +1121,18 @@ def __init__(
             all_axes = comodo.get_all_axes(ds)
             coords = {}
 
+        # check coords input validity
+        for axis, positions in coords.items():
+            for pos, dim in positions.items():
+                if dim not in ds.variables:
+                    raise ValueError(
+                        f"Could not find dimension `{dim}` (for the `{pos}` position on axis `{axis}`) in input dataset."
+                    )
```
rabernat, line 1130: This is the problem. You are requiring the dim to be in `ds.variables`. But that's not the case for datasets with no coordinate variable, e.g.
```python
ds = xr.DataArray([0], dims='x', name='foo').to_dataset()
'x' in ds.variables  # --> False
```

jbusecke:
Thanks for the explanation @rabernat, but I am still having trouble understanding the test .
I am currently debugging it to get a better overview. 

So the full test is:
```python
def test_grid_no_coords(periodic_1d):
    """Ensure that you can use xgcm with Xarray datasets that don't have dimension coordinates."""
    ds, periodic, expected = periodic_1d
    ds_nocoords = ds.drop_dims(list(ds.dims.keys()))

    coords = expected["axes"]
    grid = Grid(ds_nocoords, periodic=periodic, coords=coords)

    diff = grid.diff(ds["data_c"], "X")
    assert len(diff.coords) == 0
    interp = grid.interp(ds["data_c"], "X")
    assert len(interp.coords) == 0
```

When I go through it line by line, the dataset `ds_nocoord` used to create the grid does have no dimensions
```python
print(ds_nocoords)
```
```
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*
```
Is that really what is intended here? In my opinion, this should fail!

From my understanding of #110, what we rather want is something like this:
```
<xarray.Dataset>
Dimensions:  (XC: 100, XG: 100)
Dimensions without coordinates: XC, XG
Data variables:
    data_g   (XG) float64 0.5782 0.8786 0.01355 0.07767 ... 0.2189 0.6506 0.834
    data_c   (XC) float64 0.08648 0.664 0.4159 0.1341 ... 0.8866 0.8972 0.5521
```
no?

I got this output by replacing the line `ds_nocoords = ds.drop_dims(list(ds.dims.keys()))` with `ds_nocoords = ds.drop_vars(list(ds.dims.keys()))`

This does pass with some slight modifications to the input check as suggested above. 

Do you think I should change the test or am I still missing something here?

rabernat:
Yikes! Yes the test itself was broken!

xgcm/grid.py
```diff
@@ -1080,11 +1080,12 @@ def __init__(
         face_connections : dict
             Grid topology
         coords : dict, optional
-            Explicit specification of axis coordinates, e.g
+            Explicit specification of axis dimensions, e.g
```
jbusecke, line 1083: @rabernat @dcherian I tried to make it clearer here that the input should be dimension names. I guess in retrospect `coords` is a bit of a confusing variable name, but changing that would be a major undertaking. So I figured this would be a quick fix?

dcherian:
```suggestion
            Explicit specification dimension positions along axis, e.g
```
?

xgcm/grid.py
```diff
@@ -1080,11 +1080,12 @@ def __init__(
...
             ``{'X': {'center': 'XC', 'left: 'XG'}}``.
             Each key should be the name of an axis. The value should be
```
dcherian, line 1085: Add something like this? XC, XG are only understandable to MITgcm-adajacent people.

```
Keys should be one of `X`, `Y`, or `Z` and map to a dictionary mapping positions (`center`, `left`, `right`, `outer`, `inner`) to dimension names in the dataset (here, `XC` and `XG`).
```

xgcm/grid.py
```diff
@@ -1080,11 +1080,12 @@ def __init__(
...
             a dictionary mapping positions (e.g. ``'left'``) to names of
-            coordinates in ``ds``.
+            axes in ``ds``. If the values are not present in ``ds`` or are not dimensions,
+            an error will be raised.
         metrics : dict, optional
             Specification of grid metrics
```
dcherian, line 1090: `Specification of grid metrics mapping axis names (`X`, `Y`, `Z`) to corresponding metric variable names in the dataset`

xgcm/grid.py
jbusecke: How about `Explicit specification of positions along axis, indicated by a dataset dimension, e.g` to make it even more clear?

xgcm/grid.py
jbusecke: That is great!

xgcm/grid.py
dcherian: `Specifies positions of dimension names along axes `X`, `Y`, `Z``?

jbusecke: Noice.

dcherian: I think @jdldeauna's input would be valuable here.

xgcm/grid.py
```diff
@@ -1080,13 +1080,20 @@ def __init__(
         face_connections : dict
             Grid topology
         coords : dict, optional
-            Explicit specification of axis coordinates, e.g
+            Specifies positions of dimension names along axes X, Y, Z, e.g
```
jbusecke, line 1083: @jdldeauna I would love to get some feedback on these new docstrings. Do you think these are sufficiently clear? Or do you have any suggestions how to improve these?

jdldeauna: Looks great! I just had a minor suggestion:
```python
Keys should be one of `X`, `Y`, or `Z` and map to a dictionary 
            mapping positions 
```
maybe
```python
Each key should be an axis name (e.g., `X`, `Y`, or `Z`) and map to a dictionary which maps positions 
```

jbusecke: Thanks a lot, that is even more clear. I have integrated it and also added a few more details in the following sentence. What do you think @jdldeauna?

dcherian (approved):

PR #361, opened 2021-07-26
Drop Support for python 3.6

### LINKED ISSUE #360: CI for python 3.6  broken [BUG]
**What happened**:
The CI for [python 3.6](https://github.com/xgcm/xgcm/runs/3161900333?check_suite_focus=true) is broken. I suggest we drop support for these versions, which are not supported by xarray anymore since [0.17.0](http://xarray.pydata.org/en/stable/whats-new.html?highlight=3.6#id53)

jbusecke:
- [x] Closes #360 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,6 @@ setup(
         "Topic :: Scientific/Engineering",
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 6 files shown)


jbusecke:
Is there anything else I need to do to drop the version support?

dcherian (approved):
I think this is all that's neede

andersy005 (approved):
You could also update the minimum required version to `v0.17` (which is the version that dropped python 3.6 in xarray). 



https://github.com/xgcm/xgcm/blob/6e68f71caad8f57e2776d4dc7b4aaa3eefb1875f/setup.py#L10

setup.py
```diff
@@ -7,7 +7,7 @@
     long_description = f.read()
 
 install_requires = [
-    "xarray>=0.14.1",
+    "xarray>=0.17.0",
     "dask",
     "numpy",
     "future",
```
andersy005, line 13: Is this still needed? I thought it was only needed for applications that wanted to support Python 2 & Python 3

jbusecke: I think you are right. I just removed it. Lets see if that upsets the CI.

Issue #344, opened 2021-06-12
Proposal: grid_ufunc


rabernat:
Xarray and the whole python array ecosystem has evolved considerably since the creation of xgcm. The introduction of xarray's `apply_ufunc`, dask / numba gufuncs, etc. has evolved the way we think about array code, allowing us to do complex array operations in a compact, elegant way. At the same time, we have become more aware of some of the performance limitations of xgcm, particularly with large dask arrays. This proposal is for an internal refactoring of xgcm that could address all of the above.

### Background: generatlized ufuncs

The numpy C api defines the concept of [generalized ufuncs](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html) (gufuncs). Dask [exposes gufuncs at the python level](https://docs.dask.org/en/latest/array-gufunc.html). Numba also allows you to write [vectorized gufuncs](https://numba.pydata.org/numba-doc/latest/user/vectorize.html#the-guvectorize-decorator).  At the core of gufuncs is the concept of a _signature_. A signature is a symbolic representation of how a function acts on a set of _core dimensions_. The specifies the required core dimensions of the inputs and how the output dimensions are related to the input dimensions. It's easiest to explain signatures through examples. Here I copy directly from the numpy docs linked above.

name | signature | common usage
-- | -- | --
add | `(),()->()` | binary ufunc
sum1d | `(i)->()` | reduction
inner1d | `(i),(i)->()` | vector-vector multiplication
matmat | `(m,n),(n,p)->(m,p)` | matrix multiplication
vecmat | `(n),(n,p)->(p)` | vector-matrix multiplication
matvec | `(m,n),(n)->(m)` | matrix-vector multiplication
matmul | `(m?,n),(n,p?)->(m?,p?)` | combination of the four above
outer_inner | `(i,t),(j,t)->(i,j)` | inner over the last dimension, outer over the second to last, and loop/broadcast over the rest.
cross1d | `(3),(3)->(3)` | cross product where the last dimension is frozen and must be 3

This syntax is used by numpy, dask, numba, and likely other parts of the pydata ecosystem do describe gufuncs.

### New idea: grid_ufunc

I think we should refactor all of xgcm's primitive operations to be gufuncs. However, we need an additional layer of complexity, because xgcm deals not just with raw dimensions but with _axes_. We could imagine generalizing the gufunc signature as follows.

name | signature
-- | -- 
diff_center_to_left(a) | `(X:center)->(X:left)`
[...]

PR #362, opened 2021-07-27
as_grid_ufunc decorator

TomNicholas:
First step towards closing #344 - so far only does point (1) of @dcherian's [comment](https://github.com/xgcm/xgcm/issues/344#issuecomment-860760874), but the regex for the signature parser was tricky enough that I wanted to put this up as a WIP now that it passes my signature tests.

The regex parser is an adapted version of the one used by [numpy](https://github.com/numpy/numpy/blob/main/numpy/lib/function_base.py#L1846) and [dask](https://github.com/dask/dask/blob/8d49c3f1491b4aaccb6560814718f498abf63335/dask/array/gufunc.py#L18), and I'm going to copy dask's class structure for actually applying it to functions.

One question is how is the parser supposed to know what the `output_sizes` kwarg to `xr.apply_ufunc` should be? That information isn't in the signature specification @rabernat [gave](https://github.com/xgcm/xgcm/issues/344#issue-919651677). Or do we just not envisage a use case for this function where it creates new dimensions on the output?

 - [x] Tests added
 - [ ] ~~User visible changes (including notable bug fixes) are documented in `whats-new.rst`~~


dcherian:
> the parser supposed to know what the output_sizes kwarg to xr.apply_ufunc should be? 

You can infer the sizes of the new dimension based on a single dimension already present in `xgcm.Grid` i.e. if you know `center` is size N, the `inner` is size N-1, `outer` is N+1, etc. You might have to make up names for the new dimension though?

TomNicholas:
> You can infer the sizes of the new dimension

Oh yes I guess maybe... Only if you know that X->Y though - what if the input had mutliple dimensions. Like would we ever need to deal with a function with a signature like `"(X:center),(Y:inner)->(Z:outer)"`?

> You might have to make up names for the new dimension though?

Those names will be specified in the signature, e.g. if `signature="(X:center)->(Y:outer)"` we know the output name is `Y`.

jbusecke:
> Oh yes I guess maybe... Only if you know that X->Y though - what if the input had multiple dimensions? Like would we ever need to deal with a function with a signature like "(X:center),(Y:inner)->(Z:outer)"?

I couldnt think of any example that would require this at the moment. But let me sleep on it hehe.

dcherian:
> Those names will be specified in the signature, e.g. if signature="(X:center)->(Y:outer)" we know the output name is Y.

I don't think so. The signature tells us that the output has one core dimension, along the Y axis at the position `outer`. But if `Grid` does not have any variable currently at position `Y: outer` we don't know the dimension name.

rabernat:
> One question is how is the parser supposed to know what the `output_sizes` kwarg to `xr.apply_ufunc` should be?

I don't think it will be possible to fully populate the dimensions / coordinates without reference to the Grid dataset (the dataset used to initialize the grid, stored in `Grid._ds`). For each grid position signature (e.g. `(X:center)`) there will be at most 1 matching dimension (and possibly coordinate) in the Grid dataset.

We need to figure out the relationship between thie parser and the Grid object. But in my mind, the parser cannot be self-contained: it needs the Grid.

xgcm/grid_ufunc.py
```diff
@@ -0,0 +1,126 @@
...
+    ---------
+    signature : string
+        Generalized universal function signature, e.g., ``"(X:center)->(X:left)"``
+        for ``diff_center_to_left(a)`.
+
+    Returns
+    -------
+    Tuple of input and output core dimensions parsed from the signature, each
+    of the form List[Tuple[str, ...]].
+    """
```
rabernat, line 36: I don't see how this function can do its job without access to a Grid object.

TomNicholas:
> I don't think it will be possible to fully populate the dimensions / coordinates without reference to the Grid dataset

Okay I've obviously misunderstood. So the parser that I've written doesn't yet find the core dimensions, instead it really finds the xgcm Axes specified by the signature and the positions of those Axes. Then to get the core dims we need a second step where we consult the grid object via something like `grid.axes['Axis_name_from_parser'].coords['Axis_position_from_parser']`?

rabernat:
Yes, that matches my understanding. The parser you've written is great and does just whit it needs. There is just a final step before xr.apply_ufunc...

> Then to get the core dims we need a second step where we consult the grid object via something like `grid.axes[Axis_name_from_parser].coords[Axis_position_from_parser]`?

exactly.

TomNicholas:
Parsing the Axis and determining the core dims from the grid object seems fine, but moving forward there are potentially some problems with changing output sizes.

In `xarray.apply_ufunc` you can't change the size of core dimensions without "excluding" them:

```python
In [8]: def shorten(a):
   ...:     return a[1:]
   ...: 

In [9]: da = xr.DataArray(np.arange(10), dims='x')

In [10]: xr.apply_ufunc(shorten, da, input_core_dims=[['x']], output_core_dims=[['x']])
<ipython-input-15-268f57ff3530> in <module>
----> 1 xr.apply_ufunc(shorten, da, input_core_dims=[['x']], output_core_dims=[['x']])

~/Documents/Work/Code/xarray/xarray/core/computation.py in apply_ufunc(func, input_core_dims, output_core_dims, exclude_dims, vectorize, join, dataset_join, dataset_fill_value, keep_attrs, kwargs, dask, output_dtypes, output_sizes, meta, dask_gufunc_kwargs, *args)
   1178             join=join,
   1179             exclude_dims=exclude_dims,
-> 1180             keep_attrs=keep_attrs,
   1181         )
   1182     # feed Variables directly through apply_variable_ufunc

~/Documents/Work/Code/xarray/xarray/core/computation.py in apply_dataarray_vfunc(func, signature, join, exclude_dims, keep_attrs, *args)
    291 
    292     data_vars = [getattr(a, "variable", a) for a in args]
--> 293     result_var = func(*data_vars)
    294 
    295     if signature.num_outputs > 1:

~/Documents/Work/Code/xarray/xarray/core/computation.py in apply_variable_ufunc(func, signature, exclude_dims, dask, output_dtypes, vectorize, keep_attrs, dask_gufunc_kwargs, *args)
    779                     "dimensions specified in ``exclude_dims`` with "
    780                     "xarray.apply_ufunc are allowed to change size.".format(
--> 781                         dim, dim_sizes[dim], new_size
    782                     )
    783                 )

ValueError: size of dimension 'x' on inputs was unexpectedly changed by applied function from 10 to 9. Only dimensions specified in ``exclude_dims`` with xarray.apply_ufunc are allowed to change size.
```

Any operation that changes the Axis position from say "center" to "inner" would require changing the length of a core dimension.

You can do this

```python
In [16]: xr.apply_ufunc(shorten, da, input_core_dims=[['x']], output_core_dims=[['x']], exclude_dims={'x'})
Out[16]: 
<xarray.DataArray (x: 9)>
array([1, 2, 3, 4, 5, 6, 7, 8, 9])
Dimensions without coordinates: x
```
but I don't really understand what `exclude_dims` actually does well enough to know if this is a problem? The docs say "exclude from alignment and broadcasting entirely", but what does that mean? In what scenario will that break our use cases? One thing it will definitely do is drop any coordinates along that dimension.

rabernat:
In your example, `input_core_dims` and `output_core_dims` are the same, but in xgcm they will be different. Doesn't that fix the problem?

Behold this example:
```python
import xarray as xr

def change_len_of_last_axis(a):
    return a[..., :-1]

da = xr.DataArray(np.ones((5, 10)), dims=['x', 'y'])

xr.apply_ufunc(
    change_len_of_last_axis,
    da,
    input_core_dims=['y'],
    output_core_dims=['z'],
    output_sizes={'z': da.shape[da.get_axis_num('y')] - 1}
)
```

rabernat:
And just to clarify, the data model in xgcm is that _every axis position has a distinct dimension name_. Despite sharing coordinate values, `outer` and `inner` should have different dimensions. I agree this is confusing, but it helps keep things simple.

TomNicholas:
Riiight. That's counter-intuitive, but I can see why it's like that. 

New dimensions are allowed to have new sizes, so that should mean this isn't a problem. 

So to summarize: the only time you need a new length on the output is when the output has an (ax:pos) pair that isn't present on the input, but so long as "every axis position has a distinct dimension name" then that new (ax:pos) pair should correspond to a core dim that isn't present on the input, so we don't need `exclude_dims`.

TomNicholas:
This now works for basic 1D cases, but if I chunk them with dask then it will complain that I'm trying to chunk along core dims. What do we want to do about that? For now we could just expose the kwargs for `dask='parallelized/'allowed'` and `allow_rechunk` as optional kwargs to the `apply_grid_ufunc` decorator?

rabernat:
> This now works for basic 1D cases, but if I chunk them with dask then it will complain that I'm trying to chunk along core dims. What do we want to do about that?

This actually gets to a core issue which we need to make a decision about. Our current implementation ultimately uses xarray selections (+ concatentation for boundary conditions) in order to get the pairs of data on which to operate.

https://github.com/xgcm/xgcm/blob/95f4f33d72d2add00136e27f6b3bedecb97d4d77/xgcm/grid.py#L545-L548

This approach is agnostic about whether the "core dim" is chunked or not. It works either way, but performance is poor in both [chunked / unchunked] cases!

If we replace this with, e.g. numba-optimized ufuncs, these ufuncs will presumably require contiguous data along the core dims, thus we would not be able to use `xr.apply_ufunc(..., dask='allowed')`. In this scenario, we would lose a current feature, which is the ability to operate on chunked dimensions.

One workaround for this would be to have an extra layer of logic for chunked arrays that uses [dask.overlap](https://docs.dask.org/en/latest/generated/dask.array.overlap.overlap.html#dask.array.overlap.overlap) or its xarray equivalent. Or perhaps we always go directly to `dask.map_overlap`, which has a fast path for non-chunked dims. Not sure how this would mix with `xr.appy_ufunc`.

The question of how we handle overlap is also intimately related to the question of boundary conditions.

I'd be happy to hop on a call this week to do a little whiteboard brainstorming about this. It's tricky.

dcherian:
> Or perhaps we always go directly to dask.map_overlap, which has a fast path for non-chunked dims. Not sure how this would mix with xr.appy_ufunc

`apply_ufunc(.., dask="allowed")` with `map_overlap` mapping the jitted function will work.

>  chunk boundary

Don't you mean actual boundaries? 2, 3.5, 4 are all boundary conditions at the domain edge. You can `pad` or apply any boundary condition you want and then call `map_overlap` with `boundary=None`. 

cube sphere faces will be complicated but the rest shouldn't be crazy. It'd be nice if we could add `boundary="tripolar"` and have a `Grid.pad` that does the right thing.

TomNicholas:
> apply_ufunc(.., dask="allowed") with map_overlap mapping the jitted function will work.

This is what I was thinking. I'll have a go on a test case now.

rabernat:
> Is there a way we can insert arbitrary values at the chunk boundary like that in dask.map_overlap?

No, but we can write our own custom function to pad the arrays and then call map_overlap.

> cube sphere faces will be complicated but the rest shouldn't be crazy. It'd be nice if we could add `boundary="tripolar"` and have a `Grid.pad` that does the right thing.

Right, I think we want to have a whole padding module that handles these things. The trick will be to avoid unnecessary dask tasks for contiguous core dims. Some numba-based solutions for periodic boundary conditions are reported here: https://github.com/ocean-eddy-cpt/gcm-filters/issues/45#issuecomment-884498140

TomNicholas:
> Right, I think we want to have a whole padding module that handles these things.

Okay, I think this deserves it's own issue.


I've just pushed a test case that uses `apply_ufunc` and `map_overlap` on a chunked 1D example, but I'm getting this weird error, where the undecorated version of the function passes but the decorated one somehow doesn't. I've checked that `grid_ufunc` receives the same arguments in both cases, so I'm not really sure what's going on there...

```python
========================================================================================== FAILURES ==========================================================================================
_____________________________________________________________________________ TestGridUFunc.test_1d_dask_overlap _____________________________________________________________________________

self = <xgcm.test.test_grid_ufunc.TestGridUFunc object at 0x7fd0ceeb1340>

    def test_1d_dask_overlap(self):
        from dask.array import map_overlap
    
        def diff_center_to_left(a):
            return a - np.roll(a, shift=-1)
    
        def diff_overlap(a):
            return map_overlap(diff_center_to_left, a, depth=1, boundary=0)
    
        grid = create_1d_test_grid()
        da = np.sin(grid._ds.x_c * 2 * np.pi / 9).chunk(1)
        da.coords["x_c"] = grid._ds.x_c
    
        diffed = (da - da.roll(x_c=-1, roll_coords=False)).data
        expected = xr.DataArray(diffed, dims=["x_g"], coords={"x_g": grid._ds.x_g})
    
        # Test direct application
        result = grid_ufunc(
            diff_center_to_left,
            da,
            grid=grid,
            dask="allowed",
        ).compute()
        assert_equal(result, expected)
    
        # Test decorator
        @as_grid_ufunc(grid, "(X:center)->(X:left)", dask="allowed")
        def diff_overlap(a):
            return map_overlap(diff_center_to_left, a, depth=1, boundary=0)
    
        result = diff_overlap(da).compute()
>       assert_equal(result, expected)
E       AssertionError: Left and right DataArray objects are not equal
E       
E       Differing values:
E       L
E           array([-3.420201e-01,  1.187823e-01,  5.240053e-01,  6.840403e-01,
E                   5.240053e-01,  1.187823e-01, -3.420201e-01, -6.427876e-01,
E                  -2.449294e-16])
E       R
E           array([-0.34202 ,  0.118782,  0.524005,  0.68404 ,  0.524005,  0.118782,
E                  -0.34202 , -0.642788, -0.642788])

xgcm/test/test_grid_ufunc.py:194: AssertionError

```

rabernat:
Since numba uses a different numerical implementation than numpy, we don't necessary expect exact equality with floating point data. Maybe use `assert_allclose` instead of `assert_equal`?

TomNicholas:
> Since numba uses a different numerical implementation than numpy, we don't necessary expect exact equality with floating point data. Maybe use assert_allclose instead of assert_equal?

Yes good point, but in this case the error is caused by the final value in the array being `-2.449294e-16` instead of `-0.642788`, that's what I don't understand.

jbusecke:
Should `boundary` here
```python
def diff_overlap(a):
            return map_overlap(diff_center_to_left, a, depth=1, boundary=0)
```
be `'periodic'`?

TomNicholas:
> Should boundary here be 'periodic'?

That actually now passes?!

But I'm totally confused as to how - previously I had one passing with `boundary=0` being called through `grid_ufunc`, and one failing also with `boundary=0` , the only difference was that it was being called through the `as_grid_ufunc` decorator. But the `as_grid_ufunc` decorator immediately passes on to `grid_ufunc`, so they should have behaved identically...

Now now my test passes but my original mystery is not solved :sweat_smile:

jbusecke:
So the boundary applies only to the outer parts of the dask array, not the chunks? Am I understanding this right? To get the behavior of `np.roll` on the full domain, you need to have a periodic boundary? Not sure why that would work with 0?

xgcm/grid_ufunc.py
```diff
@@ -0,0 +1,198 @@
...
+    out_core_dims = [
+        [grid.axes[n].coords[p] for n, p in zip(arg_ns, arg_ps)]
+        for arg_ns, arg_ps in zip(out_ax_names, out_ax_pos)
+    ]
+
+    all_out_core_dims = set(dim for arg in out_core_dims for dim in arg)
+
+    # Determine expected output dimension sizes from grid._ds
+    # Only required when dask='parallelized'
+    out_sizes = {out_dim: grid._ds.dims[out_dim] for out_dim in all_out_core_dims}
```
TomNicholas, line 176: Is there any reason why I can't just do this to find the lengths of the output core dimensions? This should work so long as the output `(Axis:Position)` pairs are represented in the `Grid`.

rabernat: Looks reasonable to me!

.pre-commit-config.yaml
```diff
@@ -7,16 +7,16 @@ repos:
...
-    rev: 21.5b1
+    rev: 21.7b0
     hooks:
       - id: black
   - repo: https://github.com/keewis/blackdoc
-    rev: v0.3.3
+    rev: v0.3.4
     hooks:
       - id: blackdoc
   - repo: https://gitlab.com/pycqa/flake8
```
TomNicholas, line 22: This was changed by `pre-commit autoupdate`, which I had to do in order to silence an error

```python
Traceback (most recent call last):
  File "/home/tom/.cache/pre-commit/repo03g1je45/py_env-python3/bin/blackdoc", line 5, in <module>
    from blackdoc.__main__ import main
  File "/home/tom/.cache/pre-commit/repo03g1je45/py_env-python3/lib/python3.8/site-packages/blackdoc/__main__.py", line 12, in <module>
    from .blackcompat import (
  File "/home/tom/.cache/pre-commit/repo03g1je45/py_env-python3/lib/python3.8/site-packages/blackdoc/blackcompat.py", line 9, in <module>
    import toml
ModuleNotFoundError: No module named 'toml'
```
which appeared when trying to work from a new PC. I think this is the same as https://github.com/keewis/blackdoc/issues/100, and possibly should have its own issue.

TomNicholas:
This PR now does (1), (2), and (3) from @dcherian 's [comment](https://github.com/xgcm/xgcm/issues/344#issuecomment-860760874), and there is an example test case which uses `map_overlap` within `xarray.apply_ufunc`. 

The final item is (4) 

> Do something about metrics and boundary

That seems like a can of worms (which I haven't started thinking about properly yet) - perhaps handling metrics and boundaries should be left to a second PR?

jbusecke:
> That seems like a can of worms (which I haven't started thinking about properly yet) - perhaps handling metrics and boundaries should be left to a second PR?

Metrics handling at this point could be simply a multiplication onto the 'diffed'/'interped'/'ufunced' array (presuming we use the ufunc approach for the grid methods).
So if we could first implement a ufunc version of the basic operations like `.diff()` and `.interp() we still get the metrics for free?

In some higher level cases (potential vorticity?) we might want to pass the metrics into the ufunc, but I think your call is right to postpone this.

jbusecke:
I think conceptually, the boundary question is harder to solve, and should be the next big ticket item?

TomNicholas:
A summary of what I need to do:

```python
@as_grid_ufunc(signature="(X:center)->(X:left)")
def diff_center_to_left(a):
    return a - np.roll(a, -1)
```

`diff_centre_to_left` now has an additional `grid` argument. When `diff_centre_to_left` is called it parses the signature, looks up the axes in the grid object, then calls `xarray.apply_ufunc`. So basically the same as what `as_grid_ufunc` already does, just need to change the place where `grid` is passed (`grid` is currently passed to the decorator, but needs to be passed to the decorated function). That's the only real change to this PR needed I think.

```python
class Grid:
    def apply_grid_ufunc(self, grid_ufunc, *args, **kwargs):
        """General method that could be used to call any user-defined grid_ufunc"""
        return grid_ufunc(self, *args, **kwargs)

    def diff(self, da, axis, **kwargs):
        return self.apply_grid_ufunc(diff_center_to_left, da, axis, **kwargs)
```

rabernat:
Yes that seems correct to me!

jbusecke:
I really like this outline: 

> Right. It sounds like the first step will be to just get the grid_ufunc machinery working without touching the existing API. Then we look at how to refactor the implementation of Grid.diff, Grid.interp, etc to use grid_ufunc. Finally, once that is done, we see if we can accelerate the core grid_ufuncs further using numba.

Perhaps somewhere in there it might be good to discuss adding some 'performance metrics' into our tests. I think it would be useful to have some performance tests or even something like the number of dask chunks produced by a certain operation as a baseline. Having this implemented before e.g. optimizing with numba would enable us to track the improvements?

TomNicholas:
> grid is currently passed to the decorator, but needs to be passed to the decorated function

I've made this change in the latest commits, which seems to work nicely. I've also made a new branch in the xgcm repo and changed the base of this PR to merge into that instead (which is why it no longer needs an approving review I believe). Unless anyone has further comments I'm going to merge this and then start changing the `Grid` class in a follow-up PR.

> Perhaps somewhere in there it might be good to discuss adding some 'performance metrics' into our tests.

I'm not sure how easy it is to write automated tests that check number of dask tasks created, but I agree we do need to at least manually check that the refactoring is doing what we hope in terms of performance. After the next PR we should hopefully have two branches to compare, and can maybe use one of Ryan's motivating examples (which was in a gist I have saved somewhere...) as a test case.

TomNicholas:
While discussing with Julius I realised that it would be helpful if the grid ufuncs knew what their signatures were, so I've now promoted the return value of `@as_grid_ufunc` to be a callable `GridUFunc` class instead of a function, which possesses a `.signature` attribute. This is again inspired by how `dask.apply_gufunc` works, and opens the door to having grid ufuncs which know what `boundary` and so on that they can and can't handle.

Issue #324, opened 2021-05-05
Correct handling of interior boundaries (related to "land masks")


rabernat:
I wanted to raise an issue that has come up via @stb2145's work with SODA. 

The basic issue is that `interp` and `diff` are not aware of "interior boundaries", i.e. grid cells that are next to land. In the presence of such boundaries, the methods need to modify their weights to account for the "missing" data, applying some sort of boundary condition (e.g. Dirichlet or Neumann). Without explicit knowledge of land masks, the current implementation either grows the mask on the input data (in the case of NaN-filled data) or else gives a wrong answer near boundaries (in the case of 0- or constant-filled data). That's why I label this a bug.

I put together an example of a workaround outside of xgcm here:
https://nbviewer.jupyter.org/gist/rabernat/ade98c74909271a0d4dd8f7d7c435640

Basically this workaround uses the sort of kernel-based approach we are using in gcm-filters, also prototyped for xgcm in https://github.com/ocean-transport/coiled_collaboration/pull/12.

Shanice: to apply this method on dask-backed data, you would want to use `apply_ufunc` rather than the direct call to the `interp_right_with_land` function shown in the notebook.

Going forward, we need to decide if we want to refactor xgcm to allow for these sorts of kernel functions.

(Discussed tangentially in #293)

rabernat:
I have updated the gist with an example of using this via apply_ufunc:

https://nbviewer.jupyter.org/urls/gist.githubusercontent.com/rabernat/ade98c74909271a0d4dd8f7d7c435640/raw/726d9e3a80b08875d6aa959252388a3794a78fc9/land-aware-interp.ipynb

jbusecke:
This is marked as 'bug', why is the stale bot doing stuff here?

jbusecke:
I think it is worth starting a chat about how to implement this on a high level, since we will soon have an opportunity to implement this similarly to what @rabernat has done above with #344.

On a very general level this involves 'filling' some grid points with values. This might have to involve adding the concept of a 'mask' similar to 'metrics' to xgcm (#293).

I am  proposing the API should look something like this: 
```python
grid.interp(..., internal_boundary='something')
```
Do folks here have thoughts about this signature?

I think this will work nicely from a performance standpoint, since we could wrap the  filling within the ufunc step in the former, but not in the latter. We can discuss later how to exactly do this (e.g. writing a ton of new ufuncs or adding a kwargs and a call to a separate function in the gridops.py module).

[...]

Issue #293, opened 2021-01-05
Masks and partial cells in xgcm [FEATURE]


jbusecke:
We had several issues where the need for some sort of masking came up. This could be helpful for e.g. land/ocean masking 
and also to implement partial cells (which are used by e.g. [mitgcm and mom5])(https://github.com/xgcm/xgcm/issues/193#issuecomment-609480750).

I am in favor of adopting the existing logic for metrics to also support masking and cell fractions as suggested in (https://github.com/xgcm/xgcm/issues/182#issuecomment-592228840).

Right now I would just like to collect thoughts on this in this dedicated issue. 

I could image something like `grid = Grid(..., metrics={...}, masks={}, partial_cells={})` but would like to hear from others about the naming and general opinions on this feature.

cc @miniufo @rabernat

miniufo:
That is good to me, and I also image something like `autogenerate.generate_mask`, in addition to `autogenerate.generate_grid_ds`.  What I am not sure is that if 3D masks are provided, we don't need another kwarg `partial_cells`, do we?

I take a tour at MITgcm source code of calculations of [hdiv](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_hdiv.F#L51), [vort3](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_relvort3.F#L53), [strain](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_strain.F#L53), [tension](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_tension.F#L50), and [KE](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_ke.F#L70).  It is strange to me that while the `hdiv` scheme does include the partial cell (lopping area scheme is [hard coded to be 2](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_fluxform/mom_fluxform.F#L316)), `vort3`, `strain`, and `tension` all ignore lopping factors, and noted as *[masking is implicit in the contents of the (u,v) fields](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_tension.F#L50)*.

While adding such capability is good to me, it seems that the MITgcm folks are [not sure about which scheme is better](https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_hdiv.F#L51).  Any comment on this?

[...]

PR #368, opened 2021-09-10
Grid ufunc refactor - proof of principle

TomNicholas:
Next step towards #344:
- Defines various grid ufuncs in the file `gridops.py`
- Moves the `GCMDataset` class that was in that file to a new file `gcmdataset.py` (do we even still need that code? @jbusecke said that was some of your old code @rabernat ?)
- Starts work on a new `Grid._grid_ufunc_dispatch` method that aims to eventually replace the existing `Grid._grid_func` method and everything that calls below (including `Grid._neighbour_binary_func` etc.).
- Sketches out a mechanism to search through all the grid ufuncs defined in `gridops.py` ~~(but the searching doesn't actually work yet, still a WIP)~~ this now works, next step is to add support for a `boundary` kwarg so that some of the tests actually pass.

Builds on top of #362, so actually there are far fewer changes than this looks like (I don't know why this still lists `test_grid_ufunc.py` as a new change...)

 - [x] Continues work towards #344
 - [x] Tests added


rabernat:
> * do we even still need that code?

No. It's very old and unused.

TomNicholas:
> No. It's very old and unused.

Okay I'll just delete the whole class then.

TomNicholas:
So I'm adding the `boundary` kwargs, and I would have liked to have a `gridops.py` module that looked like this:

```python
@as_grid_ufunc(signature="(X:center)->(X:left)", boundary="fill")
def diff_center_to_left(a):
    return a - np.roll(a, -1)


@as_grid_ufunc(signature="(X:center)->(X:left)", boundary="extend")
def diff_center_to_left(a):
    return a - np.roll(a, -1)
```
where you have identically-named functions distinguished only by their keyword arguments to the decorator. 

However I don't think this is possible, as despite the members of this module being different instances of the `GridUFunc` class, python refers to both these functions via their name (`diff_center_to_left`), and so the second one ends up overwriting the first, and I have to give each function different names even if they are distinguished by their decorator arguments, i.e.

```python
@as_grid_ufunc(signature="(X:center)->(X:left)", boundary="fill")
def diff_center_to_left_fill(a):
    return a - np.roll(a, -1)


@as_grid_ufunc(signature="(X:center)->(X:left)", boundary="extend")
def diff_center_to_left_extend(a):
    return a - np.roll(a, -1)
```

rabernat:
> so the second one ends up overwriting the first, and I have to give each function different names even if they are distinguished by their decorator arguments

This might be suggesting our abstraction is wrong.

First let's talk carefully about what we mean by _boundary conditions_. I distinguish between two different types of boundary:
- The **global array boundary**: obviously our array has a finite size (Ny, Nx). We need to decide what to do at the boundaries of that array. _This is what xgcm's current "boundary" kwarg is about._ Those boundaries may or may not have any physical meaning. For example, on a lat lon grid, the X-axis array boundary has zero physical meaning--it just happens to be where we sliced the sphere. This is ultimately a topological issue. Our implementation of [face connections](https://xgcm.readthedocs.io/en/latest/grid_topology.html) has some very fancy treatment for the global array boundaries, e.g. for cubed-sphere grids.
- The **physical boundary**: ocean models (and probably other FV models) have actual boundaries, where there is no fluid, i.e. land. We often want to do something distinct on these boundaries, like apply a no-flux boundary condition. These may or may not correspond with the global array boundary. This is discussed in #324. Also closely related to #293. Currently _not supported in xgcm_.

It's very useful to think about what real GCMs do. GCMs parallelize the grids over many MPI processes, each operating on a "tile." Each title only knows about its local data, so the concept of _global array boundaries_ is not relevant. ([Example](https://mitgcm.readthedocs.io/en/latest/software_arch/software_arch.html#specifying-a-domain-decomposition))
Instead, each tile has an "overlap" or "halo" region that is duplicated with its neighbors, plus a land mask.

grid

**I would like to suggest that `as_grid_ufunc` should be unaware of _global array boundaries_ but possibly aware of _physical boundaries_.** Global array boundaries can be handled separately. For example, when we call `Grid.diff`, we might imagine taking the following steps.

1. Build halo region around the array
2. Apply `diff_center_to_left` array_ufunc
3. Remove halo region

TomNicholas:
So there are a two issues here - we're not quite talking about the same thing:

1) **How should grid ufuncs with the same signature be distinguished?** This is what my comment was about, rather than specifically about the `boundary` kwarg, because we will have this naming issue whether or not it's `metric` or `fill_value` or `boundary` that distinguishes the grid ufuncs. (It is possible though that the correct abstraction doesn't have any kwargs other than `signature` in the `@as_grid_ufunc` decorator, and everything else is either always attached to `grid` or handled before/after the ufunc...)

2) **Which keyword arguments do grid ufuncs need to be aware of?** This is what your comment discusses @rabernat . My understanding is that you're basically saying that we might need something like `@as_grid_ufunc(signature, physical_boundary={'X': 'dirichlet'})`, but we shouldn't need `@as_grid_ufunc(signature, boundary='fill')`, because that should instead be handled by code that wraps the application of the grid ufunc? In that model is the current `boundary` kwarg any different to the `boundary` kwarg that `dask.array.map_blocks` accepts? 



I've actually realised that I have a more immediate design problem to solve first though: **how do I check that two signatures are functionally equivalent?** i.e. `"(X:center)->(X:right)"` and `"(Y:center)->(Y:right)"` are equivalent, in the same way that for normal ufuncs `(i)->(i)` and `(j)->(j)` are equivalent.

TomNicholas:
> For example, when we call Grid.diff, we might imagine taking the following steps.
> 
> - Build halo region around the array
> - Apply diff_center_to_left array_ufunc
> - Remove halo region

Isn't this literally what `dask.array.map_overlap` does? If so then can we just use that? 

If we do then the `boundary` question mostly becomes whether the call to `map_overlap` should be part of the grid ufunc or not. There are a few options I can see:

1) **Call `dask.array.map_overlap` inside each grid ufunc.** This is currently supported, and there is actually a test for this case in #362. The main disadvantage is potentially duplicated code within many (if not all) of the pre-defined ufuncs in `gridops.py`. It would also mean that `boundary` needs to be an argument to each grid ufunc, which I think conflicts with your recommendation above @rabernat .

2) **Automatically wrap each grid ufunc with `dask.array.map_overlap`.** If we know all pre-defined grid ufuncs should have the same behaviour and options with respect to internal boundaries, then we can take the supplied numpy ufunc, wrap it in `dask.array.map_overlap`, and then feed the resulting function into `xarray.apply_ufunc`. We could do this within the `apply_as_grid_ufunc` call, so it would happen within the `as_grid_ufunc` decorator, and not be mentioned at all in the grid ufunc definition. The disadvantage is that this approach relies on the assumption that the underlying numpy implementation of the grid ufunc is independent of the choice of `boundary={'fill', 'extend', 'extrapolate'}`. It might also complicate the implementation of different physical boundaries.

3) **If xarray implements `map_overlap` then that might end up being a better way?** See https://github.com/pydata/xarray/issues/3147. I don't think this is the correct solution for us though. 

One problem with this idea of using `dask.array.map_overlap` though is that while xgcm's `'fill'` is essentially dask's constant value option, and xgcm's `extend` is essentially dask's `'nearest'` options, it doesn't look like `dask.array.map_overlap` has any equivalent of xgcm's `'extrapolate'` option.

rabernat:
> Isn't this literally what `dask.array.map_overlap` does? If so then can we just use that?

Not quite.
- [map_overlap](https://docs.dask.org/en/latest/array-overlap.html#dask.array.map_overlap) as is is not sufficiently flexible to handle the really global array boundary conditions. It supports the following boundary conditions:
  >  ‘reflect’, ‘periodic’, ‘nearest’, ‘none’, or any constant value like 0 or np.nan. 
  
  It can't deal with cubed sphere, tripolar grid, etc. So we may have to vendor this function and augment it with special options.
- Xgcm should work without Dask! It has always supported an eager evaluation mode.

However, we can overlook these in the short term if we want to just move forward with an implementation.

rabernat:
> **how do I check that two signatures are functionally equivalent?** i.e. `"(X:center)->(X:right)"` and `"(Y:center)->(Y:right)"` are equivalent, in the same way that for normal ufuncs `(i)->(i)` and `(j)->(j)` are equivalent.

It seems like a regex could handle this, no?

TomNicholas:
> It can't deal with cubed sphere, tripolar grid, etc. 

Hmm okay.

> we may have to vendor this function and augment it with special options.

I haven't heard the word "vendor" used in this context before - what exactly do you mean? Make our own version of `map_overlap`?

I think if we had our own function to perform a similar job (e.g. `map_over_faces` or something) then we would still have to choose between my (1) and (2) above.

> However, we can overlook these in the short term if we want to just move forward with an implementation.

This might be a good idea - use `dask.array.map_overlap` to get the basic test cases with just `'fill'` and `'extend'` working, and then re-evaluate.

> It seems like a regex could handle this, no?

Probably, but so far my solution has been to call `_parse_grid_ufunc_signature` (so re-using my previous regex in a way) and then replace dummy variables. There is probably a neater regex way though.

https://github.com/xgcm/xgcm/pull/368/files#diff-3af42ae01706abd10c8c0a6185e2598b72bacfe3ef33eb162ae8da1a9d61a257R307-R335

dcherian:
I think you want to build `Grid.pad` first so then making the halo is a call to `Grid.pad` (accounts for tripole, cubed-sphere, periodic etc.); then pass to `map_overlap(..., boundary=None)` so map_overlap only takes care of communicating between blocks; then remove the halo region.

jbusecke:
I like the suggestion by @dcherian above. This would take care of the 'data_boundaries' (which are not physical) and then 'internal_boundaries' e.g. actual land in the data should be handled internally (at the numpy level) in the ufunc. The latter would be a new feature (to be implemented after this refactor?) and we could discuss if we want to implement that with an additional keyword or with another subset of ufuncs.

TomNicholas:
>I think you want to build Grid.pad first so then making the halo is a call to Grid.pad (accounts for tripole, cubed-sphere, periodic etc.); then pass to map_overlap(..., boundary=None) so map_overlap only takes care of communicating between blocks; then remove the halo region.

Thanks @dcherian , that was a really helpful suggestion.

So in the latest commit I added a prototype of a `grid.pad` method, which is called within `apply_as_grid_ufunc`, and pads the given dataarrays by a number of elements specified by a new `boundary_width` argument to the `@as_grid_ufunc` decorator. (Thanks to @jbusecke for a very helpful pair programming session on that on Friday.)

We also identified the simplest existing failing test case that needs to pass in order to prove that this approach works: `xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_left]` - I was trying to debug via some other tests before.

The padding seems work for basic cases (though I still need to write tests that will isolate that method), but the test above still doesn't pass, instead failing with an error I told it to raise when the ufunc output hasn't been trimmed:

```python
ValueError: conflicting sizes for dimension 'XG': length 101 on <this-array> and length 100 on {'XG': 'XG'} - does your grid ufunc correctly trim off the same number of elements which were added by padding using boundary_width={'X': (1, 0)}?
```

I think the problem is that @jbusecke the `boundary_width` kwarg we added to `diff_center_to_left` in `gridops.py` isn't correct: when `diff` is called with `boundary='fill'` by the test, this ends up padding, then applying the ufunc, but the ufunc does not automatically trim the new elements back off the data, so the result is the incorrect length.

```python
@as_grid_ufunc(signature="(X:center)->(X:left)", boundary_width={"X": (1, 0)})
def diff_center_to_left(a):
    return a - np.roll(a, -1)
```

If this particular ufunc isn't supposed to have any elements padded when applying boundary conditions, then I need another example ufunc which does. In general it would be helpful if I were actually sure what the ufuncs in `gridops.py` for the various `diff_*` variations were actually supposed to be!

rabernat:
> but the ufunc does not automatically trim the new elements back off the data, so the result is the incorrect length.

So the implementation needs to change
```python
@as_grid_ufunc(signature="(X:center)->(X:left)", boundary_width={"X": (1, 0)})
def diff_center_to_left(a):
    return a[..., 1:] - a[..., :-1]
```

TomNicholas:
Thanks @rabernat - I expect that will work.

> So the implementation needs to change

There is possibly another option, where I add a `trim_after` argument to`@as_grid_ufunc`, that defaults to false. That would allow people to define ufuncs which don't trim the padded elements, and trim for them instead.

jbusecke:
> There is possibly another option, where I add a trim_after argument to@as_grid_ufunc, that defaults to false. That would allow people to define ufuncs which don't trim the padded elements, and trim for them instead.

I am actually not against that. It would enable both types of workflow and might make these 'basic' ufuncs cleaner to read?

dcherian:
> a[..., 1:] - a[..., :-1]

I think this is actually better from a memory perspective two, since we create two slices instead of one extra copy with `np.roll`

:+1: for a `trim_after` argument, though the analogous argument in `map_overlap` is just `trim`

rabernat:
My general point is that, if padding is handled separately, our grid ufuncs should **never** roll arrays or do anything else that requires global array boundary communication. They should always assume they are operating on a "local tile" of data.

I agree that `trim` may be needed, depending on the particular implementation. Perhaps this can be optional, for the grid ufunc writer to choose?

TomNicholas:
> My general point is that, if padding is handled separately, our grid ufuncs should never roll arrays or do anything else that requires global array boundary communication. 

I think that now all our basic ufuncs are just going to look like slices, e.g. `a[..., 1:] - a[..., :-1]`. The global array communication, if needed, is being handled (in a dask-aware way) by `DataArray.pad` now.

> Perhaps this can be optional, for the grid ufunc writer to choose?

Yes I meant as an additional option for someone writing a custom ufunc.

xgcm/grid.py
```diff
@@ -1558,6 +1568,219 @@ def _grid_func(self, funcname, da, axis, **kwargs):
...
+            Supplied in a mapping of the form {axis_name: (lower_width, upper_width)}.
+        boundary : {None, 'fill', 'extend', 'extrapolate', dict}, optional
+            A flag indicating how to handle boundaries:
+            * None: Do not apply any boundary conditions. Raise an error if
+              boundary conditions are required for the operation.
+            * 'fill':  Set values outside the array boundary to fill_value
+              (i.e. a Dirichlet boundary condition.)
+            * 'extend': Set values outside the array to the nearest array
+              value. (i.e. a limited form of Neumann boundary condition.)
+            * 'extrapolate': Set values by extrapolating linearly from the two
```
jbusecke, line 1590: Did you figure out a way to implement `extrapolate`? Otherwise I think we should not mention it in the docstring?

TomNicholas: I haven't tried `extrapolate` yet. I have other problems with that docstring though - for example the way that `None` really means `"periodic"`...

xgcm/grid.py
```diff
@@ -1624,6 +1624,9 @@ def pad(self, *arrays, boundary_width=None, boundary=None, fill_value=None):
                 if ax_boundary == "extrapolate":
                     # TODO implement extrapolation
                     raise NotImplementedError
+                elif ax_boundary is None:
+                    # TODO this is necessary, but also seems inconsistent with the docstring, which says that None = "no boundary condition"
+                    ax_boundary = "periodic"
```
jbusecke, line 1629: Why is this necessary? I think the problem is that in the init of the Grid class `periodic` is usually True, and a lot of tests will thus implicitly assume a periodic boundary. 
The simplest fix to that might be to define an override [somewhere here](https://github.com/xgcm/xgcm/blob/95f4f33d72d2add00136e27f6b3bedecb97d4d77/xgcm/grid.py#L1147-L1158) that basically does something like: 
```
# for each axis:
if isperiodic:
    boundary = 'periodic'
```
That way we internally drop the seperate kwarg `periodic` but the user can (for now) still use the existing API.

That should hopefully make all legacy tests pass. I am curious if after that change this is still necessary. I would like to avoid setting periodic here, and rather spit out an error if boundary is still none (after the changes suggested above).

TomNicholas: > Why is this necessary? I think the problem is that in the init of the Grid class periodic is usually True, and a lot of tests will thus implicitly assume a periodic boundary.

I think its something like that - related to the redundancy of the `periodic` argument that you mentioned last week.

> That way we internally drop the seperate kwarg periodic but the user can (for now) still use the existing API.

This seems like a good idea.

> I would like to avoid setting periodic here, and rather spit out an error if boundary is still none (after the changes suggested above).

Using `boundary=None` to mean a specific boundary condition choice is sort of inconvenient because to be able to override the Axis' default choice of bc you want the default value of the boundary kwarg in `diff` etc. to be `None` as in "no value provided".

TomNicholas: > Why is this necessary?

It's necessary because if you don't add something like this the existing periodic tests raise an error from `.pad` that `boundary=None` is not a supported type of boundary.

jbusecke: > It's necessary because if you don't add something like this the existing periodic tests raise an error from .pad that boundary=None is not a supported type of boundary.

But they hopefully would not do that anymore if the default `periodic` is propagated from the init call of `Grid`?

xgcm/grid.py
jbusecke: But that is `None` too?

TomNicholas:
I realised there was a major omission in #362 - `apply_as_grid_ufunc` had no `axis` argument! #371 fixes that, so this PR now will need to build on top of #371, which needs to be merged before this one.

TomNicholas:
For some reason now the trimming works in 1D but has an off-by-one error in 2D :confused: 

```
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_outer] PASSED                                                                                                                     [ 10%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_inner] PASSED                                                                                                                     [ 20%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_left] PASSED                                                                                                                         [ 30%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_left] PASSED                                                                                                                      [ 40%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_right] PASSED                                                                                                                        [ 50%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_right] PASSED                                                                                                                     [ 60%]
xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] FAILED                                                                                                                         [ 70%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] FAILED                                                                                                                      [ 80%]
xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] FAILED                                                                                                                        [ 90%]
xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] FAILED                                                                                                                        [100%]


FAILED xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} - ...
FAILED xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'}...
FAILED xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
FAILED xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
```

TomNicholas:
For some reason now the trimming works in 1D but has an off-by-one error in 2D :confused: 

```
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_outer] PASSED                                                                                                                     [ 10%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_inner] PASSED                                                                                                                     [ 20%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_left] PASSED                                                                                                                         [ 30%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_left] PASSED                                                                                                                      [ 40%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_right] PASSED                                                                                                                        [ 50%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_right] PASSED                                                                                                                     [ 60%]
xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] FAILED                                                                                                                         [ 70%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] FAILED                                                                                                                      [ 80%]
xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] FAILED                                                                                                                        [ 90%]
xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] FAILED                                                                                                                        [100%]


FAILED xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} - ...
FAILED xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'}...
FAILED xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
FAILED xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
```

TomNicholas:
For some reason now the trimming works in 1D but has an off-by-one error in 2D :confused: 

```
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_outer] PASSED                                                                                                                     [ 10%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_inner] PASSED                                                                                                                     [ 20%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_left] PASSED                                                                                                                         [ 30%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_left] PASSED                                                                                                                      [ 40%]
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_right] PASSED                                                                                                                        [ 50%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_right] PASSED                                                                                                                     [ 60%]
xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] FAILED                                                                                                                         [ 70%]
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] FAILED                                                                                                                      [ 80%]
xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] FAILED                                                                                                                        [ 90%]
xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] FAILED                                                                                                                        [100%]


FAILED xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} - ...
FAILED xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'}...
FAILED xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
FAILED xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] - ValueError: conflicting sizes for dimension 'YG': length 199 on <this-array> and length 200 on {'XC': 'XC', 'YG': 'YG'} -...
```

TomNicholas:
This now passes for 2d data too (despite not knowing what is wrong with https://github.com/xgcm/xgcm/pull/371#issuecomment-926220770)...

```
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_outer] PASSED
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_inner] PASSED
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_left] PASSED
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_left] PASSED
xgcm/test/test_grid.py::test_grid_ops[periodic_1d_right] PASSED
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_1d_right] PASSED
xgcm/test/test_grid.py::test_grid_ops[periodic_2d_left] PASSED
xgcm/test/test_grid.py::test_grid_ops[nonperiodic_2d_left] PASSED
xgcm/test/test_grid.py::test_grid_ops[xperiodic_2d_left] PASSED
xgcm/test/test_grid.py::test_grid_ops[yperiodic_2d_left] PASSED
```

but only up to a transposition. Specifically, I had to add 

```python
if axis_name == "Y":
    assert da_diff.T.equals(da_diff_ax)
else:
    assert da_diff.equals(da_diff_ax)
```
to the test to get all the `[*periodic_2d_left]` tests to pass. This is because `xarray.apply_ufunc` always transposes so that the axes being acted over are the last axes in the data.

Now I'm not sure what to do here because I could:
1) Write some code that attempts to transpose the result back - unfortunately I think this is not a trivial problem, given that a grid ufunc can create and remove arbitrary axes. It might not always be clear which dim has been shifted to create which new dim...
2) Change the test - I've not changed any tests at all yet, and don't want to. But it is interesting that this would pass if xarray's assert functions didn't care about dimension order, as you suggested in https://github.com/pydata/xarray/issues/5733 @jbusecke .
3) Leave it for now - this particular test only compares `grid.diff` to `axis.diff`, and the latter is still using the old code path. I expect that when everything has been refactored internally then this particular tests will pass (though any other tests elsewhere that apply over multidimensional data will presumably fail with the same problem).

TomNicholas:
Actually I think I've got it - I don't need to deal with the general problem of keeping dimension order, only the (much easier) case of one-to-one functions like `diff`, `interp` etc. That way I can transpose back for those functions defined on `grid`, but the fact that `apply_as_grid_ufunc` will stll rearrange dims shouldn't really matter.

TomNicholas:
[I tested this with dask in this notebook](https://gist.github.com/TomNicholas/47be490cbeae03686a24de87f505f364), but found that confusingly applying these two functions behave differently with respect to dask:

```python
def diff_center_to_left(a):
    return a[..., 1:] - a[..., :-1]
```

```python
def manual_diff(array):
    return array - np.roll(array, 1, axis=-1)
```
The former complains about chunking along a core dim
```
ValueError: dimension x_c on 0th function argument to apply_ufunc with dask='parallelized' consists of multiple chunks, but is also a core dimension. To fix, either rechunk into a single dask array chunk along this dimension, i.e., ``.chunk(dict(x_c=-1))``, or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` but beware that this may significantly increase memory usage.
```
whilst the latter doesn't. I don't really understand why that is. Currently this PR uses the former definition in the `gridops.py` file, which is why you see the same behaviour when calliing `grid.diff` in the notebook.

Using bare `apply_ufunc` behaves the same way as `apply_as_grid_ufunc` so I don't think it's my code causing any difference in behaviour.

Additionally I tried the former function but with `dask='allowed'` instead of `'parallelized'` - that runs but gives a surprisingly complex task graph :confused: 

So it appears that a grid ufunc defined with `np.roll` is most dask-streamlined, but I thought we were expecting the opposite?

Note: This is all before I attempt to call `map_overlap` anywhere - currently all that happens is a `dask.pad` (if `boundary_width` is nonzero) and then an `xarray.apply_ufunc`.

rabernat:
> this would be a good time for a review / input before I try to move any further forward I think

I just put this on my calendar in my next free spot, in the morning of Oct. 4.

dcherian:
Is the graph just as complex if you called the second one with `dask="allowed"`?  It should  be...

The graph for `dask="parallelized"` is neat because it's just applying it blockwise; but that  can only work if there's a single chunk along that axis.

You're getting a chunk error for the second one because pad inserts an extra chunk IIRC with the padded  values.

Anecdotally, I know dask slicing can get weirdly complex, but don't really understand why. Isn't that why we're doing this refactor in the first place?  

I think should try `dask="allowed"` with `map_overlap` and the first function to get a real sense of the before/after improvement (if any).

TomNicholas:
> Is the graph just as complex if you called the second one with dask="allowed"? It should be...

You actually get a 3rd different graph, of moderate complexity :confused: I've updated the gist with it.

> Anecdotally, I know dask slicing can get weirdly complex, but don't really understand why. Isn't that why we're doing this refactor in the first place?



> I think should try dask="allowed" with map_overlap and the first function to get a real sense of the before/after improvement (if any).

I can do that - I can define the grid ufunc to call `map_overlap` within it, like this:

```python
from dask.array import map_overlap

def diff_center_to_left(a):
    return a[..., 1:] - a[..., :-1]

def diff_overlap(a):
    return map_overlap(
        diff_center_to_left,
        a,
        depth={'X': (1, 0)},
        boundary=None,
        trim=True,
        meta=np.array(())
    )

bar_manual_diff = grid_test.apply_as_grid_ufunc(
    diff_overlap, 
    ds_test.bar, 
    axis='X', 
    boundary_width={"X": (0, 0)},
    boundary="fill",
    dask="allowed",
)
bar_manual_diff
```
which gives a nice streamlined graph. (I am a bit confused by why I needed to put `trim=True` and `boundary_width={'X': (0,0)}` to make this compute an array of the correct length - I thought I would need to put `trim=True` and `boundary_width={'X': (1,0)}`...)

Still, I think the before / after graph comparison (so using this `map_overlap` code above) looks like this

(before)
image

(after)
image

rabernat:
Merging this for now. We can keep iterating on the dask optimizations in future PRs.

PR #369, opened 2021-09-13
Correction of the Grid.__init__ docstring

rcaneill:
The keep_coord parameter of the docstring is in fact not a parameter of the `Grid.__init__` method.
I imagine that it is an error of the docstring.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/keewis/blackdoc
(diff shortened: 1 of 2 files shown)


jbusecke:
Thanks for catching this @rcaneill. I  believe this was introduced [here](https://github.com/xgcm/xgcm/pull/186). To me it seems you are right in removing this, but maybe @apatlpo could take a look before we act on this?

jbusecke (approved):
This looks good to me. @TomNicholas do you agree to merge this in now?

PR #374, opened 2021-10-04
Remove stale bot

jbusecke:
I personally feel that this bot causes more unintended closures (and thus work reopening) than it saves me. 

I think it would be better if I focus on actually going through issues a bit more, and we get rid of the bot.

Thoughts?


TomNicholas:
I think you can turn down the aggression of the bot?

jbusecke:
I think I realized that most of our issues are actually not stale really. Just suggestions for future features, so I kinda dont want to get rid of them?

rabernat:
YES

PR #379, opened 2021-10-05
Fix test_weighted_metric_multi_axis so it actually runs

TomNicholas:
I noticed that one of the tests in `xgcm/test/test_metrics_ops_single_axis.py` was over-indented, which meant that pytest was not picking it up, and not running it at all!

That test (`xgcm/test/test_metrics_ops_single_axis.py::TestParametrized::test_weighted_metric_multi_axis`) was heavily parametrized, so fixing this bug increases the total number of tests run by the test suite by over one thousand :exploding_head: 
Luckily they all pass on `master`!

I also had to move some of the parameterization decorators to make sure the tests in this class were all being passed their intended inputs.


TomNicholas:
5068 passed here vs 3868 passed on `master` :exploding_head:

TomNicholas:
While we're here is there really a need for a `test_metric_ops_multi_axis.py` file and a separate `test_metric_ops_single_axis.py` file?

rabernat:
> While we're here is there really a need for a `test_metric_ops_multi_axis.py` file and a separate `test_metric_ops_single_axis.py` file?

Probably not. I can't remember why it is that way. However, in the interest of focusing effort on fun things, I would say: we [now] have working tests and good coverage. Refactoring this test suite is probably not worth lots of time. If you can see a quick way to do it, then go for it. Otherwise LGTM and let's move on.

PR #380, opened 2021-10-08
map_overlap in grid ufunc refactor

### LINKED ISSUE #378: dask and map_overlap in the grid_ufunc refactor
Continuing on from the discussion in #368 about how exactly to employ `xarray.apply_ufunc`, `dask.array.map_overlap`, `pad`, and the new grid_ufunc definitions to get clean dask graphs.

(@rabernat your notebook can live here)

TomNicholas:
WIP attempt to add option to use `dask.array.map_overlap` in conjunction with `xarray.apply_ufunc` to parallelize grid ufuncs over chunked core dims, using method proposed by @rabernat in #378.

 - [x] Will close #378
 - [x] Tests added


TomNicholas:
This now works for the case of one input da (and one output da, of the same size after padding + application) in 1D. 

[This is an updated version of @rabernat 's notebook](https://gist.github.com/TomNicholas/ae36b14f86ec91799d7645c0c24754aa) which shows that the dask task graph is the same :champagne: That's especially good because I didn't bother to drop down from xarray when padding or rechunking, nor did I bother to compose those functions with the `map_overlap` call, but it seems dask is clever enough to still give us the same graph!

xgcm/grid_ufunc.py
```diff
@@ -336,30 +360,106 @@ def apply_as_grid_ufunc(
...
+            # create the "true" output chunks
+            # TODO first argument only because map_overlap can't handle multiple return values (I think)
+            true_chunksizes = original_chunksizes.copy()[0]
+            # TODO how should this operation work? Which specific chunk should it adjust the length of?
+            # (as is it wouldn't work because we're adding an int to a tuple)
+            true_chunksizes[core_dim_which_is_changing_size] = (
+                true_chunksizes[core_dim_which_is_changing_size]
+                + output_chunk_length_adjustment
+            )
+            true_chunksizes_per_numpy_axis = _dict_to_numbered_axes(true_chunksizes)
```
TomNicholas, line 433: @rabernat which specific chunk should I be changing the size of here?

TomNicholas: It's meant to be equivalent to the logic in cell 19 of [your notebook](https://gist.github.com/rabernat/0135a7b03b46705da7877147a3736aa8), i.e.

```python
orig_chunks = data_padded_rechunked.chunks
core_dim_chunks = orig_chunks[-1]
true_core_dim_chunks = ((core_dim_chunks[0]-1,) + core_dim_chunks[1:],)
true_chunks = orig_chunks[:-1] + true_core_dim_chunks

data_diffed = dsa.map_overlap(
    diff,
    data_padded_rechunked,
    depth={1: (1, 0)},
    boundary='none',
    trim=False,
    meta=np.array([], dtype=data_padded_rechunked.dtype),
    chunks=true_chunks
)
data_diffed
```

rabernat: If memory serves, this "chunk hacking" is directly tied to `boundary='none'` and `depth={1: (1, 0)},`. By specifying `trim=False`, we tell `map_overlap` that our function will take care of the trimming. This works great for all of the chunks that do not touch the global boundary, which are padded before being passed to our function `diff`. However, at this point, we have applied the global array boundary condition manually (creating a "padded" array), thus growing the array by one point in a particular direction (left, right, or both). The applied function (here `diff`) also trims those boundary chunks, but Dask is not expecting that to happen (since we specified `boundary=none`), so without `chunks=true_chunks`, it guesses the wrong chunk shape. (Try removing that option and see what happens.)

So there is a symmetry between what you pass to `depth` and how you set `true_chunks`. The key `1` tells you that we need to modify the expected chunks on axis 1. The value `(1, 0)` tells you that we need to shrink the first chunk along that axis by 1 element. `(0, 1)` would mean shrink the last chunk.

Make sense?

TomNicholas:
After getting really confused by the details of how to handle a chunked core dimension that could potentially change size, @rabernat and I decided to restrict this PR so that if a grid ufunc's signature includes the positions "inner" or "outer", then chunking along core dimensions is not allowed. This simplifies the code considerably because you can be sure that chunk sizes never change. We *might* be able to relax this in future, but I would want to see a convincing use case to justify it first.

xgcm/grid_ufunc.py
TomNicholas: This discussion is made redundant by https://github.com/xgcm/xgcm/pull/380/commits/070cddd70b46bd76f01191f31bad966320aec2e6

TomNicholas:
@rabernat commit https://github.com/xgcm/xgcm/pull/380/commits/b0dafb0b39f2c773ec48ec4e88ee7018ba21ea4c xfails tests which fail due to either the metrics not yet working with the refactor or connected faces not yet working with the refactor. 

Turns out [pytest lets you easily mark certain parameter combinations as failing](https://docs.pytest.org/en/6.2.x/skipping.html#skip-xfail-with-parametrize) which is useful.

xgcm/test/test_metrics_ops.py
```diff
@@ -6,8 +6,19 @@
 from xgcm.grid import Grid
 from xgcm.test.datasets import datasets_grid_metric
 
+GRID_UFUNC_REFACTOR_ERR = "Metrics have not yet been implemented for the grid ufunc refactor, and diff is currently the only function wwhich uses the new code path"
```
rabernat, line 9:
```suggestion
GRID_UFUNC_REFACTOR_ERR = "Metrics have not yet been implemented for the grid ufunc refactor, and diff is currently the only function which uses the new code path"
```

rabernat (commented):
This looks great Tom!

My comments are mostly just places where I was having a hard time following the code.

xgcm/grid_ufunc.py
```diff
@@ -335,30 +362,83 @@ def apply_as_grid_ufunc(
...
+
+        # merge any lonely chunks from padding
+        rechunked_padded_args = _rechunk_to_merge_in_boundary_chunks(
+            padded_args,
+            args,
+            boundary_width_real_axes,
+            grid,
+        )
+
+        # TODO should this be adjusted?
```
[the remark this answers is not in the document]
TomNicholas: Adjusted in the sense of "do we need to adjust the chunksize when dealing with the case of the length of an axis changing during a grid ufunc operation?". Now unnecessary because we chose not to support that case.

xgcm/grid_ufunc.py
```diff
@@ -335,30 +362,83 @@ def apply_as_grid_ufunc(
...
-    # TODO Map operation over dask chunks?
-    # def mapped_func(*a, **kw):
-    #    return map_overlap(func, *a, **kw, depths=boundary_depths, boundary=None, trim=True)
 
     # Perform operation via xarray.apply_ufunc
     results = xr.apply_ufunc(
-        func,
-        *args,
+        mapped_func,
+        *rechunked_padded_args,
```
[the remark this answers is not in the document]
TomNicholas: They are the args, exclusively. We can't easily support kwargs as xarray data objects because
a) IIRC `apply_ufunc` only does alignment and so on to xarray objects that are passed as args, any kwargs passed to `apply_ufunc` are just immediately passed on to the wrapped numpy ufunc,
b) How would our signature specify which kwarg was in which grid position? We would have to change the format at least.

I'll try and make this clearer in the docstrings.

xgcm/grid_ufunc.py
```diff
@@ -335,30 +362,83 @@ def apply_as_grid_ufunc(
...
+        # TODO should this be adjusted?
+        boundary_width_per_numpy_axis = {
+            grid.axes[ax_name]._get_axis_dim_num(args[0]): width
+            for ax_name, width in boundary_width_real_axes.items()
+        }
+
+        def _dict_to_numbered_axes(sizes):
+            return tuple(sizes[dim] for dim in sizes.keys())
+
+        output_chunk_length_adjustment = _check_if_length_would_change(  # noqa
```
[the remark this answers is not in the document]
TomNicholas: Again it's accidentally left over from when I was trying to deal with the case that the length might change.

xgcm/grid_ufunc.py
```diff
@@ -1,9 +1,16 @@
...
 import string
+from typing import TYPE_CHECKING, Any, Callable, List, Mapping, Sequence, Tuple, Union
 
 import numpy as np
 import xarray as xr
 
+if TYPE_CHECKING:
+    # Avoids circular references when type checking
+
+    from .grid import Grid
```
rabernat, line 11: I'm a bit confused by this. What is the circular reference? Does `grid` import this module?

Could it possibly be avoided by [`from __future__ import annotations`](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations)?

xgcm/grid_ufunc.py
```diff
@@ -133,7 +147,13 @@ def __repr__(self):
             f"          dask='{self.dask})', map_overlap={self.map_overlap})"
         )
 
-    def __call__(self, grid=None, *args, axis, **kwargs):
+    def __call__(
+        self,
+        grid: "Grid" = None,
```
rabernat, line 152: Does `Optional[Grid]` not work here?

xgcm/grid_ufunc.py
```diff
@@ -555,7 +586,10 @@ def _rechunk_to_merge_in_boundary_chunks(
 
 
 def _get_chunk_pattern_for_merging_boundary(
-    grid, da, original_chunks, boundary_width_real_axes
+    grid: "Grid",
```
rabernat, line 589:
```suggestion
    grid: Grid,
```

I've not seen these quoted class names before. Is that related to the circular import?

rabernat (commented):
This is awesome, thanks Tom! The type annotations definitely help understand some of the functions better.

Have you actually run mypy on this code? Now that we do have type annotations, we will probably want to add mypy to the pre-commit config.

I'm actually not seeing the [pre-commit workflow](https://github.com/xgcm/xgcm/blob/master/.github/workflows/linting.yaml) at all on this PR. I think that's because it configured to only run on PRs to master. You might want to edit that so we can see the pre-commit results for these PRs to the `grid_ufunc_refactor_project` branch.

xgcm/grid_ufunc.py
TomNicholas: > Does grid import this module?

Yes - this refactor will mean that e.g. `grid.diff` calls `apply_as_grid_ufunc`, so `grid` has to import this module.

> Could it possibly be avoided by from __future__ import annotations?

That allows me to swap out `"Grid"` for `Grid`, but I still can't remove the `if TYPE_CHECKING` without causing a circular reference error...

xgcm/grid_ufunc.py
TomNicholas: Yes it's allows for a form of delayed evaluation of the type IIUC, but I was able to remove this - see [comment above](https://github.com/xgcm/xgcm/pull/380#discussion_r762286826).

xgcm/grid_ufunc.py
TomNicholas: This should be equivalent.

TomNicholas: (Although the only reason I made `grid` optional was so that you didn't have to specify it as the first unnamed arg, e.g.

`apply_as_grid_ufunc(da, grid=grid)` rather than `apply_as_grid_ufunc(grid, da)`...)

xgcm/test/test_grid_ufunc.py
```diff
@@ -441,6 +454,198 @@ def grad_to_inner(a):
...
+            coords={"depth_c": depth_c_coord},
+        )
+
+        result = multiply_left_right(grid, a, b, axis=[("depth",), ("depth",)])
+        assert_equal(result, expected)
+
+    def test_multiple_outputs(self):
+        def diff_center_to_inner(a, axis):
+            result = a - np.roll(a, shift=1, axis=axis)
+            return np.delete(result, 0, axis)  # remove first element along axis
```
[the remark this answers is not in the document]
TomNicholas: It would be more efficient to do something like `indexer = (slice(None),) * (axis-1) + (slice(0, -1),)`, but given that this is just a single test, it doesn't seem worth bothering.

Issue #210, opened 2020-08-29
"import xgcm" yields "AttributeError: 'NoneType' object has no attribute 'expandtabs'"


angilkaka:
Hello, 
when I do "import xgcm" in Python, I get the "AttributeError: 'NoneType' object has no attribute 'expandtabs'".
I tested several xgcm installations (conda, pip, pip+git) in newly created conda environments on Mac and Linux platforms. The error stays the same. 
One example:
python
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 23:03:20)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import xgcm
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/work/lib/python3.7/site-packages/xgcm/__init__.py", line 6, in <module>
    from .grid import Grid, Axis
  File "/work//lib/python3.7/site-packages/xgcm/grid.py", line 759, in <module>
    class Grid:
  File "/work/lib/python3.7/site-packages/xgcm/grid.py", line 1067, in Grid
    @docstrings.dedent
  File "/work/lib/python3.7/site-packages/docrep/decorators.py", line 26, in update_docstring
    doc = func(self, args[0].__doc__, *args[1:], **kwargs)
  File "/work/lib/python3.7/site-packages/docrep/__init__.py", line 517, in dedent
    s = inspect.cleandoc(s)
  File "/work/lib/python3.7/inspect.py", line 619, in cleandoc
    lines = doc.expandtabs().split('\n')
AttributeError: 'NoneType' object has no attribute 'expandtabs'

I also tested with an installation of pip install flake8 flake8-expandtab; still the same error. 

Could anyone help me importing xgcm working?

Thanks a lot! A.

jbusecke:
Hi @angilkaka, 

thanks for using xgcm and posting this issue. I have never encountered this. Could you post a complete list of installed conda packages in your environment for starters?

Could you try to set up a fresh conda environment with minimal dependencies, so we can narrow down if the problem is with xgcm itself?

@rabernat @dcherian do you have a clue what could be going on here?

rabernat:
> Could you post a complete list of installed conda packages in your environment for starters?

Perhaps we could explain how to do this? 😉 Not everyone will know.

> Could you try to set up a fresh conda environment with minimal dependencies, so we can narrow down if the problem is with xgcm itself?

Also this?

malmans2:
[...]

PR #384, opened 2021-12-01
Update docrep

### LINKED ISSUE #308: Docrep has GPL license
Hey @rabernat. Our legal team was recently reviewing our dependency stack in preparation for us open sourcing some of our code and we noticed that `docrep` has as a [GPL license](https://github.com/Chilipp/docrep/blob/master/LICENSE). Since this is dynamically imported here:
https://github.com/xgcm/xgcm/blob/34a77de7894a30cb80abcdd9c20571dbe54988ad/xgcm/grid.py#L10
This implies that xgcm is really available under a GPL license rather than the claimed MIT license. 

Since this import seems rather minor, would it possible to remove this runtime dependency on docrep? Currently it's not really possible for us to use xgcm. I think it's still okay of the docs need GPL code to be built, as long as the sdist on PyPI doesn't include GPL-derived stuff.

As an aside, we joined in on similar issues with necdf4 and cftime (see e.g. https://github.com/Unidata/netcdf4-python/issues/1073).

TomNicholas:
An attempt to fix #308.

This PR actually just completely removes the need for docrep at all - @jbusecke and I attempted to [update docrep](https://docrep.readthedocs.io/en/latest/#changed-in-0-3) to v0.3 but kept getting errors we didn't understand (in particular this one https://github.com/xgcm/xgcm/issues/210).

So we used the nuclear option to remove docrep by literally just copying and pasting the duplicated docstrings... :confused: 

That's annoying to have to do, but in fact we should probably wait until the [grid ufunc refactor](https://github.com/xgcm/xgcm/projects/10) is complete before trying to fix this situation/reinstate docrep, because that refactor will change how these docstrings all relate to one another anyway. Seems more important right now that our users don't have license problems (@nbren12).

 - [x] Closes #308

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -228,8 +228,6 @@ class Axis:
             summary.append(coord_info)
         return summary
 
-    @docstrings.get_sectionsf("neighbor_binary_func")
-    @docstrings.dedent
     def _neighbor_binary_func(
         self,
         da,
@@ -295,8 +293,6 @@ class Axis:
         da_new = self._wrap_and_replace_coords(da, data_new, to, keep_coords)
 
         return da_new
-
-    docstrings.delete_params("neighbor_binary_func.parameters", "f")
 
     def _neighbor_binary_func_raw(
         self,
@@ -575,7 +571,6 @@ class Axis:
 
         return left, right
 
-    @docstrings.dedent
     def interp(
         self,
         da,
@@ -592,8 +587,26 @@ class Axis:
 
         Parameters
         ----------
-        %(neighbor_binary_func.parameters.no_f)s
+        da : xarray.DataArray
+            The data on which to operate
+        to : {'center', 'left', 'right', 'inner', 'outer'}
+            The direction in which to shift the array. If not specified,
+            default will be used.
+        boundary : {None, 'fill', 'extend'}
+            A flag indicating how to handle boundaries:
 
+            * None:  Do not apply any boundary conditions. Raise an error if
+              boundary conditions are required for the operation.
+            * 'fill':  Set values outside the array boundary to fill_value
+              (i.e. a Dirichlet boundary condition.)
+            * 'extend': Set values outside the array to the nearest array
+              value. (i.e. a limited form of Neumann boundary condition.)
+        fill_value : float, optional
+            The value to use in the boundary condition with `boundary='fill'`.
+        vector_partner : dict, optional
+            A single key (string), value (DataArray)
+        keep_coords : boolean, optional
+            Preserves compatible coordinates. False by default.
 
         Returns
         -------
@@ -613,7 +626,6 @@ class Axis:
             keep_coords=keep_coords,
         )
 
-    @docstrings.dedent
     def diff(
         self,
         da,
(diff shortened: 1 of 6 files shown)


xgcm/grid.py
```diff
@@ -1772,16 +1989,43 @@ def interp_2d_vector(self, vector, **kwargs):
...
+            * 'fill':  Set values outside the array boundary to fill_value
+              (i.e. a Dirichlet boundary condition.)
+            * 'extend': Set values outside the array to the nearest array
+              value. (i.e. a limited form of Neumann boundary condition.)
+
+            Optionally a dict with separate values for each axis can be passed (see example)
+        fill_value : {float, dict}, optional
+            The value to use in the boundary condition with `boundary='fill'`.
+            Optionally a dict with seperate values for each axis can be passed (see example)
+        vector_partner : dict, optional
```
jbusecke, line 2020: I believe the 'vector_partner' input is only valid for the methods ending in `_vector`, and should be removed from all others.

xgcm/grid.py
```diff
@@ -592,8 +583,26 @@ def interp(
...
 
+            * None:  Do not apply any boundary conditions. Raise an error if
+              boundary conditions are required for the operation.
+            * 'fill':  Set values outside the array boundary to fill_value
+              (i.e. a Dirichlet boundary condition.)
+            * 'extend': Set values outside the array to the nearest array
+              value. (i.e. a limited form of Neumann boundary condition.)
+        fill_value : float, optional
+            The value to use in the boundary condition with `boundary='fill'`.
+        vector_partner : dict, optional
```
jbusecke, line 602: remove (see comment below)

xgcm/grid.py
```diff
@@ -1822,17 +2065,44 @@ def integrate(self, da, axis, **kwargs):
...
-            Name of the axis on which to act
-        %(grid_func.parameters)s
+        axis : str or list or tuple
+            Name of the axis on which to act. Multiple axes can be passed as list or
+            tuple (e.g. ``['X', 'Y']``). Functions will be executed over each axis in the
+            given order.
+        to : str or dict, optional
+            The direction in which to shift the array (can be ['center','left','right','inner','outer']).
+            If not specified, default will be used.
+            Optionally a dict with seperate values for each axis can be passed (see example)
```
jbusecke, line 2082: The `(see example)` does not apply?

jbusecke (changes requested):
Just minor stuff in addition to removing `vector_partner` from most of the docstrings. Otherwise I support this nuclear option for now.

xgcm/grid.py
jbusecke: Sorry this is actually false. This only applies to certain methods, namely: `cumint` and `cumsum`.

jbusecke:
Could you rephrase

```
metric_weighted : str or tuple of str or dict, optional
            If an axis or list of axes is specified,
            the appropriate grid metrics will be used to determined the weight for interpolation.
            E.g. if passing `metric_weighted=['X', 'Y']`, values will be weighted by horizontal area.
            If `False` (default), the points will be weighted equally.
            Optionally a dict with seperate values for each axis can be passed (see example)
```

to 

```
metric_weighted : str or tuple of str or dict, optional
            Optionally use metrics to multiply/divide with appropriate metrics before/after the operation.
            E.g. if passing `metric_weighted=['X', 'Y']`, values will be weighted by horizontal area.
            If `False` (default), the points will be weighted equally.
            Optionally a dict with seperate values for each axis can be passed.
```

Otherwise this looks good to me. Thanks a lot @TomNicholas.

jbusecke (approved):

rabernat:
Thanks so much for taking care of this chore Tom! 👏

nbren12:
Thanks! Doing God's work here!

PR #387, opened 2021-12-03
Enabled mypy in pre-commit config

TomNicholas:
@rabernat I think you merged #386 a bit pre-emptively - I've been running mypy just on `grid_ufunc.py`, and when I run on the whole code on the `master` branch I get a couple of typing errors that I don't understand:

```
xgcm/duck_array_ops.py:29: error: All conditional function variants must have identical signatures
doc/conf.py:244: error: Need type annotation for "latex_elements" (hint: "latex_elements: Dict[<type>, <type>] = ...")
Found 2 errors in 2 files (checked 19 source files)
```

I copied the actual pre-commit config from xarray, removing some datetime stubs that seemed irrelevant.

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -14,7 +14,7 @@ from .grid_ufunc import GridUFunc, _signatures_equivalent, apply_as_grid_ufunc
 from .metrics import iterate_axis_combinations
 
 try:
-    import numba
+    import numba  # type: ignore
 
     from .transform import conservative_interpolation, linear_interpolation
 except ImportError:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -5,7 +5,6 @@ import operator
 import warnings
 from collections import OrderedDict
 
-import docrep
 import numpy as np
 import xarray as xr
 
@@ -22,8 +21,6 @@ except ImportError:
     numba = None
 
 from typing import Dict, Optional, Union
-
-docstrings = docrep.DocstringProcessor(doc_key="My doc string")
 
 
 def _maybe_promote_str_to_list(a):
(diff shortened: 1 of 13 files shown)


xgcm/grid.py
```diff
@@ -595,8 +586,26 @@ def interp(
 
         Parameters
         ----------
-        %(neighbor_binary_func.parameters.no_f)s
+        da : xarray.DataArray
```
jbusecke, line 589: Wasnt this all part of another PR already? I am not quite sure why these show up as changes here.

TomNicholas: No I don't really understand that either :confused:

TomNicholas:
I fixed the type errors - this should be ready to merge now.

xgcm/grid.py
jbusecke: Can you try to force resolve the merge to `theirs`? @rabernat has done that once to resolve an issue like this. https://stackoverflow.com/a/21777677

jbusecke (changes requested):
Looks good, but I would really like to keep the history clean and not add all these old commits here.

xgcm/grid.py
TomNicholas: I tried `git pull -X theirs` but that literally didn't do anything...

Then I tried rebasing this PR onto `grid_ufunc_refactor` but that hasn't got rid of the docstring changes :angry:

TomNicholas: It seems now at least I have all the commits I actually want to comprise this PR in a row at the end of the history... Wonder if that means I could easily make a new PR and cherry-pick the all in one go?

TomNicholas: Closing this in favour of #413 - cherry-picking to create a new PR did work but if anyone could explain to me what the correct thing to do to clean up the history of this PR would have been I would appreciate it!

PR #394, opened 2021-12-10
Refactor grid.interp, min, max

TomNicholas:
Extends the grid_ufunc refactor to replace other grid functions aside from just `grid.diff`.

Builds on #380


TomNicholas:
`Grid.diff`, `Grid.interp`, `Grid.min`, `Grid.max` now all work through the new `apply_as_grid_ufunc` code path, but `Grid.cumsum` is a bit more complicated.

Currently xgcm does this
```python
# first use xarray's cumsum method
da_cum = da.cumsum(dim=dim)

boundary_kwargs = dict(boundary=boundary, fill_value=fill_value)

# now pad / trim the data as necessary
# here we enumerate all the valid possible shifts
if (pos == "center" and to == "right") or (pos == "left" and to == "center"):
    # do nothing, this is the default for how cumsum works
    data = da_cum.data
elif (pos == "center" and to == "left") or (pos == "right" and to == "center"):
    data = _pad_array(
        da_cum.isel(**{dim: slice(0, -1)}), dim, left=True, **boundary_kwargs
    )
```
So I tried this
```python
@as_grid_ufunc(signature="(X:center)->(X:left)", boundary_width={"X": (1, 0)})
def cumsum_center_to_left(a):
    summed = np.cumsum(a, axis=-1)
    return summed[..., :-1]
```
But this gives an AssertionError when we compare the results.

My new result starts like this
```
array([[ 0.90961497,  1.81922994,  2.5864991 , ...
```
Whereas the old version starts like this
```
array([[ 0.90961497,  0.90961497,  1.67688413, ...
```
There seems to be a difference in that the old way sums, then trims, then pads

Whereas the new ufunc way has to pad first, then sum, then trim...

TomNicholas:
Maybe I'm just being silly and cumsum shouldn't go through a grid ufunc at all...

rabernat:
I am pretty confident that cumsum as currently implemented does the thing we want it to do. You're correct that, with cumsum, the boundary conditions are applied _after_ cumsum is called. This has to do with the fact that cumsum is basically an integral operation, while diff is a differential operation. I think of cumsum as the inverse of diff. So to me it makes sense that cumsum does things in the opposite order.

Another complication that would eventually arise is how to handle chunked core dims. Your grid_ufunc implementation would not work with chunked data because the cumsum would start over from 0 at each chunk. This is how dask implements cumsum: https://docs.dask.org/en/latest/_modules/dask/array/reductions.html#cumsum

If it seems complicated to shoehorn cumsum into the grid_ufunc framework, I am fine with leaving the implementation as is for now.

TomNicholas:
We accidentally merged this prematurely during a hack - it actually currently fails a bunch of tests which do padding via extrapolation.

This means there must have been zero tests which did diff + extrapolation, but a bunch of tests which do {interp/min/max} + extrapolate.

We didn't catch the fact that this PR was failing because the CI didn't run properly + we though the crosses were just a non-important doctest failure :man_facepalming:

Now we should fix this by making a new PR which xfails the tests which call extrapolate, until we can close that by actually implementing extrapolate within our new padding function.

PR #393, opened 2021-12-08
Metric aware ufuncs

TomNicholas:
Makes `metric_weighted` kwarg work for new grid_ufunc refactor. Builds on #380

 - [ ] Closes #xxxx
 - [ ] Tests added
 - [ ] Passes `pre-commit run --all-files`
 - [ ] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
 - [ ] New functions/methods are listed in `api.rst`

PR #396, opened 2021-12-15
xfail extrapolation tests

TomNicholas:
See [this comment](https://github.com/xgcm/xgcm/pull/394#issuecomment-995208660)


TomNicholas:
This also xfails a load of tests which need `metric_weighted` to work - they should be un-xfailed after #393 is merged.

PR #395, opened 2021-12-15
start moving padding to standalone module

rabernat:
- [ ] Closes #xxxx
 - [ ] Tests added
 - [x] Passes `pre-commit run --all-files`
 - [ ] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
 - [ ] New functions/methods are listed in `api.rst`

PR #399, opened 2021-12-16
Refactor padding

jbusecke:
This is a duplicate of #395. I tried to make a PR to that PR, but ended up with a huge diff, so I just opted to start a new PR. Hope that is ok @rabernat 
 
- [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1043,13 +1043,6 @@ class Axis:
         return da.get_axis_num(coord_name)
 
 
-_XGCM_BOUNDARY_KWARG_TO_XARRAY_PAD_KWARG = {
-    "periodic": "wrap",
-    "fill": "constant",
-    "extend": "edge",
-}
-
-
 class Grid:
     """
     An object with multiple :class:`xgcm.Axis` objects representing different
@@ -1567,85 +1560,6 @@ class Grid:
                 out = out / metric_new
 
         return out
-
-    def pad(self, *arrays, boundary_width=None, boundary=None, fill_value=None):
-        """
-        Pads the boundary of given arrays along given Axes, according to information in Axes.boundary.
-
-        Parameters
-        ----------
-        arrays : Sequence[xarray.DataArray]
-            Arrays to pad according to boundary and boundary_width.
-        boundary_width : Dict[str: Tuple[int, int]
-            The widths of the boundaries at the edge of each array.
-            Supplied in a mapping of the form {axis_name: (lower_width, upper_width)}.
-        boundary : {None, 'fill', 'extend', 'extrapolate', dict}, optional
-            A flag indicating how to handle boundaries:
-            * None: Do not apply any boundary conditions. Raise an error if
-              boundary conditions are required for the operation.
-            * 'fill':  Set values outside the array boundary to fill_value
-              (i.e. a Dirichlet boundary condition.)
-            * 'extend': Set values outside the array to the nearest array
-              value. (i.e. a limited form of Neumann boundary condition.)
-            * 'extrapolate': Set values by extrapolating linearly from the two
-              points nearest to the edge
-            Optionally a dict mapping axis name to separate values for each axis
-            can be passed.
-        fill_value : {float, dict}, optional
-            The value to use in boundary conditions with `boundary='fill'`.
-            Optionally a dict mapping axis name to separate values for each axis
-            can be passed. Default is 0.
-        """
-        # TODO accept a general padding function like numpy.pad does as an argument to boundary
-
-        if not boundary_width:
-            raise ValueError("Must provide the widths of the boundaries")
-
-        if boundary and isinstance(boundary, str):
-            boundary = {ax_name: boundary for ax_name in self.axes.keys()}
-        if fill_value is None:
-            fill_value = 0.0
-        if isinstance(fill_value, float):
-            fill_value = {ax_name: fill_value for ax_name in self.axes.keys()}
-
-        padded = []
-        for da in arrays:
-            new_da = da
-            for ax, widths in boundary_width.items():
-                axis = self.axes[ax]
-                _, dim = axis._get_position_name(da)
-
-                # Use default boundary for axis unless overridden
-                if boundary:
-                    ax_boundary = boundary[ax]
-                else:
-                    ax_boundary = axis.boundary
-
-                if ax_boundary == "extrapolate":
-                    # TODO implement extrapolation
-                    raise NotImplementedError
-                elif ax_boundary is None:
-                    # TODO this is necessary, but also seems inconsistent with the docstring, which says that None = "no boundary condition"
-                    ax_boundary = "periodic"
-
-                # TODO avoid repeatedly calling xarray pad
-                try:
-                    mode = _XGCM_BOUNDARY_KWARG_TO_XARRAY_PAD_KWARG[ax_boundary]
-                except KeyError:
-                    raise ValueError(
-                        f"{ax_boundary} is not a supported type of boundary"
-                    )
-
-                if mode == "constant":
-                    new_da = new_da.pad(
-                        {dim: widths}, mode, constant_values=fill_value[ax]
-                    )
-                else:
-                    new_da = new_da.pad({dim: widths}, mode)
-
-            padded.append(new_da)
-
-        return padded
 
     def _1d_grid_ufunc_dispatch(
         self, funcname, da, axis, to=None, keep_coords=False, **kwargs
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -34,6 +34,11 @@ def _maybe_promote_str_to_list(a):
 
 
 _VALID_BOUNDARY = [None, "fill", "extend", "extrapolate"]
+_VALID_AXIS_PROPERTIES = [
+    "periodic",
+    "boundary",
+    "fill_value",
+]
 
 
 class Axis:
@@ -73,7 +78,7 @@ class Axis:
         ds,
         axis_name,
         periodic=True,
-        default_shifts={},
+        default_shifts={},  # should this be named `to` for simplicity?
         coords=None,
         boundary=None,
         fill_value=None,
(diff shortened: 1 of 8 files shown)


jbusecke:
As a first step I have added tests for `pad` with all the methods that we currently support.

jbusecke:
Ok I have rewritten quite a bit, but the foundation stands. I got `test_diff_interp_connected_grid_x_to_x` and `test_diff_interp_cubed_sphere` to pass. 

`test_vector_diff_interp_connected_grid_x_to_y` still is giving me trouble (I assume this is a numpy/xarray transpose issue).

I skipped `test_get_periodic_grid_edge` for now. I will try to refactor this into the `test_padding.py` module, since this is really a padding issue. 

Still very rough around the edges, but feedback on the general structure would be cool.

jbusecke:
Huh, seems like I broke something. Ill check on that tomorrow.

jbusecke:
Ok I have tried to really get into the face connection logic and would like to lay out my current thinking, so that maybe somebody can point out flaws or at least I have a reference for after the break. 

Figuring out which face is connected to what other face on which axis is relatively easy. The hard part is that for certain connections, the 'added' array slice needs to be flipped. There are two different flippings that might be required (see graphic below): 
1) Orthogonal flip: The array slice has to be flipped orthogonal to the axis that is is going to be padded on. 
2) Parallel flip: This is a new flip, only stricly required when extracting more than one row. This assures that for `reverse` connections, the order of rows is as if they would have been moved one after another.

After many sketches I think I have identified 4 relevant cases of face connections that reveal a general pattern:
Blank diagram

here shown for a `padding_width=(0,2)`. I think that the decision tree covers all the cases of face connections we need to make.

> Note that the current code only covers the `is axis swapped` case? I think that if the connection is reversed and the axes are swapped, we actually do not need to execute an orthogonal flip. 

Any thoughts?

TomNicholas:
Obvious question - is it possible to have a connection which is both "reverse" and "same axis"? i.e. a 4th case?

jbusecke:
I was thinking about the same, in the simple case of two faces it would end up being some Möbius strip sort of thing? I honestly think this is not used in practice. But maybe @rabernat has some more insight here? Do we need to support that case?

TomNicholas:
> I think that is the 3rd case in the graphic, no?

Yes, but I was just wondering because it's not in your flow chart.

jbusecke:
Quick update. I think I have made some significant progress here. I have added tests for each of the scenarios above and they currently pass. The failures in `test_exchange.py` can be traced back to a problem with setting the default method of padding, not the padding method itself (at least I confirmed it for `test_diff_interp_connected_grid_x_to_x`). I will look into that in a bit. 
As of now the tests are quite long, but I hope they are somewhat understandable. 

Happy to explain my logic of padding in more detail.

I would love to get some feedback on the general structure of the padding module (in particular the passing of kwargs, which @rabernat was not to fond of in the beginning?).

jbusecke:
> The failures in `test_exchange.py` can be traced back to a problem with setting the default method of padding, not the padding method itself (at least I confirmed it for `test_diff_interp_connected_grid_x_to_x`). 

I just noticed that `test_boundary_kwarg_same_as_grid_constructor_kwarg` is failing which supports this idea.

xgcm/padding.py
```diff
@@ -0,0 +1,295 @@
...
+    fill_value :
+        The value to use in boundary conditions with `boundary='fill'`.
+        Optionally a dict mapping axis name to separate values for each axis
+        can be passed. Default is 0.
+    """
+    # TODO accept a general padding function like numpy.pad does as an argument to boundary
+
+    # TODO: boundary width should not really be optional here I think?
+    if not boundary_width:
+        raise ValueError("Must provide the widths of the boundaries")
```
TomNicholas, line 277: Presumably it never makes sense to pad with zero-width...

xgcm/padding.py
```diff
@@ -0,0 +1,295 @@
...
+    # Brute force approach for now. If any axis has connections use
+    # face connection method, this dispatches the simple pad stuff internally if needed.
+    if any([any(grid.axes[ax]._connections.keys()) for ax in grid.axes]):
+        da_padded = _pad_face_connections(*padding_args, padding=padding, **kwargs)
+    else:
+        # TODO: we need to have a better detection and checking here. For now just pipe everything else into the
+        # legacy cases
+
+        # Legacy cases
+        da_padded = _pad_basic(*padding_args, padding=padding, **kwargs)
```
TomNicholas, line 293: Sorry what exactly are the legacy cases that your code doesn't handle?

xgcm/padding.py
```diff
@@ -0,0 +1,295 @@
...
+        da = da.rename({to_name: to_name + "dummy"})
+        if from_name in da.dims:
+            da = da.rename({from_name: to_name})
+        da = da.rename({to_name + "dummy": from_name})
+    else:
+        da = da.rename({from_name: to_name})
+    return da
+
+
+def _pad_face_connections(da, grid, padding_width, **kwargs):
```
TomNicholas, line 33: This function deserves its own docstring

xgcm/padding.py
TomNicholas: So the legacy cases are with no face connections, handled in the way you and I did it using `xarray.pad`?

TomNicholas (commented):
Nice! Obviously this is a WIP but the general approach looks okay to me so far, at least with my limited understanding of the face connections.

jbusecke:
Thanks for all the comments @TomNicholas. I will try to work them in tomorrow. In the meantime I think I have all of the exchange tests working (with some caveats below). I had to do some quite substantial refactoring and would like to discuss some aspects.

- I converted `axis._periodic` to a public attribute `axis.periodic`, this is more in line with how `axis.boundary` and `axis.fill_value` are stored. This will be removed in the future, so that should not be that big of a deal? cc @rabernat
- I added a rather hacky workaround in `_1d_grid_ufunc_dispatch`, that basically converts the `periodic` input to an argument for `boundary` @TomNicholas would you be ok with this until we have phased out `periodic`?
- I refactored the way we set up axes in `Grid.__init__`. I think this is much more readable.
- I added the private method `_check_axis_kwarg_mapping` which checks input arguments for defaults set on the axes, and parses them in if the argument was not explicitly set. We can use this logic in many different places, to make sure that our defaults are properly set (which is right now an absolute nightmare to track down!).
- I added a new test case for the situation @TomNicholas mentioned above and have parametrized the tests a bit more compactly.

I have not yet tackled a situation where vector partners are involved. I think given the volume of this PR, we should tackle this in a separate PR?

I was getting some really weird flaky test failures in the main `test/test_grid.py` module. Ill see if they persist in the CI.

xgcm/grid.py
```diff
@@ -73,7 +78,7 @@ def __init__(
         ds,
         axis_name,
         periodic=True,
-        default_shifts={},
+        default_shifts={},  # should this be named `to` for simplicity?
```
jbusecke, line 81: General question I had. But totally minor

xgcm/grid.py
```diff
@@ -1150,44 +1157,39 @@ def __init__(
...
-            if isinstance(fill_value, dict):
-                axis_fillvalue = fill_value.get(axis_name, None)
-            elif isinstance(fill_value, (int, float)) or fill_value is None:
-                axis_fillvalue = fill_value
-            else:
-                raise ValueError(
-                    f"fill_value={fill_value} is invalid. Please specify a dictionary "
-                    "mapping axis name to a boundary option; a number or None."
-                )
+        # TODO: Check the provided input to be valid?
```
jbusecke, line 1172: Still have to think about a way to do this elegantly.

xgcm/padding.py
jbusecke: Good point! It is quite a beast at this point. Ill see if I can refactor it more once this matures.

xgcm/padding.py
jbusecke: Agreed. I wanted to make this a positional argument in the first place. Do you think I should just go ahead with that?

xgcm/padding.py
jbusecke: That is maybe a bad naming. These are all the 'basic' cases we implemented before (e.g. everything xarray can handle). Does that make sense? The comments here were more for myself, thinking forward when we implement other options.

xgcm/test/test_exchange.py
```diff
@@ -130,6 +130,7 @@ def test_create_periodic_grid(ds):
         assert connect_right[2] is False
 
 
+@pytest.mark.skip(reason="We shouldnt test private functions")
```
jbusecke, line 133: This will not work by design (the private function should be removed as part of this PR), so how do we best deal with this? I think skipping it for now might be good?

xgcm/test/test_exchange.py
```diff
@@ -167,12 +168,20 @@ def test_create_connected_grid(ds, ds_face_connections_x_to_x):
...
+    with pytest.raises(KeyError):
+        Grid(ds, face_connections=ds_face_connections_x_to_x)
+
+
 def test_diff_interp_connected_grid_x_to_x(ds, ds_face_connections_x_to_x):
     # simplest scenario with one face connection
-    grid = Grid(ds, face_connections=ds_face_connections_x_to_x)
-    diff_x = grid.diff(ds.data_c, "X", boundary="fill")
-    interp_x = grid.interp(ds.data_c, "X", boundary="fill")
+    grid = Grid(ds, face_connections=ds_face_connections_x_to_x, periodic=False)
```
jbusecke, line 180: @rabernat I am very confused here. The original [test](https://github.com/xgcm/xgcm/blob/2ecc59cbd0af9ac1d21eee9c2b18a639cf0e02b4/xgcm/test/test_exchange.py#L170-L186) did not specify `periodic=False`. Was the expected behavior that setting boundary in a grid method overwrites periodic?

Eventually we will not have this ambiguity anymore, when we move periodic as an argument to `boundary`/`padding`, but I wanted to make this PR as backwards compatible as possible.

xgcm/test/test_grid.py
```diff
@@ -104,7 +104,7 @@ def _assert_axes_equal(ax1, ax2):
     for pos, coord in ax1.coords.items():
         assert pos in ax2.coords
         assert coord == ax2.coords[pos]
-    assert ax1._periodic == ax2._periodic
+    assert ax1.periodic == ax2.periodic
```
jbusecke, line 107: I wonder if this change is small and obvious enough to justify me changing tests? @TomNicholas, do you have opinions?

jbusecke:
Closing in favor of #408/

PR #404, opened 2022-01-06
Migration of boundary axis properties for refactor

### LINKED ISSUE #403: Internal Changes to check for axis property values
@TomNicholas and I had a bit of a frustrating session on xgcm today. We discovered that there is common logic needed to properly check for axis defaults of `periodic`, `boundary` and `fill_value` and also migrate values from `periodic` into `boundary`.
It is important to note that these changes will eventually not be needed once `boundary` is replaced with `padding` (issue?).
The logic implemented in #399 caused some really bonkers test failures. So instead of lumping these changes into either PR, we will try to make a new PR that only addresses this specific problem.

Next Steps: 
1. Start a new PR on top of `grid_ufunc_refactor_project`
2. Move `boundary_global_input` to grid tests (This is not testing anything with the metrics) and remove xfails.
3. Reimplement the keyword parsing and checking the axis defaults (#393 and #399 provide a blueprint).
4. Pay special attention to the flaky test behavior we observed in #399
5. Rewrite both #393 and #399 to make them more focussed.

jbusecke:
- [x] Closes #403 

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1150,12 +1150,38 @@ class Grid:
                         f"Input `{dim}` (for the `{pos}` position on axis `{axis}`) is not a dimension in the input datasets `ds`."
                     )
 
+        # Convert all inputs to axes-kwarg mappings
+        # Parse axis properties
+        boundary = self._as_axis_kwarg_mapping(boundary, axes=all_axes)
+        fill_value = self._as_axis_kwarg_mapping(fill_value, axes=all_axes)
+
+        # this case does only apply to periodic. Since we plan on deprecating it soon
+        # handle it here
+        if isinstance(periodic, list):
+            periodic = {axname: True for axname in periodic}
+        periodic = self._as_axis_kwarg_mapping(periodic, axes=all_axes)
+
+        # Set properties on
+
+        # Populate axes. Much of this is just for backward compatibility. I think we are better of
+        # getting completely rid of the axis object and storing/getting info from the grid object properties
+        # (all of which are axis-value mapping).
+        self.boundary = boundary
+        self.fill_value = fill_value
+        self.periodic = periodic
+        # TODO we probably want to properly define these as class properties with setter/getter?
+
         self.axes = OrderedDict()
         for axis_name in all_axes:
-            try:
-                is_periodic = axis_name in periodic
-            except TypeError:
-                is_periodic = periodic
+            # try:
+            #     is_periodic = axis_name in periodic
+            # except TypeError:
+            #     is_periodic = periodic
+
+            # setting this properly is still needed for the transform axis method
+            # TODO: migrate that over to grid
+            is_periodic = periodic.get(axis_name, False)
+
             if axis_name in default_shifts:
                 axis_default_shifts = default_shifts[axis_name]
             else:
@@ -1170,7 +1196,7 @@ class Grid:
                     f"boundary={boundary} is invalid. Please specify a dictionary "
                     "mapping axis name to a boundary option; a string or None."
                 )
-
+            # TODO Reimplement value checking on grid properties.
             if isinstance(fill_value, dict):
                 axis_fillvalue = fill_value.get(axis_name, None)
             elif isinstance(fill_value, (int, float)) or fill_value is None:
@@ -1200,18 +1226,23 @@ class Grid:
             for key, value in metrics.items():
                 self.set_metrics(key, value)
 
-    def _parse_axes_kwargs(self, kwargs):
-        """Convvert kwarg input into dict for each available axis
+        # Finish setup
+
+    def _as_axis_kwarg_mapping(self, kwargs, axes=None):
+        """Convert kwarg input into dict for each available axis
         E.g. for a grid with 2 axes for the keyword argument `periodic`
         periodic = True --> periodic = {'X': True, 'Y':True}
         or if not all axes are provided, the other axes will be parsed as defaults (None)
         periodic = {'X':True} --> periodic={'X': True, 'Y':None}
         """
+        if axes is None:
+            axes = self.axes
+
         parsed_kwargs = dict()
         if isinstance(kwargs, dict):
             parsed_kwargs = kwargs
         else:
-            for axis in self.axes:
+            for axis in axes:
                 parsed_kwargs[axis] = kwargs
         return parsed_kwargs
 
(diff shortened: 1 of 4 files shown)


xgcm/grid.py
```diff
@@ -1150,12 +1150,38 @@ def __init__(
...
+        # handle it here
+        if isinstance(periodic, list):
+            periodic = {axname: True for axname in periodic}
+        periodic = self._as_axis_kwarg_mapping(periodic, axes=all_axes)
+
+        # Set properties on
+
+        # Populate axes. Much of this is just for backward compatibility. I think we are better of
+        # getting completely rid of the axis object and storing/getting info from the grid object properties
+        # (all of which are axis-value mapping).
```
TomNicholas, line 1168: This is where type hinting the attributes of the `Grid` class would be helpful

xgcm/grid.py
```diff
@@ -1150,12 +1150,38 @@ def __init__(
...
+        self.fill_value = fill_value
+        self.periodic = periodic
+        # TODO we probably want to properly define these as class properties with setter/getter?
+
         self.axes = OrderedDict()
         for axis_name in all_axes:
-            try:
-                is_periodic = axis_name in periodic
-            except TypeError:
-                is_periodic = periodic
```
TomNicholas, line 1158: I'm pleased you got rid of this catching TypeError pattern

xgcm/grid.py
```diff
@@ -1200,18 +1226,23 @@ def __init__(
             for key, value in metrics.items():
                 self.set_metrics(key, value)
 
-    def _parse_axes_kwargs(self, kwargs):
-        """Convvert kwarg input into dict for each available axis
+        # Finish setup
+
+    def _as_axis_kwarg_mapping(self, kwargs, axes=None):
+        """Convert kwarg input into dict for each available axis
```
TomNicholas, line 1232: Another place where type hinting would be very nice

xgcm/grid.py
```diff
@@ -1667,10 +1700,53 @@ def _1d_grid_ufunc_dispatch(
...
+        kwargs = {
+            k: self._as_axis_kwarg_mapping(v) if k in VALID_BOUNDARY_KWARGS else v
+            for k, v in kwargs.items()
+        }
+        # Now check if some of the values not provided in the kwargs are set as grid properties
+        # In the future this should be the only way to access that information!
+        kwargs_with_defaults = {}
+
+        for k in np.unique(
+            list(kwargs.keys()) + VALID_BOUNDARY_KWARGS
```
TomNicholas, line 1724: Can't you use `set()` instead of `np.unique()`?

xgcm/grid.py
jbusecke: Eventually we will be able to remove that whole block below too. For now I opted for the least intrusive change.

xgcm/grid.py
jbusecke: Um I guess. Let me try.

xgcm/grid.py
jbusecke: Getting right on it...

jbusecke: does `self` get a type hint?

xgcm/grid.py
jbusecke: I am not quite sure I follow. I see the advantage of doing that, but what exactly is ambiguous here?

xgcm/grid.py
jbusecke: I added type hints, but am very unsure if I did it correctly. Would you mind checking?

xgcm/grid.py
jbusecke: Yup seems to work.

xgcm/test/test_grid.py
```diff
@@ -662,10 +662,13 @@ def test_grid_ops(all_datasets):
             for boundary in bcs:
                 da_interp = grid.interp(ds[varname], axis_name, boundary=boundary)
                 da_interp_ax = axis.interp(ds[varname], boundary=boundary)
+                print((da_interp - da_interp_ax).sum())
```
TomNicholas, line 665:
```suggestion
```

xgcm/test/test_grid.py
```diff
@@ -870,3 +873,54 @@ def test_input_dim_notfound():
...
+    "fill_value",
+    [0, 10, None],
+)
+def test_boundary_global_input(funcname, boundary, fill_value):
+    """Test that globally defined boundary values result in
+    the same output as when the parameters are defined the grid methods
+    """
+    ds, coords, metrics = datasets_grid_metric("C")
+    axis = "X"
+    print(axis)
```
TomNicholas, line 904:
```suggestion
```

jbusecke: Oops

xgcm/grid.py
```diff
@@ -1200,18 +1221,27 @@ def __init__(
...
 
-    def _parse_axes_kwargs(self, kwargs):
-        """Convvert kwarg input into dict for each available axis
+        # Finish setup
+
+    def _as_axis_kwarg_mapping(
+        self,
+        kwargs: Union[str, float, int, Dict[str, Union[str, float, int]]],
+        axes: Optional[list] = None,
+    ):
```
TomNicholas, line 1230: You can specify a return type like this
```suggestion
    ) -> Dict[str, Union[str, float, int]]:
```

xgcm/grid.py
TomNicholas: Thanks - using set is much more pythonic, less error-prone, and I expect a lot quicker.

xgcm/grid.py
jbusecke: Ah yes, totally forgot that. Thanks.

xgcm/grid.py
```diff
@@ -1150,12 +1151,32 @@ def __init__(
...
+            periodic = {axname: True for axname in periodic}
+        periodic = self._as_axis_kwarg_mapping(periodic, axes=all_axes)
+
+        # Set properties on grid object. I think we are better of
+        # getting completely rid of the axis object and storing/getting info from the grid object properties
+        # (all of which need to be supplied/converted to axis-value mapping).
+        self.boundary = boundary
+        self.fill_value = fill_value
+        self.periodic = periodic
+        # TODO we probably want to properly define these as class properties with setter/getter?
```
TomNicholas, line 1172: This would be the other good place to add type hints

https://stackoverflow.com/questions/40173982/how-can-i-type-hint-an-attribute-in-python-3-5

xgcm/grid.py
TomNicholas: `self` doesn't need one (unless your method returns self later, which yours doesn't).

Your type hints look okay to me! Just need to add a return value to this function.

xgcm/grid.py
TomNicholas: If you had type hints for the grid object's class attributes then the comment `# (all of which are axis-value mapping).` would be superfluous, because that info would be specified by the type hint.

jbusecke:
Ok I think this is ready to go. Merge when the CI passes?

TomNicholas (approved):
Looks good to me! Thanks for adding the type hints.

PR #407, opened 2022-01-06
Enable metric weighted ufuncs

jbusecke:
- [x] Replaces #393, built on top of #404
 - [x] Passes `pre-commit run --all-files`

Credit: @TomNicholas who wrote #393, this is merely a quick reimplementation on top of #393

Issue #406, opened 2022-01-06
Should Grid/Axis kwargs be read-only properties instead of attributes?


TomNicholas:
Currently we have important pieces of information about the grid just stored as basic python class attributes, e.g. `Grid.metrics`, or `Axis._periodic`. Do we want these to be read-only? i.e. are we comfortable with users being able to edit features of their grids whenever they like, or should they have to create an entirely new grid object if they want to change anything?

Either way we should promote these to properties. If we don't want people to edit them, then properties are by default read-only. If we do want people to be able to edit them, then we can use property's setting decorator thing to check that the users new choice of kwarg is actually valid.

This is related to #403 , #405, and the idea to deprecate the `periodic` kwarg.

jbusecke:
In some cases (e.g. metrics) it is crucial that these can be modified! 

> Either way we should promote these to properties. If we don't want people to edit them, then properties are by default read-only. If we do want people to be able to edit them, then we can use property's setting decorator thing to check that the users new choice of kwarg is actually valid.

I like this! I think it would be great if we can remove the various instances of checking/setting defaults/promoting to dict from everywhere other than the Grid init. Basically the init + setter methods would take care that each one of these properties is 'complete' with regard to the axes in the grid object. All other logic then assumes that e.g. boundary is always given as a dict.

jbusecke:
Just FYI I believe we have two archetypes of 'properties':
1. Basically keyword arguments like `boundary`, `fill_value`... These will be organized in a dictionary where a single axis serves as key.
2. 'array' dictionaries, like metrics (but masks #293 could be organized similarly). These dictionaries store a list of arrays as values with **either a single axis or a combination of axes as key**. The key indicates the orientation of the arrays in space, while the arrays can cover different axis positions (and combinations thereof) along that orientation.

rabernat:
We may wish to consider refactoring these classes to be frozen Dataclasses, which would solve a lot of problems simultaneously.

Python dataclasses are extremely useful. If they had existed when we started this package, it would have been a lot simpler.

jbusecke:
> We may wish to consider refactoring these classes to be frozen Dataclasses, which would solve a lot of problems simultaneously.

What problems are you thinking about?

I think this would be great for the properties of case 1 above, but just to reiterate, metrics need to be modified for a lot of the newest functionality.

TomNicholas:
[...]

Issue #405, opened 2022-01-06
Lets get rid of `Axis`


jbusecke:
I propose that as part of our current substantial refactor, we eliminate the `Axis` object altogether!

What we have to gain:
- 1200+ lines of code we do not have to maintain anymore. 🤯

What we loose: 
- Pretty much nothing! @TomNicholas work will make the axis methods obsolete, and the the two remaining aspects (storing axis attributes + `transform`) can comfortably live in the grid object IMO. 
    - The axis attributes can be stored just as easy as dictionaries `{axisname:value, ...}` in the grid object directly (actually certain things like #404 are easier this way). Similarly in #399 I actually ended up storing the `face_connections` on the grid.
    - I am confident that the transform functionality is not to hard to migrate, since it is pretty separate from everything else.

What do people think? @rabernat did I miss something here?

PR #408, opened 2022-01-06
Refactor padding logic into module

### LINKED ISSUE #403: Internal Changes to check for axis property values

jbusecke:
- [x] Closes #403 
 - [x] Replaces #399 built on top of #404 and #407

If anyone is curious, some of the graphics in #399 might be interesting to follow this logic.

### DIFF
--- a/xgcm/padding.py
+++ b/xgcm/padding.py
@@ -30,7 +30,26 @@ def _maybe_swap_dimension_names(da, from_name, to_name):
     return da
 
 
-def _pad_face_connections(da, grid, padding_width, **kwargs):
+def _pad_face_connections(
+    da, grid, padding_width, padding, other_component, fill_value
+):
+
+    # i made some alterations to the grid init to store the
+    # following info on the grid level. We can discuss if that makes more sense
+    # TODO: Do we need that information on the axis still?
+    facedim = grid._facedim
+    connections = grid._connections
+
+    if isinstance(da, dict):
+        assert len(da) == 1  # TODO raise a better error here
+        isvector = True
+        vectoraxis, da = da.popitem()
+    else:
+        isvector = False
+
+    if isvector:
+        assert other_component is not None  # TODO raise a better error here
+        vectorpartneraxis, da_partner = other_component.popitem()
 
     # Can I get rid of all the mess with the coordinates by stripping them here?
     # The output of the padding will not fit any coordinates anymore anyways and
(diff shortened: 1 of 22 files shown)


jbusecke:
So this PR has one caveat for now. If you look at the test [here](https://github.com/xgcm/xgcm/blob/317f35131f8c4784e28947abf616b286c3b8fe35/xgcm/test/test_exchange.py#L173) I was not able to get this to pass without adding `periodic=False`. 

This indicates to me that some default is not set right. @rabernat could you perhaps explain what the expected padding behavior for a `periodic` axis with face connection should be? I might have to add one more test for this case.
This code from [`Axis`](https://github.com/xgcm/xgcm/blob/2a90c55cef600e2916960dae42c81bfaa00c6ace/xgcm/grid.py#L203-L206) seems relevant:
```python
        # now we implement periodic coordinates by setting appropriate
        # connections
        if periodic:
            self._connections = {None: ((None, self, False), (None, self, False))}
```
I am not sure I understand what is happening here. The test seems to expect to get zeros on the non-connected edge of the face (which works if we set `periodic=False`, which then pads with the default value of 0) but somehow the old logic did that even though periodic was true?

rabernat:
Currently we create a "default" set of connections for each Axis

https://github.com/xgcm/xgcm/blob/2a90c55cef600e2916960dae42c81bfaa00c6ace/xgcm/grid.py#L201-L206

The `if periodic` block creates a face connection spec that mimics a periodic boundary condition. However, **I don't remember if this is actually used anywhere!**

This is the current test in master

https://github.com/xgcm/xgcm/blob/2a90c55cef600e2916960dae42c81bfaa00c6ace/xgcm/test/test_exchange.py#L170-L172

vs your branch

https://github.com/xgcm/xgcm/blob/317f35131f8c4784e28947abf616b286c3b8fe35/xgcm/test/test_exchange.py#L171-L173

I can't understand why there is any difference at all. If `face_connections` is specified to the `Grid` constructor, the default connections for each Axis will always be overwritten

https://github.com/xgcm/xgcm/blob/317f35131f8c4784e28947abf616b286c3b8fe35/xgcm/grid.py#L1324-L1326

My best guess is that this is the result of the fact that we can create a periodic boundary in two different ways:
1. By passing `periodic=True` to the axis
2. By creating a periodic face connection

This is because periodic boundary conditions are just one special case of the more general topology supported by face connections.

I suspect we can resolve the problem by searching for places where we explicitly check for `axis.periodic` and do special actions in that case, which we do NOT do for method 2.

In general, I would not adhere strictly to the existing tests; they are possibly inconsistent / incorrect, particularly around face connections. However, I would strive to eliminate this redundancy and have only _one correct internal state_ for periodic boundary conditions.

jbusecke:
> I can't understand why there is any difference at all. If face_connections is specified to the Grid constructor, the default connections for each Axis will always be overwritten

Neither can I. From the expected results in the test, periodic should have never been true! 

> If face_connections is specified to the Grid constructor, the default connections for each Axis will always be overwritten

This is not currently the case in my PR. Let me try to summarize the process. I am basically padding all faces with the given `boundary` before applying the face connections. These then get applied afterwards, which has the advantage that any faces which have no connections are already padded and can easily be attached with possible padding perpendicular to the face. This should work fine in all cases except `periodic`. I believe what my current code would do there is wrap the other side of each **face** (which makes no sense!).

More generally though, should we not simply forbid having both `boundary='periodic'` and `face_connections` set? Either we have a setup with a single face and then providing face connections is mostly not needed (but we could still do it and as such emulate periodic) or you have a multiface setup and then periodic is not well defined (or am I missing something here ?). Either way, one should never provide `face_connections` and `boundary='periodic' from how I see it here?

EDIT: I guess we could actually check for periodic and whether the face connections can be modified to accommodate this (the current behavior), but this seems a bit convoluted to me. This would NOT solve the current issue with the test though!

>I suspect we can resolve the problem by searching for places where we explicitly check for axis.periodic and do special actions in that case, which we do NOT do for method 2.

I would like to strive towards #405 and with that we could check in one single place that we do not set 1 AND 2 in any scenario. This would also satisfy this here:
>However, I would strive to eliminate this redundancy and have only one correct internal state for periodic boundary conditions.

My suggestion then would be that we leave the modification of the test in this PR and add 2 more test:
- One that tests that periodic boundary conditions via `face_connections` are equivalent to `boundary='periodic'`.
- Test that an error is raised when both are specified?

jbusecke:
I think I have a conceptual solution to the vector_partner logic (also extending it to tangential vectors and reverse connections). But that will be a bit more work (and some discussion needed). For the sake of moving this along, I would suggest to make this a separate PR. The vector partner logic is currently still going through `Axis`, but all tests are passing now.

My current idea is to do something along these lines:
1. Merge the ufunc refactor into the main branch
2. Work on another branch, which will eventually eliminate `Axis` (#405).
    - This would include a revised vector_partner logic, #406 and maybe others? 

What do you think @TomNicholas, @rabernat ?

TomNicholas:
> Merge the ufunc refactor into the main branch
>
> Work on another branch, which will eventually eliminate Axis

Presumably we should still redirect the `Axis` methods through the new ufunc code before we remove Axis later? Otherwise we're never going to have the checkpoint of passing all the old tests using only the new code...

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+
+def _pad_face_connections(da, grid, padding_width, **kwargs):
+
+    # Can I get rid of all the mess with the coordinates by stripping them here?
+    # The output of the padding will not fit any coordinates anymore anyways and
+    # after the ufunc is applied they will all be back on place. So lets make our
+    # lives easier here.
+    da = da.reset_coords(drop=True).reset_index(
+        [di for di in da.dims if di in da.coords], drop=True
+    )
```
TomNicholas, line 41: Is this a good idea? What if the `da` has a coordinate that doesn't depend on the dimension of the axis along which we are applying the ufunc? In that case `xarray.apply_ufunc` should keep it, and if you drop all coordinates here it would change that behaviour. (Are you dropping all coordinates? I can't easily understand what this line does...)

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def _expand_boundary_width(padding_width):
+        all_widths = []
+        for widths in padding_width.values():
+            all_widths.extend(list(widths))
+        max_width = max(all_widths)
+        expanded_padding_width = {
+            k: (max_width, max_width) for k in padding_width.keys()
+        }
+        return expanded_padding_width
```
TomNicholas, line 58: Can you give a simple example of why this causes a problem?

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+
+    da_prepadded = _pad_basic(da, grid, padding_width, **kwargs)
+
+    pad_axes = padding_width.keys()
+
+    # i made some alterations to the grid init to store the
+    # following info on the grid level. We can discuss if that makes more sense
+    # TODO: Do we need that information on the axis still?
+    facedim = grid._facedim
+    connections = grid._connections
```
TomNicholas, line 81: If it can be stored in the `Axis` class it should, and then `Grid` just looks it up from the `Axis` objects it contains.

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # TODO: I will have to modify this when using vector partners?
```
TomNicholas, line 83: Definitely leave vector logic for a follow-up PR

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+    n_facedim = len(da[facedim])
+    da_split = []
+
+    # Iterate over each face and pad accordingly
+    for i in range(n_facedim):
```
TomNicholas, line 88: Could we refactor the contents of this loop by using some kind of `_pad_axis(da, axis, vector_partner=None)` function?

xgcm/padding.py
```diff
@@ -0,0 +1,306 @@
...
+            da = da.isel({dim: slice(start, stop)})
+        return da
+
+    return _trim_expanded_padding_width(
+        da_padded, grid, padding_width_original, padding_width
+    )
+
+
+def _pad_basic(da, grid, padding_width, padding="fill", fill_value=None):
+    """Implement basic xarray/numpy padding methods"""
```
TomNicholas, line 206: Ideally we would eventually remove this, otherwise we're going to end up with two different code paths that can do the same thing.

TomNicholas: Also we it would be a good idea to look at how `xarray.pad`/`dask.pad` works internally in case we can improve dask performance somehow.

TomNicholas:
Is this still relevant or was it superceded by https://github.com/xgcm/xgcm/pull/459 ?

jbusecke:
I believe this is completely superceded by #459. Closing now.

Issue #410, opened 2022-01-10
Extended functionality for vector data with complex grid topologies


jbusecke:
As part of #408 I have been thinking a lot about face connections and the associated padding of arrays.

Here, I would like to propose two things:

1) An expansion of xgcm's capabilities to apply any operation on both orthogonal and tangential vector components, and include `reverse` face connections.
2) A simplified API which makes the use of helper methods like `interp_2d_vector` obsolete.

# 1) Expanded padding functionality
## Introduction

In the following I will assume that a dataset with face connections is given by an array of 2+ dimensional data that is stacked along an additional `face` dimension. Vector components on each face indicate the axis directionality of a quantity on each face.
![Blank diagram-2](https://user-images.githubusercontent.com/14314623/148831997-7e4076fa-1c3d-491a-9153-54b1d7c3f8fe.png)

To operate on such a dataset we need to enable 'exchanges' between faces, which are given by [face connections](https://xgcm.readthedocs.io/en/latest/grid_topology.html#cubed-sphere). In practice this means we have to pad various sides of each face with the appropriate data from another face. Operations can then be applied to each face, and after trimming the padding result in the same shape output, but the operations respect the connectivity of the faces.

### Methodology

In practice, this means that we iterate through each face (target), and then again iterate through all target faces, identify the data used for padding (source), modify this and concatenate it with the target.

As I pointed out previously for tracer fields, there are only 4 general cases that need to be considered:
![Blank diagram-4](https://user-images.githubusercontent.com/14314623/148833278-70c415b5-3db5-4111-bcc2-8b5398865e54.png)

For scalar data we only need to worry about flipping the source data appropriately before padding. We either flip the data along the axis of concatenation (orthogonal flip) or along the 'other' axis (tangential flip).

#### Considerations for vector components
This logic is however not enough to accommodate vector data. For vectors data, two additional things need to be taken into account. If the connection is switching axes, the vector components need to be exchanged (e.g. pad values of `v` of one face to the `u` data of another). Secondly, in certain cases the sign of the components needs to be changed. 
I think I have captured the relevant cases for orthogonal and tangential vector components with the blue arrows in this picture:

![Blank diagram-5](https://user-images.githubusercontent.com/14314623/148834379-c13ea67c-dfbd-4dee-865b-a37f144612f5.png)

This gives a clear pattern: The sign of tangential vectors has to be changed if a tangential flip is performed, and the sign of orthogonal vectors has to be changed if an orthogonal flip is performed.

[...]

PR #411, opened 2022-01-10
Deprecation warnings

### LINKED ISSUE #409: Deprecation warnings
@rabernat @jbusecke and I think we should add pre-emptive warnings for various API changes that we want to ship alongside the grid ufunc refactor. We will push users warnings for these changes into master soon so that by the time we merge all the refactor stuff users have at least been warned a bit:

- Rename boundary padding kwarg
- `periodic` kwarg is being dropped
- `Axis` is going to go entirely in favour of always using grid
- `fill_value` default being changed from zero to NaN
- extrapolate functionality might get dropped
- (at some point) change `interp_2d_vector` API to be more intuitive

jbusecke:
- [x] Closes #409 

- [x] Rename boundary padding kwarg
- [x] `periodic` kwarg is being dropped
- [x] `Axis` is going to go entirely in favour of always using grid
- [x] `fill_value` default being changed from zero to NaN
- [x] extrapolate functionality might get dropped
- [x] (at some point) change `interp_2d_vector` API to be more intuitive

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1210,6 +1210,16 @@ class Grid:
         self._ds = ds
         self._check_dims = check_dims
 
+        # Deprecation Warnigns
+        if boundary:
+            warnings.warn(
+                "The `boundary` argument will be renamed "
+                "to `padding` to better reflect the process "
+                "of array padding and avoid confusion with "
+                "physical boundary conditions (e.g. ocean land boundary).",
+                category=DeprecationWarning,
+            )
+
         if coords:
             all_axes = coords.keys()
         else:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1235,6 +1235,25 @@ class Grid:
                 category=DeprecationWarning,
             )
 
+        if fill_value:
+            warnings.warn(
+                "The default fill_value will be changed to nan (from 0.0 previously) "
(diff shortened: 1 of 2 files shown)


xgcm/grid.py
```diff
@@ -1908,6 +1952,12 @@ def _apply_vector_function(self, function, vector, **kwargs):
         # the keys, should be axis names
         assert len(vector) == 2
 
+        warnings.warn(
+            "`interp_2d_vector` and `diff_2d_vector` will be removed from future releases."
+            "The same functionality will be available under the `xgcm.Grid` methods.",
+            category=DeprecationWarning,
```
TomNicholas, line 1958: We could hopefully update this with a recommendation once we decide on what the vector logic API will look like?

TomNicholas (approved):
Seems okay to me

xgcm/grid.py
jbusecke: Yes, but I think that will depend on #410, no? Do you suggest to add a note in the code?

TomNicholas: I suppose it's like a chicken and egg problem.

jbusecke: Lets just go with this now, and release today, so that we get the max warning time? Does that sound ok for you?

PR #412, opened 2022-01-10
Enable mamba in CI

### LINKED ISSUE #398: Build CI environments with mamba
Building out CI environments takes forever. We should switch the GH actions over to mamba.

jbusecke:
- [x] Closes #398


jbusecke:
Seems like this is actually not really faster? How?

PR #413, opened 2022-01-10
Mypy pre commit attempt 2

TomNicholas:
Replaces #387 because I couldn't figure out a better way to clean the git history :grimacing:

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -21,7 +21,7 @@ try:
 except ImportError:
     numba = None
 
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Iterable, List, Union
 
 docstrings = docrep.DocstringProcessor(doc_key="My doc string")
 
@@ -1225,9 +1225,9 @@ class Grid:
 
     def _as_axis_kwarg_mapping(
         self,
-        kwargs: Union[str, float, int, Dict[str, Union[str, float, int]]],
-        axes: Optional[list] = None,
-    ) -> Dict[str, Union[str, float, int]]:
+        kwargs: Union[Any, Dict[str, Any]],
+        axes: List[str] = None,
+    ) -> Dict[str, Any]:
         """Convert kwarg input into dict for each available axis
         E.g. for a grid with 2 axes for the keyword argument `periodic`
         periodic = True --> periodic = {'X': True, 'Y':True}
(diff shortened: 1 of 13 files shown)


xgcm/grid.py
```diff
@@ -1225,9 +1225,9 @@ def __init__(
 
     def _as_axis_kwarg_mapping(
         self,
-        kwargs: Union[str, float, int, Dict[str, Union[str, float, int]]],
-        axes: Optional[list] = None,
-    ) -> Dict[str, Union[str, float, int]]:
+        kwargs: Union[Any, Dict[str, Any]],
+        axes: List[str] = None,
```
jbusecke, line 1229: I guess this could be a tuple too? I was never clear on that. But it might be good if we try to enforce one one type here?

xgcm/grid.py
```diff
@@ -1709,7 +1709,7 @@ def _1d_grid_ufunc_dispatch(
         if isinstance(axis, str):
             axis = [axis]
 
-        # convert inpug arguments into axes-kwarg mappings
```
jbusecke, line 1712: Lol, dont you like pugs?

jbusecke (approved):
Looks great (from someone who doesnt understand that much about type hinting) 😃

jbusecke:
Oh dont forget to add an entry to `doc/whats-new.rst`!

TomNicholas:
Yeah I remember what's new just now, it's already in https://github.com/xgcm/xgcm/pull/413/commits/9059e54766e9fa5c961ad06ae274a9effbf7b18a

xgcm/grid.py
TomNicholas: What could be a tuple? The list? In which case we should put Iterable there instead.

TomNicholas:
I think we need to add the equivalent of [this line](https://github.com/pydata/xarray/blob/69dc2a0b76346b4f02c4e9423f3532c35025607a/setup.cfg#L82) to get the upstream CI to pass, or alternatively just stop supporting python 3.7.

jbusecke:
Not quite sure I understand this. According to [this here](https://github.com/xgcm/xgcm/blob/f099a82afa7977785c62190b4b256840a208fe48/.github/workflows/ci.yaml#L114) we are using python 3.8 in the upstream testing? Would the problem be solved by changing that to 3.9 or maybe even 3.10?

TomNicholas:
Fixed it! Was being dumb and hadn't added `typing_extensions` to the CI environment. I don't really understand at what version we no longer need that import, but it's not a big deal imo.

PR #415, opened 2022-01-11
Remove Axis from public API

### LINKED ISSUE #405: Lets get rid of `Axis`

TomNicholas:
Removing huge amounts of code that the refactor makes redundant! Early stage so far but wanted to push so that it was backed up somewhere other than my laptop.

We decided not to remove the `Axis` class completely, as the abstraction is still useful internally, but we did remove it from public API. That means we were free to remove all the Axis computation methods.

Also we tried changing `Axis` to a `dataclass` but found it was more trouble than it was worth.

 - [x] Will close #405, and #406
 - [ ] Tests added


xgcm/grid.py
```diff
@@ -37,106 +40,42 @@ def _maybe_promote_str_to_list(a):
...
-
-         Outer
-         |------o-------|------o-------|------o-------|------o-------|
-        [0]            [1]            [2]            [3]            [4]
-
-    The `center` position is the only one without the `c_grid_axis_shift`
-    attribute, which must be present for the other four. However, the actual
-    value of `c_grid_axis_shift` is ignored for `inner` and `outer`, which are
-    differentiated by their length.
-    """
```
TomNicholas, line 70: Probably want to find a better place to preserve the nice explanation in this docstring.

jbusecke:
Oh yeahhhh. I am excited for this. Ill take a deeper look soon.

jbusecke:
Will this be superseeded by #557 entirely? @TomNicholas

jbusecke:
Closing as superseeded by #557

PR #417, opened 2022-02-03
Grid information through type annotations

TomNicholas:
- [x] Implements the [suggestion for specifying variable grid information in type hints](https://github.com/xgcm/xgcm/issues/344#issuecomment-1020334503)
 - [x] Tests added


TomNicholas:
Can someone *please* help me understand why I'm getting such enormous `git diff`s when I've only committed like twice since branching off `grid_ufunc_refactor_project`?! (Which I've confirmed is the same locally and here)

setup.cfg
```diff
@@ -15,6 +15,8 @@ ignore =
     W503
     E265
     F811
+    # Allows type hinting as Gridded[DataArray, "(X:center)"], where we did `from typing import Annotated as Gridded`
+    F722
```
TomNicholas, line 19: So I wanted to get type hinting to look like this
```python
@as_grid_ufunc
def f(a: Gridded[xr.DataArray, "X:center"]):
   ...
```
instead of this
```python
@as_grid_ufunc
def f(a: Annotated[xr.DataArray, "X:center"]):
   ...
```
but after much messing about I think the only way to achieve that whilst retaining `mypy` comptaibility is the hacky trick of doing
```python
from typing import Annotated as Gridded
```
As far as I can tell this is perfectly safe, but it does then mean I need to silence this flake8 warning.

TomNicholas:
Closed in favour of #418 418

PR #418, opened 2022-02-03
Grid information through type annotations

TomNicholas:
(Replacement for #417 without the massive diff)

 - [x] Implements the [suggestion for specifying variable grid information in type hints](https://github.com/xgcm/xgcm/issues/344#issuecomment-1020334503)
 - [x] Tests added

We have to upgrade to python 3.9 to use `typing.Annotated`.


setup.cfg
```diff
@@ -15,6 +15,8 @@ ignore =
     W503
     E265
     F811
+    # Allows type hinting as Gridded[DataArray, "(X:center)"], where we did `from typing import Annotated as Gridded`
+    F722
```
TomNicholas, line 19: So I wanted to get type hinting to look like this
```python
@as_grid_ufunc
def f(a: Gridded[xr.DataArray, "X:center"]):
   ...
```
instead of this
```python
@as_grid_ufunc
def f(a: Annotated[xr.DataArray, "X:center"]):
   ...
```
but after much messing about I think the only way to achieve that whilst retaining `mypy` compatibility is the hacky trick of doing
```python
from typing import Annotated as Gridded
```
As far as I can tell this is perfectly safe, but it does then mean I need to silence this flake8 warning.

I think using `Gridded` is clearer for users, but maybe others think this is unnecessary?

xgcm/test/test_grid_ufunc.py
```diff
@@ -81,6 +83,100 @@ def test_invalid_signatures(self, signature):
...
+            ...
+
+        assert ufunc.signature == "()->()"
+
+    def test_annotated_args(self):
+        # TODO test hints without annotations
+        # TODO test hints with annotations that don't conform to Xgcm
+
+        @as_grid_ufunc()
+        def ufunc(a: Gridded[xr.DataArray, "X:center"]):
```
TomNicholas, line 99: I need to try and change this so that we [type hint with numpy arrays instead](https://github.com/xgcm/xgcm/issues/344#issuecomment-1020344225), i.e.
```python
Gridded[np.ndarray, "X:center"]
```
(or maybe even `ArrayLike` which I saw somewhere...)

xgcm/test/test_grid_ufunc.py
```diff
@@ -678,22 +813,28 @@ class GridOpsMockUp:
     """
 
     @staticmethod
-    @as_grid_ufunc(signature="(X:center)->(X:left)")
-    def diff_center_to_left(a):
```
TomNicholas, line 682: Do we actually want to use type hinting here? Or in our `gridops.py` module? It's arguably harder to read quickly for the simple functions we are currently storing in there.

xgcm/test/test_grid_ufunc.py
```diff
@@ -445,8 +572,13 @@ def grad_to_inner(a):
...
         # Test decorator
-        @as_grid_ufunc("(X:center,Y:center)->(X:inner,Y:center),(X:center,Y:inner)")
-        def grad_to_inner(a):
+        @as_grid_ufunc()
+        def grad_to_inner(
+            a: Gridded[xr.DataArray, "X:center,Y:center"]
+        ) -> Tuple[
+            Gridded[xr.DataArray, "X:inner,Y:center"],
+            Gridded[xr.DataArray, "X:center,Y:inner"],
+        ]:
```
TomNicholas, line 581: This is a good example of what the syntax looks like for a somewhat complicated function signature.

TomNicholas:
I have a weird mypy/pre-commit problem, where if I do `pre-commit run --all-files` then mypy passes, but once I actually try to commit anything I get mypy errors:
```
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...........................................(no files to check)Skipped
isort....................................................................Passed
black....................................................................Passed
blackdoc.................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

xgcm/grid_ufunc.py:6: error: Module "typing" has no attribute "Annotated"
xgcm/grid_ufunc.py:202: error: Unexpected keyword argument "include_extras" for "get_type_hints"
Found 2 errors in 1 file (checked 1 source file)
```
To me those errors look like something to do with the version of python or `typing_extensions`, because `Annotated` and `include_extras` were only added in python 3.9. However (a) I thought I updated all that and (b) I don't understand how pre-commit can lead to different results in these two cases. For the moment I'm just forcing all the commits through with `--no-verify`.

xgcm/test/test_grid_ufunc.py
```diff
@@ -81,6 +83,112 @@ def test_invalid_signatures(self, signature):
...
+        # This should fail mypy
+        # TODO but it doesn't
+        @as_grid_ufunc()
+        def ufunc(a: Gridded[str, "(X:center)"]):
+            assert isinstance(a, np.ndarray)
+
+        # This should pass mypy
+        @as_grid_ufunc()
+        def ufunc(a: Gridded[np.ndarray, "(X:center)"]):
+            assert isinstance(a, np.ndarray)
```
TomNicholas, line 189: I don't understand why the first of these two doesn't fail `mypy`. This means right now you can put literally anything in the annotated type hint and mypy won't complain, which is not what is supposed to happen.

xgcm/test/test_grid_ufunc.py
```diff
@@ -81,6 +83,112 @@ def test_invalid_signatures(self, signature):
...
+        # TODO I want this to fail mypy but it doesn't
+        @as_grid_ufunc()
+        def ufunc1(a: Gridded[str, "(X:center)"]):
+            ...
+
+        # This should pass mypy
+        @as_grid_ufunc()
+        def ufunc3(a: Gridded[np.ndarray, "(X:center)"]):
+            # np.ndarray has a .strides method but str doesn't (and nor does xr.DataArray)
+            print(a.strides)
```
TomNicholas, line 189: What do I need to do to get this to fail mypy? Is it because [`xarray.apply_ufunc` allows its arguments to be `Any`](https://github.com/pydata/xarray/blob/33067cd24f66d4855babaa6801b009480c4e2cb2/xarray/core/computation.py#L804)?

TomNicholas: I tried narrowing the `Any` upstream, but I think my debugging is being thwarted by pre-commit behaving unexpectedly again:

```
(py39) tom@pop-os:~/Documents/Work/Code/xgcm$ mypy xgcm
xgcm/grid_ufunc.py:399: note: Revealed type is "builtins.tuple[xarray.core.dataarray.DataArray]"
xgcm/grid_ufunc.py:554: note: Revealed type is "builtins.list[xarray.core.dataarray.DataArray]"
(py39) tom@pop-os:~/Documents/Work/Code/xgcm$ pre-commit run --all-files
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...............................................................Passed
isort....................................................................Passed
black....................................................................Passed
blackdoc.................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

xgcm/grid_ufunc.py:399:5: F821 undefined name 'reveal_type'
xgcm/grid_ufunc.py:554:5: F821 undefined name 'reveal_type'

mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

xgcm/grid_ufunc.py:399: note: Revealed type is "builtins.tuple[Any]"
xgcm/grid_ufunc.py:554: note: Revealed type is "builtins.list[Any]"
```

xgcm/test/test_grid_ufunc.py
```diff
@@ -81,6 +83,112 @@ def test_invalid_signatures(self, signature):
             _parse_grid_ufunc_signature(signature)
 
 
+class TestGetSignatureFromTypeHints:
+    def test_no_args_to_annotate(self):
+        @as_grid_ufunc()
+        def ufunc():
+            ...
+
+        assert ufunc.signature == "()->()"
```
TomNicholas, line 92: I'm not really sure if I'm somehow hiding some bugs by doing this. In normal numpy ufuncs doesn't `()` mean a scalar? Whereas here I'm using it to mean "no argument". I don't really test these cases but there might be bugs lurking here.

xgcm/test/test_grid_ufunc.py
TomNicholas: @dcherian do you have any idea what I would need to change to get `mypy` to understand that 

```python
@as_grid_ufunc()
def ufunc1(a: Annotated[str, "(X:center)"]):
    ...
```
should raise a typing error? (Because `str` is not `np.ndarray`)

jbusecke (commented):
This is super cool! I dont have enough experience with type hinting in general so I would punt on that. 
A bit torn about making 3.9 minimum, but then again, if people install this new version there are going to be a bunch of breaking changes, so maybe backwards compatibility is not that big of a concern.

TomNicholas:
> A bit torn about making 3.9 minimum

Xarray is already on 3.8 minimum, and they are fairly conservative, so we're hardly pushing all our users into a brave new world!

jbusecke:
> Xarray is already on 3.8 minimum, and they are fairly conservative, so we're hardly pushing all our users into a brave new world!

Sure, then l have nothing really against it. @rabernat ?

TomNicholas:
> backwards compatibility

Also in general I personally don't really think of version changes as "breaking backwards compatibility". At the most updating our minimum version requires the user only to update their conda environment, whereas when I think of "breaking backwards compatibility" I imagine changes that actually require users to update their actual production code. Obviously no changes at all would be best, but one is less of a big deal than the other imo.

jbusecke:
Fair point.

xgcm/test/test_grid_ufunc.py
dcherian: Sorry no idea.

setup.py
```diff
@@ -48,14 +48,12 @@
...
         "Operating System :: OS Independent",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
     ],
     packages=find_packages(exclude=["docs", "tests", "tests.*", "docs.*"]),
     install_requires=install_requires,
     extras_require=extras_require,
-    python_requires=">=3.7",
+    python_requires=">=3.9",
```
TomNicholas, line 56: I think the docs build CI environment should also potentially be updated to python 3.9...

xgcm/grid_ufunc.py
```diff
@@ -104,23 +100,84 @@ def _parse_grid_ufunc_signature(
...
+        input_arg_signature = ",".join(
+            [
+                f"({variable_annotation})"
+                for variable_annotation in arg_annotations.values()
+            ]
+        )
+    else:
+        input_arg_signature = "()"
+
+    return f"{input_arg_signature}->{return_arg_signature}"
```
rabernat, line 145: Rather than parsing the type hints to a str input signature, could not not go directly to a `Signature` object? Seems a little inefficient to use the signature formatted string as our internal serialization format of the signature.

xgcm/grid_ufunc.py
```diff
@@ -104,23 +100,84 @@ def _parse_grid_ufunc_signature(
...
+    in_ax_names: T_AX_POS_LIST
+    out_ax_names: T_AX_POS_LIST
+    in_ax_positions: T_AX_POS_LIST
+    out_ax_positions: T_AX_POS_LIST
 
     _REPLACEMENT_DUMMY_INDEX_NAMES = [f"__{char}" for char in string.ascii_letters]
 
     def __init__(self, signature: str):
-        self.signature: str = signature
+        self._parse(signature)
```
rabernat, line 158: Implementing the above suggestion would basically involve refactoring this class to allow a more direct construction path, e.g. 

```python
    def __init__(self, in_ax_names, out_ax_names, in_ax_positions, out_ax_positions)
```

and then adding another class method for `from_signature_string`.

xgcm/test/test_grid_ufunc.py
```diff
@@ -81,6 +83,106 @@ def test_invalid_signatures(self, signature):
...
+            @as_grid_ufunc(signature="(X:center)->(X:left)")
+            def ufunc(
+                a: Gridded[np.ndarray, "X:center"]
+            ) -> Gridded[np.ndarray, "X:left"]:
+                ...
+
+    def test_type_hint_as_numpy_ndarray(self):
+        # TODO I want this to fail mypy but it doesn't
+        @as_grid_ufunc()
+        def ufunc1(a: Gridded[str, "X:center"]):
```
rabernat, line 176: We should turn this into a MRE and open a mypy issue.

rabernat (commented):
Great work here Tom.

One comment about the implementation and the internal representation of the signature.

xgcm/grid_ufunc.py
TomNicholas: Yep can do - the design as it is might be slightly weird just because I actually wrote this PR before I implemented the `Signature` class in #441 . You're right that it makes sense to build strings from the internal representation though - I'll change it.

xgcm/test/test_grid_ufunc.py
TomNicholas: I think we decided that this case is irrelevant because no conceivable grid ufunc would have zero inputs or zero outputs.

xgcm/grid_ufunc.py
TomNicholas: I understand what you mean, and I agree. I was referring to (really thinking aloud to myself) about how if I do that then the `__str__` method of `Signature` should rebuild the string from the internal representation. Let me just have a go at it and I think you'll see we are on the same page.

xgcm/test/test_grid_ufunc.py
TomNicholas: I've now explicitly forbidden creating signatures which don't have either any input arguments or any output arguments (although I'm still not 100% in my head that there are no cases where you would want that). I think now there is a little inconsistency/inefficiency because the overall signature regex allows for zero or more arguments, when it should now allow for one or more.

xgcm/grid_ufunc.py
TomNicholas: Okay I've rewritten it so the constructors follow the pattern you specified @rabernat

xgcm/test/test_grid_ufunc.py
TomNicholas: Changing `Gridded` to `Annotated` appears to have fixed this.

xgcm/grid_ufunc.py
TomNicholas: I've done this so I'm just going to merge this PR.

PR #421, opened 2022-02-04
Fix docs

TomNicholas:
Attempt to fix the problem that is causing all our docs CI runs to fail on the refactor branch


TomNicholas:
It worked! @jbusecke

Issue #419, opened 2022-02-03
Should dask be optional?


TomNicholas:
In theory dask is meant to be optional for xgcm, but in practice we don't actually test that it works without dask installed in the CI.

Some of my changes in the grid ufunc refactor currently break it if dask isn't installed, but I think that if we decide we definitely do want it to work without dask then it's not a major fix.

Q is just: Do we want to bother with a no-dask option? (If so we should add a new CI env for it)

jbusecke:
I think a non dask CI env would be a good idea, but I am not sure how much refactoring is needed in the tests? If you say its not a big deal, I support it.

Probably we actually need to modify the legacy tests to use dask at all?

VeckoTheGecko:
Should also be quite straightforward with the work that https://github.com/lucascolley has done on the array API ecosystem in Python.

Reference:
- [Pydata Paris talk: Lucas Colley - A Hitchhiker's Guide to the Array API Standard Ecosystem](https://www.youtube.com/watch?v=LBVzkCSmjsQ) ([slides](https://lucascolley.github.io/talks/pydata-paris-25-array-api/#/6/2))
- [Array API compat library](https://data-apis.org/array-api-compat/)


Becomes easier with #683 as we can easily create envs without dask

I think I can work on this (and also take over/supercede #422 ) at some point

TomNicholas:
I don't think that Lucas' work extends to chunked arrays. Computing on those requires using additional primitives such as `map_overlap` or `apply_gufunc`. But regardless I think this could be made to work fairly easily in xGCM.

PR #422, opened 2022-02-04
Test suite without dask

TomNicholas:
Removes dask from the "core dependencies" CI environment file, so that no we have a CI run that tests everything without dask installed.

I altered an existing run rather than added a new run because this seemed to better fit the intention of "just core dependencies".

I don't expect this to pass - but we can push to this branch to start fixing any bugs that this reveals.

 - [x] Part of closing #419

### DIFF
--- a/ci/environment-core-deps.yml
+++ b/ci/environment-core-deps.yml
@@ -5,7 +5,6 @@ channels:
   - conda-forge
 dependencies:
   - codecov
-  - dask
   - docrep<=0.2.7
   - numpy
   - scipy


TomNicholas:
Wait why didn't the CI run the code tests??

jbusecke:
Umm that is weird...no idea why this wouldnt work...

jbusecke:
Do you envision this being part of 0.7.0? I would say this is probably more suited for the next round where we modify tests substantially?

TomNicholas:
I think this is low-priority, and can wait until we have finalised the internals based on other criteria.

jbusecke:
Not sure this should be closed. Could be part of #466

TomNicholas:
The only reason this was closed is because when I tried deleting the `grid_ufunc_refactor_project` branch it auto-closed every PR which built off that branch.

PR #425, opened 2022-02-04
Update environment-upstream-dev.yml

jbusecke:
Hotfix for test failure in the [upstream dev CI](https://github.com/xgcm/xgcm/runs/4768058407?check_suite_focus=true)


jbusecke:
Huh? This is soo weird, why is the code formatter failing on a completely unrelated line?

jbusecke:
I do not understand that formatting failure, but I have fixed it now to move on. Since it is not anything pertaining to the actual code, the risk here seems low.

Issue #443, opened 2022-02-25
[BUG] Can't apply grid ufunc with padding if axis position doesn't change


TomNicholas:
I realised that I overlooked quite an important case: currently if you try to apply a grid ufunc which pads along an axis but still returns a variable which has that same `axis:position` pair as the input, `xr.apply_ufunc` will throw an error. For example whilst this works:

```python
@as_grid_ufunc("(X:center)->(X:left)", boundary_width={"X": (1, 0)})
def diff_center_to_left(a):
    return a[..., 1:] - a[..., :-1]
```

this does not:

```python
@as_grid_ufunc("(X:center)->(X:center)", boundary_width={"X": (2, 0)})
def diff_center_to_center_second_order(a):
    return 0.5 * (a[..., 2:] - a[..., :-2])
```

The latter throws an error like

```
ValueError: size of dimension 'depth_c' on inputs was unexpectedly changed by applied function from 11 to 9. 
Only dimensions specified in ``exclude_dims`` with xarray.apply_ufunc are allowed to change size.
```

The reason is that the dimensions of the output variable are set by the grid positions in the signature - in the first case we pad a dimension to be longer, then act on it with `xr.apply_ufunc` and want the output to be shorter _along a different dimension_, whereas in the second case we pad a dimension to be longer, then act on it with `xr.apply_ufunc` and want the output to be shorter _along that same dimension_.

There are potentially two ways to fix this:
1) Do something clever with renaming dimensions to have temporary names whilst inside `xr.apply_ufunc`, so xarray doesn't realise there is a conflict in sizes,
2) Move the padding inside the ufunc as described in #436.

--- 

This bug is currently what's stopping me from computing vorticity of LLC data @rabernat FYI, because the simplest grid ufunc for that looks like this (see #436)

```python
def diff(a, axis):
    def _diff(a):
        return a[..., 1:] - a[..., :-1]
    
    return np.apply_along_axis(_diff, axis, a)


@as_grid_ufunc("(lon:left,lat:center),(lon:center,lat:left)->(lon:left,lat:left)",
               boundary_width={"lon": (1, 0), "lat": (1, 0)})
def vort(u, v):
    """This needs to return an array 1 element smaller along both axis -1 & -2."""
    u_trimmed = u[..., 1:]
    v_trimmed = v[..., 1:, :]
    
    return (
        diff(v_trimmed, axis=-1)
        - diff(u_trimmed, axis=-2)
    )
```
[...]

PR #447, opened 2022-03-07
Padding inside ufunc

### LINKED ISSUE #436: Fusing padding tasks with ufunc tasks
@rabernat pointed out that in the grid ufunc refactor, we are currently doing padding with xarray and then applying a numpy operation using `xarray.apply_ufunc` afterwards. This works, but it will never be the approach that minimizes the total number of dask tasks.

Instead what we should do is compose the padding function with the numpy ufunc, to create a new ufunc which boths pads and operates. Then we apply that composed function using apply ufunc. That guarantees that we only use one task per chunk. (In theory we could instead manipulate the graph to fuse these tasks, but that would be much more difficult.)

This composed function should have the nice property of leaving the size of the array unchanged.

One drawback of this approach is that it means that our padding internals need to be able to be piped through `apply_ufunc` - so they have to work at the numpy level. Our `grid.pad` API can still be high-level, but the internals of the padding module would need to be lower-level. Sorry I didn't realise this before @jbusecke :grimacing:

TomNicholas:
Sketch of how we might refactor the code to pad inside the ufunc, as discussed in #443.

The idea is that all the complexity of composing the the padding operation with the desired ufunc operation is abstracted away into this `Padder` class, making `apply_as_grid_ufunc` as simple as possible to understand.

Also adds tests for the bug described in #436.

 - [x] Intended to eventually close #436 and #443
 - [x] Tests added

### DIFF
--- a/xgcm/grid_ufunc.py
+++ b/xgcm/grid_ufunc.py
@@ -27,6 +27,8 @@ if TYPE_CHECKING:
 
     from .grid import Grid
 
+from padding import Padder
+
 
 # Modified version of `numpy.lib.function_base._parse_gufunc_signature`
 # Modifications:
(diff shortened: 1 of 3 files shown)


xgcm/padding.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    Internal class which can pad numpy arrays, or compose a numpy ufunc with a particular padding operation.
+
+    One instance of the Padder class should be used for one padding operation, as its internals store the options chosen
+    for that particular use.
+    """
+    grid: "Grid"
+    boundary_width: ...  # TODO work out what all these types need to be
+    boundary: ...
+    fill_value: ...
```
TomNicholas, line 16: @jbusecke I'm posting this as a WIP so you can see what I'm imagining here

xgcm/grid_ufunc.py
```diff
@@ -475,54 +477,37 @@ def apply_as_grid_ufunc(
...
-            padded_args,
-            args,
-            boundary_width_real_axes,
-            grid,
-        )
-    else:
-        rechunked_padded_args = padded_args
+    # Compose the ufunc with the padding needed to replace trimmed elements
+    padder = Padder(grid, boundary_width_real_axes, boundary, fill_value)
+    pad_then_func = padder.compose_with(func)
```
TomNicholas, line 495: Means that this is all the reference to padding I need to have in `apply_as_grid_ufunc`

PR #451, opened 2022-03-24
Exclude common dims

### LINKED ISSUE #443: [BUG] Can't apply grid ufunc with padding if axis position doesn't change

TomNicholas:
Fixes the bug in #443 via [this simple approach](https://github.com/xgcm/xgcm/issues/443#issuecomment-1071248035)

 - [x] Closes #443
 - [x] Tests added

### DIFF
--- a/xgcm/test/test_grid_ufunc.py
+++ b/xgcm/test/test_grid_ufunc.py
@@ -162,7 +162,7 @@ def create_2d_test_grid(ax_name_1, ax_name_2, length1=9, length2=11):
     )
 
 
-class TestGridUFunc:
+class TestGridUFuncNoPadding:
     def test_stores_ufunc_kwarg_info(self):
         signature = "(X:center)->(X:left)"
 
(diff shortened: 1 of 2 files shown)


xgcm/test/test_grid_ufunc.py
```diff
@@ -511,6 +452,168 @@ def grad_to_inner(a):
...
+        result = grid.apply_as_grid_ufunc(
+            vort,
+            U,
+            V,
+            axis=2 * [("lon", "lat")],
+            signature="(lon:left,lat:center),(lon:center,lat:left)->(lon:left,lat:left)",
+            boundary_width={"lon": (1, 0), "lat": (1, 0)},
+            dask="parallelized",  # data isn't chunked along lat/lon
+        )
+        assert_equal(result, expected)
```
TomNicholas, line 575: This test is failing with different values in the assert and I'm not yet sure if it's my result or expected DataArray that is incorrect.

PR #454, opened 2022-03-28
Copied input dataarray to avoid in-place bugs

### LINKED ISSUE #453: [BUG] Operations removing attributes on data
We found a bug in the grid ufunc refactor where using a method like `grid.diff` on data could remove attrs from the data in-place, so that if you tried to operate on it again then it might have lost some of the information needed to set up the grid coordinates correctly.

Its what caused [these failures](https://github.com/xgcm/xgcm/runs/5694540853?check_suite_focus=true).

TomNicholas:
- [x] Closes #453

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1778,7 +1778,8 @@ class Grid:
         else:
             dask = "forbidden"
 
-        array = da
+        # Need to copy to avoid modifying in-place. Ideally we would test for this behaviour specifically
+        array = da.copy()
         # Apply 1D function over multiple axes
         # TODO This will call xarray.apply_ufunc once for each axis, but if signatures + kwargs are the same then we
         # TODO only actually need to call apply_ufunc once for those axes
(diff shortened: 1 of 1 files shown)


xgcm/grid.py
```diff
@@ -1778,7 +1778,8 @@ def _1d_grid_ufunc_dispatch(
         else:
             dask = "forbidden"
 
-        array = da
+        # Need to copy to avoid modifying in-place. Ideally we would test for this behaviour specifically
+        array = da.copy()
```
rabernat, line 1782: Just for my understanding--this copies the metadata but not the actual numpy data, correct?

TomNicholas: Good question, I think actually we might want to pass `deep=False` (see the `.copy` [docstring](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.copy.html))

jbusecke: Uh yeah we definitely do not want to load data into memory here! I always assumed the default was deep=False.

PR #431, opened 2022-02-15
Grid ufunc explanation docs

TomNicholas:
- [x] Closes the "explanation" part of #423
 - [x] Passes `pre-commit run --all-files`

PR #456, opened 2022-03-30
Enable boundary kwargs in ufunc decorator

TomNicholas:
Fixes a bug I found whilst writing examples for #431.

 - [x] Tests added


jbusecke:
Nice catch!

Issue #458, opened 2022-03-31
[FEATURE] Rewiring `cumsum` via (parts of) the grid_ufunc logic.


jbusecke:
While working on https://github.com/jbusecke/xgcm/tree/vector_padding I am encountering some test failures, specifically the `xgcm/test/test_grid.py::test_keep_coords[xx-cumint]`, `xgcm/test/test_grid.py::test_keep_coords[xx-cumsum]` tests which are testing the consistency of output coordinates. 

While checking on this I noticed that the `Grid.cumsum` method in the `grid_ufunc_refactor` branch still calls the `Axis.cumsum` method, which has a bunch of [custom logic for padding/rewrapping the output](https://github.com/jbusecke/xgcm/blob/c742c47dbfee0a48b034baec965d33ef3fb21579/xgcm/grid.py#L702-L721) ([which itself is simply xarray.cumsum()](https://github.com/jbusecke/xgcm/blob/c742c47dbfee0a48b034baec965d33ef3fb21579/xgcm/grid.py#L700)). 
We will ultimately have to refactor this logic into the Grid method, when we remove the Axis methods. 

Should/Can this logic be implemented via the gufunc logic? That would solve my issue here, since [these changes I made](https://github.com/jbusecke/xgcm/blob/c742c47dbfee0a48b034baec965d33ef3fb21579/xgcm/grid_ufunc.py#L994-L1026) would ensure that the treatment of output coordinates is always consistent.

TomNicholas:
My plan is to add an extra `pad_before` kwarg to `apply_as_grid_ufunc`, which defaults to `True`. If it's set to false then we instead pad *after* the `xr.apply_ufunc` call. Then we have various `cumsum` ufuncs defined in `gridops.py` that have different `boundary_width` and trimming behaviour to replicate the bespoke `_pad_array` logic that @rabernat originally wrote.

This approach has the advantage of being easy to change if we do end up [moving padding inside `xr.apply_ufunc`](https://github.com/xgcm/xgcm/issues/436) later: it would become just a flip in the order of the function composition. It also means that someone could define new grid ufuncs which pad after their application (I cannot think of any examples of this other than cumulative sums/integrals, but we would get this option for free).

TomNicholas:
The other complication is that we don't want to use `map_overlap=True` for chunked core dims for the specific case of cumsum, because unlike all the other pre-defined ufuncs it handles dask arrays itself

TomNicholas:
[...]

PR #459, opened 2022-03-31
Extended functionality vector padding

### LINKED ISSUE #410: Extended functionality for vector data with complex grid topologies

jbusecke:
- [x] Closes #410 
 - [x] Tests added

This PR moves all padding logic into a separate module `padding.py` and adds functionality for padding vector components (with extended functionality for vectors and face connections (see #410 for details).

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -14,7 +14,7 @@ from . import comodo, gridops
 from .duck_array_ops import _apply_boundary_condition, _pad_array, concatenate
 from .grid_ufunc import (
     GridUFunc,
-    Signature,
+    _GridUFuncSignature,
     _has_chunked_core_dims,
     _maybe_unpack_vector_component,
     apply_as_grid_ufunc,
@@ -1731,6 +1731,8 @@ class Grid:
         else:
             array = da.copy(deep=False)
 
+        # Need to copy to avoid modifying in-place. Ideally we would test for this behaviour specifically
+        array = da.copy(deep=False)
         # Apply 1D function over multiple axes
         # TODO This will call xarray.apply_ufunc once for each axis, but if signatures + kwargs are the same then we
         # TODO only actually need to call apply_ufunc once for those axes
@@ -1771,7 +1773,9 @@ class Grid:
 
         return self._transpose_to_keep_same_dim_order(da, array, axis)
 
-    def _create_1d_grid_ufunc_signatures(self, da, axis, to) -> List[Signature]:
+    def _create_1d_grid_ufunc_signatures(
+        self, da, axis, to
+    ) -> List[_GridUFuncSignature]:
         """
         Create a list of signatures to pass to apply_grid_ufunc.
 
@@ -1790,7 +1794,10 @@ class Grid:
             if to_pos is None:
                 to_pos = ax._default_shifts[from_pos]
 
-            signature_1d = Signature(f"({ax_name}:{from_pos})->({ax_name}:{to_pos})")
+            # TODO build this more directly?
+            signature_1d = _GridUFuncSignature.from_string(
+                f"({ax_name}:{from_pos})->({ax_name}:{to_pos})"
+            )
             signatures.append(signature_1d)
 
         return signatures
@@ -1820,7 +1827,7 @@ class Grid:
         func: Callable,
         *args: xr.DataArray,
         axis: Sequence[Sequence[str]] = None,
-        signature: Union[str, Signature] = "",
+        signature: Union[str, _GridUFuncSignature] = "",
         boundary_width: Mapping[str, Tuple[int, int]] = None,
         boundary: Union[str, Mapping[str, str]] = None,
         fill_value: Union[float, Mapping[str, float]] = None,
@@ -2197,9 +2204,7 @@ class Grid:
         da_i : xarray.DataArray
             The differentiated data
         """
-
-        ax = self.axes[axis]
-        diff = ax.diff(da, **kwargs)
+        diff = self.diff(da, axis, **kwargs)
         dx = self.get_metric(diff, (axis,))
         return diff / dx
 
@@ -2364,7 +2369,7 @@ class Grid:
         return ax.transform(da, target, **kwargs)
 
 
-def _select_grid_ufunc(funcname, signature: Signature, module, **kwargs):
+def _select_grid_ufunc(funcname, signature: _GridUFuncSignature, module, **kwargs):
     # TODO to select via other kwargs (e.g. boundary) the signature of this function needs to be generalised
 
     def is_grid_ufunc(obj):
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1150,12 +1150,38 @@ class Grid:
                         f"Input `{dim}` (for the `{pos}` position on axis `{axis}`) is not a dimension in the input datasets `ds`."
                     )
 
+        # Convert all inputs to axes-kwarg mappings
+        # Parse axis properties
+        boundary = self._as_axis_kwarg_mapping(boundary, axes=all_axes)
+        fill_value = self._as_axis_kwarg_mapping(fill_value, axes=all_axes)
+
+        # this case does only apply to periodic. Since we plan on deprecating it soon
+        # handle it here
+        if isinstance(periodic, list):
+            periodic = {axname: True for axname in periodic}
+        periodic = self._as_axis_kwarg_mapping(periodic, axes=all_axes)
+
+        # Set properties on
+
+        # Populate axes. Much of this is just for backward compatibility. I think we are better of
+        # getting completely rid of the axis object and storing/getting info from the grid object properties
+        # (all of which are axis-value mapping).
+        self.boundary = boundary
+        self.fill_value = fill_value
+        self.periodic = periodic
+        # TODO we probably want to properly define these as class properties with setter/getter?
+
         self.axes = OrderedDict()
         for axis_name in all_axes:
-            try:
-                is_periodic = axis_name in periodic
-            except TypeError:
-                is_periodic = periodic
+            # try:
+            #     is_periodic = axis_name in periodic
+            # except TypeError:
+            #     is_periodic = periodic
+
+            # setting this properly is still needed for the transform axis method
+            # TODO: migrate that over to grid
+            is_periodic = periodic.get(axis_name, False)
+
             if axis_name in default_shifts:
                 axis_default_shifts = default_shifts[axis_name]
             else:
@@ -1170,7 +1196,7 @@ class Grid:
                     f"boundary={boundary} is invalid. Please specify a dictionary "
                     "mapping axis name to a boundary option; a string or None."
                 )
-
+            # TODO Reimplement value checking on grid properties.
             if isinstance(fill_value, dict):
                 axis_fillvalue = fill_value.get(axis_name, None)
             elif isinstance(fill_value, (int, float)) or fill_value is None:
@@ -1200,18 +1226,23 @@ class Grid:
             for key, value in metrics.items():
                 self.set_metrics(key, value)
 
-    def _parse_axes_kwargs(self, kwargs):
-        """Convvert kwarg input into dict for each available axis
+        # Finish setup
+
+    def _as_axis_kwarg_mapping(self, kwargs, axes=None):
+        """Convert kwarg input into dict for each available axis
         E.g. for a grid with 2 axes for the keyword argument `periodic`
         periodic = True --> periodic = {'X': True, 'Y':True}
         or if not all axes are provided, the other axes will be parsed as defaults (None)
         periodic = {'X':True} --> periodic={'X': True, 'Y':None}
         """
+        if axes is None:
+            axes = self.axes
+
         parsed_kwargs = dict()
         if isinstance(kwargs, dict):
             parsed_kwargs = kwargs
         else:
-            for axis in self.axes:
+            for axis in axes:
                 parsed_kwargs[axis] = kwargs
         return parsed_kwargs
 
@@ -1543,7 +1574,7 @@ class Grid:
         if (not isinstance(axis, list)) and (not isinstance(axis, tuple)):
             axis = [axis]
         # parse multi axis kwargs like e.g. `boundary`
-        multi_kwargs = {k: self._parse_axes_kwargs(v) for k, v in kwargs.items()}
+        multi_kwargs = {k: self._as_axis_kwarg_mapping(v) for k, v in kwargs.items()}
 
         out = da
         for axx in axis:
@@ -1607,6 +1638,8 @@ class Grid:
             fill_value = 0.0
         if isinstance(fill_value, float):
             fill_value = {ax_name: fill_value for ax_name in self.axes.keys()}
+        # xgcm uses a default fill value of 0, while xarray uses nan.
+        fill_value = {k: 0.0 if v is None else v for k, v in fill_value.items()}
 
         padded = []
         for da in arrays:
@@ -1667,10 +1700,53 @@ class Grid:
 
         if isinstance(axis, str):
             axis = [axis]
-        if to is None:
-            to = {ax: None for ax in axis}
-        elif isinstance(to, str):
-            to = {ax: to for ax in axis}
+
+        # convert inpug arguments into axes-kwarg mappings
+        to = self._as_axis_kwarg_mapping(to)
+        # if to is None:
+        #     to = {ax: None for ax in axis}
+        # elif isinstance(to, str):
+        #     to = {ax: to for ax in axis}
+
+        # If I understand correctly, this could contain some other kwargs, which are not appropriate to convert
+        # Note that the option to pass boundary (actually padding) options on the method will be deprecated,
+        # so for now I will just store everything I need here.
+        VALID_BOUNDARY_KWARGS = ["boundary", "periodic", "fill_value"]
+        kwargs = {
+            k: self._as_axis_kwarg_mapping(v) if k in VALID_BOUNDARY_KWARGS else v
+            for k, v in kwargs.items()
+        }
+        # Now check if some of the values not provided in the kwargs are set as grid properties
+        # In the future this should be the only way to access that information!
+        kwargs_with_defaults = {}
+
+        for k in np.unique(
+            list(kwargs.keys()) + VALID_BOUNDARY_KWARGS
+        ):  # iterate over all axis properties and any additional keys of `kwarg`
+            if k in VALID_BOUNDARY_KWARGS:
+                defaults = getattr(self, k, {})
+                kwargs_input = kwargs.get(k, {})
+                # overwrite if given as input
+                defaults.update(kwargs_input)
+                kwargs_with_defaults[k] = defaults
+            else:
+                kwargs_with_defaults[k] = kwargs[k]
+        kwargs = kwargs_with_defaults
+
+        # Lastly go through the inputs axis by axis, and replace
+        # boundary with 'periodic' if periodic is True
+        # (again this is temporary until `periodic` is deprecated)
+        if "periodic" in kwargs.keys():
+            boundary = kwargs.get("boundary", {})
+            new_boundary = {
+                k: "periodic"
+                for k in kwargs["periodic"].keys()
+                if kwargs["periodic"].get(k, False)
+            }
+            boundary.update(new_boundary)
+            kwargs["boundary"] = boundary
+        # remove the periodic input
+        kwargs = {k: v for k, v in kwargs.items() if k != "periodic"}
 
         signatures = self._create_1d_grid_ufunc_signatures(da, axis=axis, to=to)
 
@@ -1698,10 +1774,10 @@ class Grid:
         One separate signature is created for each axis the 1D ufunc is going to be applied over.
         """
         signatures = []
-        for ax_name, to_pos in itertools.zip_longest(axis, to):
+        for ax_name in axis:
             ax = self.axes[ax_name]
 
-            from_pos, dim = ax._get_position_name(da)
+            from_pos, _ = ax._get_position_name(da)  # removed `dim` since it wasnt used
 
             to_pos = to[ax_name]
             if to_pos is None:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1765,14 +1765,10 @@ class Grid:
 
         if (not isinstance(axis, list)) and (not isinstance(axis, tuple)):
             axis = [axis]
-        # # parse multi axis kwargs like e.g. `boundary`
-        # multi_kwargs = {k: self._as_axis_kwarg_mapping(v) for k, v in kwargs.items()}
 
         out = da
         for axx in axis:
-            # kwargs = {k: v[axx] for k, v in multi_kwargs.items()}
             ax = self.axes[axx]
-            # kwargs.setdefault("boundary", ax.boundary)
             func = getattr(ax, funcname)
             metric_weighted = kwargs.pop("metric_weighted", False)
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -1680,7 +1680,16 @@ class Grid:
         return padded
 
     def _1d_grid_ufunc_dispatch(
-        self, funcname, da, axis, to=None, keep_coords=False, **kwargs
+        self,
+        funcname,
+        da,
+        axis,
+        to=None,
+        keep_coords=False,
+        metric_weighted: Union[
+            str, list, tuple, Dict[str, Union[str, list, tuple]]
+        ] = False,
+        **kwargs,
     ):
         """
         Calls appropriate 1D grid ufuncs on data, along the specified axes, sequentially.
@@ -1702,6 +1711,10 @@ class Grid:
 
         # convert inpug arguments into axes-kwarg mappings
         to = self._as_axis_kwarg_mapping(to)
+
+        if isinstance(metric_weighted, str):
+            metric_weighted = (metric_weighted,)
+        metric_weighted = self._as_axis_kwarg_mapping(metric_weighted)
 
         # If I understand correctly, this could contain some other kwargs, which are not appropriate to convert
         # Note that the option to pass boundary (actually padding) options on the method will be deprecated,
@@ -1756,10 +1769,19 @@ class Grid:
             grid_ufunc, remaining_kwargs = _select_grid_ufunc(
                 funcname, signature_1d, module=gridops, **kwargs
             )
+            ax_metric_weighted = metric_weighted[ax_name]
+
+            if ax_metric_weighted:
+                metric = self.get_metric(array, ax_metric_weighted)
+                array = array * metric
 
             array = grid_ufunc(
                 self, array, axis=[ax_name], keep_coords=keep_coords, **remaining_kwargs
             )
+
+            if ax_metric_weighted:
+                metric = self.get_metric(array, ax_metric_weighted)
+                array = array / metric
 
         return self._transpose_to_keep_same_dim_order(da, array, axis)
 
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -21,6 +21,7 @@ try:
 except ImportError:
     numba = None
 
+from typing import Dict, Optional, Union
 
 docstrings = docrep.DocstringProcessor(doc_key="My doc string")
 
@@ -1155,31 +1156,25 @@ class Grid:
         boundary = self._as_axis_kwarg_mapping(boundary, axes=all_axes)
         fill_value = self._as_axis_kwarg_mapping(fill_value, axes=all_axes)
 
-        # this case does only apply to periodic. Since we plan on deprecating it soon
-        # handle it here
+        # Parse list input. This case does only apply to periodic.
+        # Since we plan on deprecating it soon handle it here, so we can easily
+        # remove it later
         if isinstance(periodic, list):
             periodic = {axname: True for axname in periodic}
         periodic = self._as_axis_kwarg_mapping(periodic, axes=all_axes)
 
-        # Set properties on
-
-        # Populate axes. Much of this is just for backward compatibility. I think we are better of
+        # Set properties on grid object. I think we are better of
         # getting completely rid of the axis object and storing/getting info from the grid object properties
-        # (all of which are axis-value mapping).
+        # (all of which need to be supplied/converted to axis-value mapping).
         self.boundary = boundary
         self.fill_value = fill_value
         self.periodic = periodic
         # TODO we probably want to properly define these as class properties with setter/getter?
+        # TODO: This also needs to check valid inputs for each one.
 
+        # Populate axes. Much of this is just for backward compatibility.
         self.axes = OrderedDict()
         for axis_name in all_axes:
-            # try:
-            #     is_periodic = axis_name in periodic
-            # except TypeError:
-            #     is_periodic = periodic
-
-            # setting this properly is still needed for the transform axis method
-            # TODO: migrate that over to grid
             is_periodic = periodic.get(axis_name, False)
 
             if axis_name in default_shifts:
@@ -1196,7 +1191,7 @@ class Grid:
                     f"boundary={boundary} is invalid. Please specify a dictionary "
                     "mapping axis name to a boundary option; a string or None."
                 )
-            # TODO Reimplement value checking on grid properties.
+            # TODO Reimplement value checking on grid properties. See comment above
             if isinstance(fill_value, dict):
                 axis_fillvalue = fill_value.get(axis_name, None)
             elif isinstance(fill_value, (int, float)) or fill_value is None:
@@ -1228,7 +1223,11 @@ class Grid:
 
         # Finish setup
 
-    def _as_axis_kwarg_mapping(self, kwargs, axes=None):
+    def _as_axis_kwarg_mapping(
+        self,
+        kwargs: Union[str, float, int, Dict[str, Union[str, float, int]]],
+        axes: Optional[list] = None,
+    ):
         """Convert kwarg input into dict for each available axis
         E.g. for a grid with 2 axes for the keyword argument `periodic`
         periodic = True --> periodic = {'X': True, 'Y':True}
@@ -1703,10 +1702,6 @@ class Grid:
 
         # convert inpug arguments into axes-kwarg mappings
         to = self._as_axis_kwarg_mapping(to)
-        # if to is None:
-        #     to = {ax: None for ax in axis}
-        # elif isinstance(to, str):
-        #     to = {ax: to for ax in axis}
 
         # If I understand correctly, this could contain some other kwargs, which are not appropriate to convert
         # Note that the option to pass boundary (actually padding) options on the method will be deprecated,
@@ -1716,11 +1711,13 @@ class Grid:
             k: self._as_axis_kwarg_mapping(v) if k in VALID_BOUNDARY_KWARGS else v
             for k, v in kwargs.items()
         }
+
         # Now check if some of the values not provided in the kwargs are set as grid properties
-        # In the future this should be the only way to access that information!
+        # In the future grid object properties should be the only way to access that information
+        # and we can remove this.
         kwargs_with_defaults = {}
 
-        for k in np.unique(
+        for k in set(
             list(kwargs.keys()) + VALID_BOUNDARY_KWARGS
         ):  # iterate over all axis properties and any additional keys of `kwarg`
             if k in VALID_BOUNDARY_KWARGS:
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -277,6 +277,7 @@ class Axis:
         da_i : xarray.DataArray
             The differenced data
         """
+
         position_from, dim = self._get_position_name(da)
         if to is None:
             to = self._default_shifts[position_from]
@@ -285,6 +286,7 @@ class Axis:
             boundary = self.boundary
         if fill_value is None:
             fill_value = self.fill_value
+
         data_new = self._neighbor_binary_func_raw(
             da,
             f,
@@ -1169,6 +1171,9 @@ class Grid:
         self.boundary: Dict[str, Union[str, float, int]] = boundary
         self.fill_value: Dict[str, Union[str, float, int]] = fill_value
         self.periodic: Dict[str, Union[str, float, int]] = periodic
+        if face_connections is not None:
+            self._connections = face_connections
+            self._facedim = list(face_connections.keys())[0]
         # TODO we probably want to properly define these as class properties with setter/getter?
         # TODO: This also needs to check valid inputs for each one.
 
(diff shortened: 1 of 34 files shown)


xgcm/grid.py
```diff
@@ -2102,6 +2059,13 @@ def cumsum(self, da, axis, **kwargs):
     def _apply_vector_function(self, function, vector, **kwargs):
         # the keys, should be axis names
         assert len(vector) == 2
+        assert isinstance(vector, dict)
+
+        warnings.warn(
```
jbusecke, line 2064: Not sure why this is showing up as an addition? I thought we had those warnings already merged into main? Is the `grid_ufunc...` branch up to date with main?

xgcm/grid.py
```diff
@@ -2124,25 +2088,67 @@ def _apply_vector_function(self, function, vector, **kwargs):
...
+
+        %(neighbor_binary_func.parameters.no_f)s
+
+        Returns
+        -------
+        vector_diff : dict
+            A dictionary with two entries. Keys are axis names, values
+            are differenced vector components along each axis
+        """
+        # # the keys, should be axis names
```
jbusecke, line 2128: Need to remove these commented lines

xgcm/grid.py
```diff
@@ -2163,8 +2169,31 @@ def interp_2d_vector(self, vector, **kwargs):
             A dictionary with two entries. Keys are axis names, values
             are interpolated vector components along each axis
         """
+        # ! I am implementing this here and remove _apply_vector_function, since these are going to get removed anywyas
+
+        # # the keys, should be axis names
+        # assert len(vector) == 2
 
-        return self._apply_vector_function(Axis.interp, vector, **kwargs)
+        # x_axis_name, y_axis_name = list(vector)
```
jbusecke, line 2177: Same here

xgcm/grid.py
```diff
@@ -2346,29 +2375,6 @@ def transform(self, da, axis, target, **kwargs):
         ax = self.axes[axis]
         return ax.transform(da, target, **kwargs)
 
-    @docstrings.dedent
```
jbusecke, line 2349: I moved this up, because it was kind of weird to have those methods split far apart in the code.

xgcm/grid_ufunc.py
```diff
@@ -738,7 +793,10 @@ def apply_as_grid_ufunc(
 
     # Restore any dimension coordinates associated with new output dims that are present in grid
     results_with_coords = _reattach_coords(
-        results, grid, out_core_dims, boundary_width, keep_coords
+        results,
+        grid,
+        boundary_width,
+        keep_coords,  # TODO: Do we still need all these inputs?
```
jbusecke, line 799: We do, disregard this comment

xgcm/test/test_faceconnections.py
```diff
@@ -211,36 +229,80 @@ def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
...
     )
 
     # last point should be fancy
     np.testing.assert_allclose(
-        u_c_interp.data[0, -1, :], 0.5 * (ds.u.data[0, -1, :] + ds.v.data[1, ::-1, 0])
+        u_c_interp.data[0, -1, :], 0.5 * (u.data[0, -1, :] + v.data[1, ::-1, 0])
     )
     np.testing.assert_allclose(
-        u_c_diff.data[0, -1, :], -ds.u.data[0, -1, :] + ds.v.data[1, ::-1, 0]
+        u_c_diff.data[0, -1, :], -u.data[0, -1, :] + v.data[1, ::-1, 0]
```
jbusecke, line 305: This was just for testing. I should restore this test to the old form.

xgcm/test/test_grid.py
```diff
@@ -783,13 +795,22 @@ def test_keep_coords(funcname, gridtype):
         result = func(ds.tracer, axis_name)
         base_coords = list(result.dims)
         augmented_coords = [
-            c for c in ds.tracer.coords if set(ds[c].dims).issubset(result.dims)
+            c
```
jbusecke, line 798: Need to revert this too, please disregard for now.

jbusecke:
Still not quite there, but I would really appreciate an initial review (particularly the bits that tie into `grid.py` and `grid_ufunc.py` @TomNicholas. Cheers!

xgcm/grid.py
TomNicholas: No it probably isn't up to date with main... I can fix that and then you merge `grid_ufunc_refactor_project` in to this branch?

xgcm/grid_ufunc.py
```diff
@@ -593,8 +612,36 @@ def apply_as_grid_ufunc(
...
+            # if more than one input is provided check that we get the same number
+            # of other_component inputs as args
+            if not len(args) == len(other_component):
+                raise ValueError(
+                    "When providing multiple input arguments, `other_component` needs to provide one dictionary per input"
+                )
+            # TODO: Raise this error in tests
+        return other_component
+
+    other_component = _parse_and_check_other_component(other_component)
```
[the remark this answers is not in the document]
TomNicholas: If it beats us we can tell mypy to ignore it.

xgcm/grid_ufunc.py
```diff
@@ -39,6 +41,16 @@
...
 
 
+def _maybe_unpack_vector_component(da):
+    if isinstance(da, dict):
+        da_component = da[
+            list(da.keys())[0]
+        ]  # we only allow len 1 dicts previously. Not sure if this is the most elegant way to do this
+    else:
+        da_component = da
+    return da_component
```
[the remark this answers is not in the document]
TomNicholas: Actually if we had an internal class for vectors then this could be a method on that class...

TomNicholas, line 51:
```suggestion
def _maybe_unpack_vector_component(data: Union[DataArray, dict[str, DataArray]]) -> DataArray:
    if isinstance(data, dict):
        [da] = list(data)  # this will raise if more than one element
        return da
    else:
        return da
```

xgcm/grid_ufunc.py
```diff
@@ -510,6 +522,9 @@ def apply_as_grid_ufunc(
     keep_coords: bool = True,
     dask: Literal["forbidden", "parallelized", "allowed"] = "forbidden",
     map_overlap: bool = False,
+    other_component: Union[
+        None, Dict[str, xr.DataArray], Sequence[Dict[str, xr.DataArray]]
+    ] = None,
```
TomNicholas, line 527:
```suggestion
    other_component: Union[
        Dict[str, xr.DataArray], Sequence[Dict[str, xr.DataArray]]
    ] = None,
```
The first `None` is redundant here.

xgcm/grid_ufunc.py
```diff
@@ -510,6 +522,9 @@ def apply_as_grid_ufunc(
     keep_coords: bool = True,
     dask: Literal["forbidden", "parallelized", "allowed"] = "forbidden",
     map_overlap: bool = False,
+    other_component: Union[
+        None, Dict[str, xr.DataArray], Sequence[Dict[str, xr.DataArray]]
```
TomNicholas, line 526: Is this what you were referring to with lists of dicts?

xgcm/grid_ufunc.py
```diff
@@ -527,7 +542,8 @@ def apply_as_grid_ufunc(
 
         Passed directly on to `xarray.apply_ufunc`.
     *args : xarray.DataArray
-        One or more xarray DataArray objects to apply the function to.
+        One or more xarray DataArray objects (for scalar fields) or dictonaries mapping the direction
+        to vector components to apply the function to.
```
TomNicholas, line 546: This could be clearer.

xgcm/grid_ufunc.py
```diff
@@ -571,6 +587,9 @@ def apply_as_grid_ufunc(
     map_overlap : bool, optional
         Whether or not to automatically apply the function along chunked core dimensions using dask.array.map_overlap.
         Default is False. If True, will need to be accompanied by dask='allowed'.
+    other_component : Union[None, Dict[str,xr.DataArray], Sequence[Dict[str,xr.DataArray]]], default: False
```
TomNicholas, line 590: The default you gave it isn't `False`, it's `None`.

xgcm/grid_ufunc.py
```diff
@@ -593,8 +612,36 @@ def apply_as_grid_ufunc(
     if grid is None:
         raise ValueError("Must provide a grid object to describe the Axes")
 
-    if any(not isinstance(arg, xr.DataArray) for arg in args):
-        raise TypeError("All data arguments must be of type DataArray")
+    # TODO: we need some more comprehensive tests that check proper padding with
+    # TODO: e.g. multiple vector components or mixed (tracer, vector components)
+    if not all(isinstance(a, xr.DataArray) or isinstance(a, dict) for a in args):
```
TomNicholas, line 617:
```suggestion
    if not all(isinstance(a, (xr.DataArray, dict)) for a in args):
```
You can check against multiple types by using a tuple, like this.

But also this isn't really a sufficient check, because it doesn't check that the dicts only contain 1 dataarray each. It's important to check this here because it's a place where it's likely that users get confused and try inputting the wrong types.

xgcm/grid_ufunc.py
```diff
@@ -593,8 +612,36 @@ def apply_as_grid_ufunc(
...
+        raise TypeError(
+            "All data arguments must be of type DataArray or dict (for vector components"
+        )
+
+    def _parse_and_check_other_component(
```
TomNicholas, line 622: Can we think of a better name for this function?

TomNicholas (changes requested):
Great work @jbusecke !

However I haven't even got to reviewing the `padding.py` module yet and I already have lots of comments :sweat_smile: 
Will try and finish reviewing the rest tomorrow!

xgcm/grid.py
jbusecke: Sounds like a plan!

xgcm/grid.py
```diff
@@ -1376,6 +1392,7 @@ def set_metrics(self, key, value, overwrite=False):
                 self._metrics[metric_axes].append(metric_var)
 
     def _get_dims_from_axis(self, da, axis):
+        da = _maybe_unpack_vector_component(da)
```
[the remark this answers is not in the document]
jbusecke: 👍 for the type hints. But do you really expect users to access this functionality? Its pretty thin?

xgcm/grid.py
```diff
@@ -1610,97 +1627,19 @@ def _grid_func(self, funcname, da, axis, **kwargs):
...
-            padded.append(new_da)
-
-        return padded
-
     def _1d_grid_ufunc_dispatch(
         self,
         funcname,
-        da,
+        da: Union[
+            xr.DataArray, Dict[str, xr.DataArray]
```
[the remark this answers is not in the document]
jbusecke: I am already doing that somewhere, but can promote it to here

xgcm/grid.py
```diff
@@ -1721,6 +1660,13 @@ def _1d_grid_ufunc_dispatch(
         if isinstance(axis, str):
             axis = [axis]
 
+        # Check validity of vector
+        if isinstance(da, dict):
+            if len(da) != 1:
+                raise ValueError(
+                    f"When providing vector components as input, the provided dictionary can only have one key/value pair. Found {len(da)}"
+                )
+
```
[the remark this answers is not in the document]
jbusecke: Yeah thats good. Just checking: You want me to globally rename da-> data in this function (and possibly others that also accept scalars/vectors?

xgcm/grid.py
```diff
@@ -2102,6 +2059,13 @@ def cumsum(self, da, axis, **kwargs):
     def _apply_vector_function(self, function, vector, **kwargs):
         # the keys, should be axis names
         assert len(vector) == 2
+        assert isinstance(vector, dict)
```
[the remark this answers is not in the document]
jbusecke: Oh sorry that was my bad. This was supposed to be converted before!

xgcm/grid_ufunc.py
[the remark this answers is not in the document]
jbusecke: This seems like a recurring theme. Lets talk about it!

xgcm/grid_ufunc.py
jbusecke: Yes exactly

xgcm/grid_ufunc.py
jbusecke: Oh neat! I didnt actually know that. I think we need to first add some tests for the cases where we want errors to be raised and then refactor the input checking as you suggested.

jbusecke:
Thanks a ton. This is super helpful. Let me get on implementing your suggestions while you review padding.py (i am a bit afraid hahahaha)

xgcm/grid.py
```diff
@@ -1181,6 +1193,10 @@ def __init__(
         self.boundary: Dict[str, Union[str, float, int]] = boundary
         self.fill_value: Dict[str, Union[str, float, int]] = fill_value
         self.periodic: Dict[str, Union[str, float, int]] = periodic
+        if face_connections is not None:
+            self._connections = face_connections
+            self._facedim = list(face_connections.keys())[0]
```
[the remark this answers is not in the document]
jbusecke: Do you think something like
```python
self._connections = None
self._facedim = None
if face_connections is not None:
            self._connections = face_connections
            self._facedim = list(face_connections.keys())[0]
```
would be sufficient? Or did I misunderstand your comment?

xgcm/test/test_grid.py
jbusecke: I take that back. I believe the old test was not properly testing the behavior we want to ensure here. I hope this small change to tests is ok.

xgcm/grid.py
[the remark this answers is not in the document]
TomNicholas: Could probably make it neater but that's the idea yeah.

xgcm/grid.py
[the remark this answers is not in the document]
jbusecke: As per our offline discussion we leave this for a follow up. Ill try to make this work for now, and then need to raise an issue as a reminder.

xgcm/test/test_faceconnections.py
```diff
@@ -7,11 +7,14 @@
 
 @pytest.fixture(scope="module")
 def ds():
-    N = 25
+    N = 25  # TODO: Reduce the size here. I think something like 4 or 5 is sufficient
```
TomNicholas, line 10: God yeah it doesn't need to be anywhere near that big. I think 5 at most, at least if you want to test moving two boundaries, each of which could be flipped (so need to be length >=2).

xgcm/test/test_faceconnections.py
```diff
@@ -7,11 +7,14 @@
...
     return xr.Dataset(
         {
             "data_c": (["face", "y", "x"], np.random.rand(2, N, N)),
-            "u": (["face", "xl", "y"], np.random.rand(2, N, N)),
+            "u": (
+                ["face", "xl", "y"],
+                np.random.rand(2, N, N),
+            ),  # TODO: Will it make testing easier if I make these not random?
```
TomNicholas, line 17: I would use a test function that has some kind of linear increase across the domain, so that it's really obvious when values are in the wrong place, and where they should be.

Making the gradient different along different dimensions (or the magnitude completely different) would help avoid mixing up one dimension with another).

xgcm/test/test_faceconnections.py
```diff
@@ -211,19 +229,59 @@ def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
...
+    u = ds.u * 0 + u_modifier
+    v = ds.v * 0 + v_modifier
+
+    # no need to check for diff vs interp. They all go through the same dispatch
+    # v is the interesting variable here because it involves a sign change for this
+    # connection (see https://github.com/xgcm/xgcm/issues/410#issue-1098348557)
+    v_out = grid.interp({"Y": v}, "X", other_component={"X": u})
+    # the test case is set up in a way that all interpolated values for u should be 1
+    # if the face connection is done properly
+    np.testing.assert_allclose(v_out.data, 1)
```
TomNicholas, line 262: Why not use [`xarray.testing.assert_allclose`](https://xarray.pydata.org/en/v2022.03.0/generated/xarray.testing.assert_allclose.html)?

xgcm/test/test_faceconnections.py
```diff
@@ -211,19 +229,59 @@ def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
     # TODO: checking all the other boundaries
 
 
-@pytest.mark.xfail(reason="connected grids not implemented for grid ufunc refactor yet")
+# TODO: Relaease the periodic test here once we unified the API with padding.
+@pytest.mark.parametrize(
+    "boundary", [pytest.param("periodic", marks=pytest.mark.xfail), "fill"]
+)
```
TomNicholas, line 235: I dont' really understand why this is currently xfailed

xgcm/test/test_faceconnections.py
```diff
@@ -211,19 +229,59 @@ def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
...
     # simplest scenario with one face connection
     grid = Grid(ds, face_connections=ds_face_connections_x_to_y)
 
     # interp u and v to cell center
     vector_center = grid.interp_2d_vector(
-        {"X": ds.u, "Y": ds.v}, to="center", boundary="fill"
+        {"X": ds.u, "Y": ds.v},
+        to="center",
+        boundary="fill",
+        # test
```
TomNicholas, line 274: ?

xgcm/test/test_grid.py
```diff
@@ -768,9 +768,21 @@ def test_invalid_fill_value_error():
...
+        ),
+        pytest.param(
+            "cumint",
+            marks=pytest.mark.xfail(
+                reason="cumsum is not yet goint through grid ufunc logic TODO"
+            ),
+        ),
         "derivative",
+        # TODO: we can get rid of many of these after the release. With the grid_ufunc logic many of these go through the same codepath
+        # e.g. diff/interp/min/max all are the same, so we can probably reduce this to diff, cumsum, integrate, derivative, cumint
```
TomNicholas, line 785: Yes. There are other places in the testing codebase that we could do this too.

xgcm/test/test_grid.py
```diff
@@ -783,13 +795,17 @@ def test_keep_coords(funcname, gridtype):
...
+            for c in ds.coords
+            if set(ds[c].dims).issubset(result.dims) and c not in result.dims
         ]
+
         if funcname in ["integrate", "average"]:
             assert set(result.coords) == set(base_coords + augmented_coords)
         else:
             assert set(result.coords) == set(base_coords)
 
+        # TODO: why is the behavior different for integrate and average?
```
TomNicholas, line 808: What is different about them?

TomNicholas (changes requested):
Another 22 comments, just `test_padding.py` to go...

One general comment is that whilst the full face-connected vector padding is complicated, it should be possible to split it up into a lot of small functions which each do one thing. I want to be able to understand the gist of your padding algorithm just by reading the names of the functions you call one-by-one, e.g. `extend this`, `flip that`, `attach this`, `trim that`.

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+        {"X": (0, 1), "Y": (2, 0)},
+    ],
+)
+class TestPadding:
+    @pytest.mark.parametrize("fill_value", [np.nan, 0, 1.5])
+    def test_padding_fill(self, boundary_width, fill_value):
+
+        ds, coords, _ = datasets_grid_metric("C")
+        grid = Grid(ds, coords=coords)
+        data = ds.tracer
```
TomNicholas, line 26: One function could supply all these in one line?

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+        result = pad(
+            data,
+            grid,
+            boundary="fill",
+            boundary_width=boundary_width,
+            fill_value=fill_value,
+        )
+        xr.testing.assert_allclose(expected, result)
+
+    def test_padding_extend(self, boundary_width):
```
TomNicholas, line 50: I thought we were deprecating extend? Or was that another option we were deprecating?

TomNicholas: Oh I was thinking of extrapolate

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+
+@pytest.mark.parametrize(
+    "boundary_width",
+    [
+        {"X": (1, 1)},
+        {"Y": (0, 1)},
+        {"X": (0, 1), "Y": (2, 0)},
+    ],
+)
+class TestPadding:
```
TomNicholas, line 20: Can all but the last test in this class be parameterized into one?

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+        expected = data.pad({"xt": (0, 1)}, "wrap")
+
+        # we intentionally strip all coords from padded arrays
+        expected = _strip_all_coords(expected)
+
+        result = pad(data, grid, boundary=None, boundary_width={"X": (0, 1)})
+        xr.testing.assert_allclose(expected, result)
+
+
+# TODO: We should make both boundary and boundary width positional arguments. Then this class can be deleted.
```
TomNicholas, line 171: You mean make them positional arguments to the pad function? I think they should remain optional arguments to `apply_as_grid_ufunc`.

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+        ds, coords, _ = datasets_grid_metric("C")
+        grid = Grid(ds, coords=coords)
+        data = ds.tracer
+
+        with pytest.raises(
+            ValueError, match="Must provide the widths of the boundaries"
+        ):
+            pad(data, grid, boundary="fill")
+
+    @pytest.mark.xfail(reason="This currently defaults to boundary='periodic'")
```
TomNicholas, line 183: We can change that default immediately though right? Because we never released a public pad function, so we can change it how we like in this PR

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+            [v_face_0_padded, v_face_0_addition],
+            dim="x",
+        )
+        v_face_1_expected = xr.concat(
+            [v_face_1_addition, v_face_1_padded],
+            dim="yl",
+        )
+
+        u_expected = xr.concat([u_face_0_expected, u_face_1_expected], dim="face")
+        v_expected = xr.concat([v_face_0_expected, v_face_1_expected], dim="face")
```
TomNicholas, line 934: This is almost 100 lines just to generate the expected output of a single test. Perhaps we should consider just hard-coding the result arrays instead? If we keep them as small and low-dimensional as possible?

TomNicholas: hmm but then you have parameterized it over a lot of cases... How much of this case be factored out into one big function that generates the expected output then?

xgcm/test/test_padding.py
```diff
@@ -0,0 +1,1120 @@
...
+            "X": (0, 1),
+            "Y": (1, 0),
+        },
+        {
+            "X": (0, 2),
+            "Y": (1, 0),
+        },
+    ],
+)
+class TestPaddingFaceConnection:
```
TomNicholas, line 368: Could you divide this into two classes, one for padding only involving `X` and one for the more complex cases that involve `Y` too?

TomNicholas (changes requested):
Okay done. This is a big lift, thanks @jbusecke !

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+    for axname, axis in grid.axes.items():
+        if axname not in padding.keys():
+            # Use default axis boundary for axis unless overridden
+            padding[
+                axname
+            ] = (
+                axis.boundary
+            )  # TODO: rename the axis property, or are we not keeping this on the axis level?
+
+    da_padded = da.copy()
```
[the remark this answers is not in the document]
jbusecke: Should probably do `deep=False`?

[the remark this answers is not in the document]
jbusecke: Thinking about this, I think we should have a high-level test that pads+operates on a dask array, and make sure that nothing is computed?

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+    # TODO: Do we need that information on the axis still?
+    facedim = grid._facedim
+    connections = grid._connections
+
+    if isinstance(da, dict):
+        assert len(da) == 1  # TODO raise a better error here
+        isvector = True
+        vectoraxis, da = da.popitem()
+    else:
+        isvector = False
```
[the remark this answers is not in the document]
jbusecke: I think this requires quite a bit of rewriting. Ill think about it though.

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+
+    width = _max_boundary_width(padding_width)
+
+    # TODO should i just return da if width is 0 here? I have a check further down, that I could eliminate that way.
+    max_padding_width = {k: (width, width) for k in padding_width.keys()}
+
+    # The edges of the array which are not connected, need to be padded with the
+    # 'basic' padding. This also ensures that the resulting padded faces result in
+    # the same shape. We will pad everything now and then replace the connections.
+    # That might however not be the most computational efficient way to do it.
```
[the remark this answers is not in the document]
jbusecke: Agreed. I might have some ideas about how to make this sleeker. I wonder if that would delay the release too much? Optimizing the padding seems like something that can be neatly separated from other improvements, and thus might be one of many follow up PRs? Honestly not sure what the best way forward is.

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+
+    if isvector:
+        da_partner_prepadded = _pad_basic(
+            da_partner,
+            grid,
+            max_padding_width,
+            padding,
+            fill_value,
+        )
+
```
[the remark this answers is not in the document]
jbusecke: I dont think that works. How do I get to `da_partner_prepadded` later? 
I do see the point here though: I should not actually pad the partner if I do not use it later?

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+                            concat_list = [source_slice, target_slice]
+
+                        target_da = xr.concat(
+                            concat_list,
+                            dim=target_dim,
+                            coords="minimal",
+                            compat="override",
+                            join="override",
+                        )
+                        # TODO: Can we do this with an assignment in xarray? Maybe not important yet.
```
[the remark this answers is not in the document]
jbusecke: Yes exactly. Given your suggestions (and considerations about dask tasks), it might actually be useful for us to hack together on this and find a better logic? Again could be done in a follow up PR not blocking this release.

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+        faces.append(target_da)
+
+    da_padded = xr.concat(faces, dim=facedim)
```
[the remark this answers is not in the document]
jbusecke: Because each face is potentially padded on a different edge. I believe there is definitely a more efficient way to do this, but this was the easiest for me to track/develop.

xgcm/padding.py
```diff
@@ -0,0 +1,422 @@
...
+            if stop == 0:
+                stop = None
+            else:
+                stop = -stop
+            da = da.isel({dim: slice(start, stop)})
+        return da
+
+    return _trim_expanded_padding_width(
+        da_padded, grid, padding_width, max_padding_width
+    )
```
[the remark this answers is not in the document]
jbusecke: Sure thing. Just out of curiosity, this is purely a style thing? Or are there performance/other implications that would improve?

xgcm/test/test_faceconnections.py
jbusecke: Even constant values (as I have overwritten in other tests for development) are helpful to find mistakes, but this does work too. Same as above, I think this could (and should in the interest of minimal test changes) be changed in a dedicated PR though.

xgcm/test/test_faceconnections.py
jbusecke: Same as above

xgcm/test/test_grid.py
jbusecke: I think this is ultimately related to a situation as #458.

xgcm/grid.py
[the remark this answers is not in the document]
jbusecke: I did 
```python
self._facedim = list(face_connections.keys())[0] if face_connections else None
self._connections = face_connections if face_connections else None
``` 
what do you think about that?

xgcm/grid.py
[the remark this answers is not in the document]
jbusecke: Added the type hints, but still not sure about making it public. I dont really care either way, just let me know what you think is best.

xgcm/grid.py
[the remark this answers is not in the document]
jbusecke: I have implemented a rudimentary input checking in `grid_ufunc/_check_data_input` now, which is called twice: At the most basic level within `apply_as_grid_ufunc` (where it is checking all of the possible multiple inputs+other_components) and in `grid/_1d_grid_ufunc_dispatch` where it is applied directly to the input. This implicitly also assures that we cannot pass a list/tuple of stuff there. See my comments, I wonder if we should explicitly check that to have a better error message in that case? I should probably also test that error message regardless?

xgcm/grid.py
```diff
@@ -1610,97 +1627,19 @@ def _grid_func(self, funcname, da, axis, **kwargs):
...
-
-        return padded
-
     def _1d_grid_ufunc_dispatch(
         self,
         funcname,
-        da,
+        da: Union[
+            xr.DataArray, Dict[str, xr.DataArray]
+        ],  # ? @Tom: could this accept a list of dicts?
```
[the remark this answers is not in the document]
jbusecke: This was clarified, for now we will not allow multiple data inputs to `_1d_grid_ufunc_dispatch`.

xgcm/grid_ufunc.py
jbusecke: how about `ensure_matching_other_components`?

xgcm/padding.py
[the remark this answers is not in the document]
TomNicholas: This PR is already massive so leave whatever you can until later.

xgcm/padding.py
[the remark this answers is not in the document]
TomNicholas: Oh I think I misread your code the first time actually, perhaps just ignore this suggestion.

xgcm/padding.py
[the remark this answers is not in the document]
TomNicholas: That makes sense as a first approach at least.

xgcm/test/test_faceconnections.py
TomNicholas: Definitely do test changes in a later PR

xgcm/test/test_faceconnections.py
TomNicholas: Okay sure

xgcm/grid.py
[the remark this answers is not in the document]
TomNicholas: That's better!

xgcm/grid.py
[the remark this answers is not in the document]
TomNicholas: Leave for later

xgcm/grid.py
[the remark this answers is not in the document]
TomNicholas: What you have sounds fine for now.

xgcm/test/test_padding.py
jbusecke: Should we bench that for a later PR as discussed?

xgcm/test/test_padding.py
jbusecke: All in the follow up PR I would say. Ill add it to the list.

xgcm/test/test_grid.py
jbusecke: Thinking again about this. Maybe I was not hundred percent right. Its a similar situation (as in this does not go through the ufunc logic), but why that ultimately does not preserve the coordinates is a bit weird. Probably for a future PR too though.

xgcm/test/test_faceconnections.py
jbusecke: I did not either. This is actually one of the issues that I think I need to resolve here. Lets chat about this today?

xgcm/grid.py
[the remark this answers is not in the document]
jbusecke: https://github.com/xgcm/xgcm/issues/465

xgcm/grid.py
```diff
@@ -1721,6 +1658,15 @@ def _1d_grid_ufunc_dispatch(
         if isinstance(axis, str):
             axis = [axis]
 
+        # This function is restricted to a single data input, so we need to check the input validity
+        # here early.
+        # TODO: This will fail if a sequence of inputs is passed, but not with a very helpful error
+        # TODO: message. @TOM do you think it is worth to check the type and raise another error in that case?
```
jbusecke, line 1664: @TomNicholas should I add an additional check that explicitly raises a more helpful error if multiple arguments are passed?

xgcm/test/test_faceconnections.py
jbusecke: Using `xr.testing.assert_allclose(v_out, 1)` gives me this rather unhelpful error:
```
raceback (most recent call last):
  File "<string>", line 1, in <module>
  File "[/Users/juliusbusecke/miniconda/envs/test_env_xgcm/lib/python3.9/site-packages/xarray/testing.py]()", line 34, in wrapper
    return func(*args, **kwargs)
  File "[/Users/juliusbusecke/miniconda/envs/test_env_xgcm/lib/python3.9/site-packages/xarray/testing.py]()", line 146, in assert_allclose
    assert type(a) == type(b)
AssertionError
```
Since the numpy version passes I assume that there is some issue with the coordinates? Not sure its worth digging into this here. The assertion is covering the part that matters here.

xgcm/test/test_padding.py
jbusecke: Actually this is an interesting case. I currently have 4 tests total (fill with two values, then extend, and periodic) if I would parametrize the method aswell I would end up with 6 tests - (3 methods * 2 fill_values). In most of those cases the fill_value is not even used. Is there a nifty way to only use the fill_value parametrization when method='fill'?

TomNicholas: Isn't there something similar in the xhistogram test suite where it skips a bunch of parameter combinations on a particular test?

xgcm/test/test_faceconnections.py
TomNicholas: It's complaining because it's pointing out that the DataArray is a different type to the 1, because the 1 is an int. I assume the numpy version of the assert must first attempt to turn the 1 into a numpy array.

jbusecke:
After a bit of a grueling mini refactor I think I have narrowed down the 101 ways to set a default for e.g. `boundary`. With the current version all of the tests are passing except: [test_grid.py/test_weighted_metric_multi_axis](https://github.com/jbusecke/xgcm/blob/6e85744eb787c527063084c2e77dfd0c43c8a480/xgcm/test/test_metrics_ops.py#L57-L85) specifically for the cumsum method, which will be replaced in #461 anyways.

xgcm/test/test_faceconnections.py
jbusecke: I was able to unxfail this with my latest changes

xgcm/test/test_padding.py
jbusecke: This also got sorted in the newest version

xgcm/grid.py
```diff
@@ -1307,25 +1330,31 @@ def __init__(
...
+        self._facedim = list(face_connections.keys())[0] if face_connections else None
+        self._connections = face_connections if face_connections else None
+        # TODO: I think of the face connection data as grid not axes properties, since they almost by defintion
+        # TODO: involve multiple axes. In a future PR we should remove this info from the axes
+        # TODO: but make sure to properly port the checking functionality!
+
+        # TODO: jb: I have changed my mind on boundary/fill_value though. I think these should be living in the axis.
+        # self.boundary: Dict[str, Union[str, float, int]] = boundary
+        # self.fill_value: Dict[str, Union[str, float, int]] = fill_value
+        # self.periodic: Dict[str, Union[str, float, int]] = periodic
```
jbusecke, line 1343:
```suggestion
```

xgcm/grid.py
```diff
@@ -1712,14 +1765,14 @@ def _grid_func(self, funcname, da, axis, **kwargs):
 
         if (not isinstance(axis, list)) and (not isinstance(axis, tuple)):
             axis = [axis]
-        # parse multi axis kwargs like e.g. `boundary`
-        multi_kwargs = {k: self._as_axis_kwarg_mapping(v) for k, v in kwargs.items()}
+        # # parse multi axis kwargs like e.g. `boundary`
+        # multi_kwargs = {k: self._as_axis_kwarg_mapping(v) for k, v in kwargs.items()}
```
jbusecke, line 1769:
```suggestion
```

xgcm/grid.py
```diff
@@ -1712,14 +1765,14 @@ def _grid_func(self, funcname, da, axis, **kwargs):
...
 
         out = da
         for axx in axis:
-            kwargs = {k: v[axx] for k, v in multi_kwargs.items()}
+            # kwargs = {k: v[axx] for k, v in multi_kwargs.items()}
```
jbusecke, line 1773:
```suggestion
```

xgcm/grid.py
```diff
@@ -1712,14 +1765,14 @@ def _grid_func(self, funcname, da, axis, **kwargs):
...
             ax = self.axes[axx]
-            kwargs.setdefault("boundary", ax.boundary)
+            # kwargs.setdefault("boundary", ax.boundary)
```
jbusecke, line 1775:
```suggestion
```

xgcm/grid_ufunc.py
```diff
@@ -40,6 +41,52 @@
...
+    else:
+        da = data
+    return da
+
+
+def _check_data_input(
+    data: Union[xr.DataArray, Dict[str, xr.DataArray]],
+    grid: "Grid",
+) -> Union[xr.DataArray, Dict[str, xr.DataArray]]:
+
```
jbusecke, line 58:
```suggestion
"""
Checks for valid data input (either a scalar or a single vector component). Checks types and that vector component axes actually exist
"""
```

xgcm/grid_ufunc.py
```diff
@@ -593,9 +648,22 @@ def apply_as_grid_ufunc(
...
-    if any(not isinstance(arg, xr.DataArray) for arg in args):
-        raise TypeError("All data arguments must be of type DataArray")
+    # Check data input arguments
+    args = _promote_to_sequence_and_check(args, grid)  # type: ignore
+    other_component = _promote_to_sequence_and_check(other_component, grid)
+
+    if len(other_component) == 1:
+        # Make sure that the default (None) for other_component is properly broadcasted
+        # TODO: Do we want to 'broadcast' any type of input like this or just a None?
+        other_component = [other_component[0] for i in range(len(args))]
```
jbusecke, line 660:
```suggestion
        other_component = other_component * len(args)
```

xgcm/grid_ufunc.py
```diff
@@ -593,9 +648,22 @@ def apply_as_grid_ufunc(
...
 
-    if any(not isinstance(arg, xr.DataArray) for arg in args):
-        raise TypeError("All data arguments must be of type DataArray")
+    # Check data input arguments
+    args = _promote_to_sequence_and_check(args, grid)  # type: ignore
+    other_component = _promote_to_sequence_and_check(other_component, grid)
+
+    if len(other_component) == 1:
+        # Make sure that the default (None) for other_component is properly broadcasted
+        # TODO: Do we want to 'broadcast' any type of input like this or just a None?
```
jbusecke, line 659:
```suggestion
    if len(other_component) == 1 and other_component[0] is None:
        # Make sure that the default (None) for other_component is properly broadcasted
```

xgcm/test/test_faceconnections.py
```diff
@@ -130,6 +141,9 @@ def test_create_periodic_grid(ds):
         assert connect_right[2] is False
 
 
+@pytest.mark.xfail(
+    reason="Test is too low level. This functionality has been replaced by grid_ufunc internals"
```
jbusecke, line 145:
```suggestion
    reason="Test is too low level. This functionality has been replaced by grid_ufunc internals",
    strict=True
```

xgcm/test/test_faceconnections.py
```diff
@@ -187,12 +200,15 @@ def test_diff_interp_connected_grid_x_to_x(ds, ds_face_connections_x_to_x):
...
-@pytest.mark.xfail(reason="connected grids not implemented for grid ufunc refactor yet")
 def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
     # one face connection, rotated
     grid = Grid(ds, face_connections=ds_face_connections_x_to_y)
 
     diff_y = grid.diff(ds.data_c, "Y", boundary="fill")
+    # ! this will only apply `fill` to the Y axis, but for face
+    # ! connections across axes we need to currently pad along the other dimension too.
+    # ! This causes an issue with periodic (aka wrap padding) and the coordinates in xarray
+    # ! (see https://github.com/pydata/xarray/issues/6425)
```
jbusecke, line 211:
```suggestion
```

xgcm/test/test_faceconnections.py
```diff
@@ -207,23 +223,58 @@ def test_diff_interp_connected_grid_x_to_y(ds, ds_face_connections_x_to_y):
         interp_y.data[1, 0, :].ravel(),
         0.5 * (ds.data_c.data[1, 0, :].ravel() + ds.data_c.data[0, ::-1, -1].ravel()),
     )
-
     # TODO: checking all the other boundaries
 
 
-@pytest.mark.xfail(reason="connected grids not implemented for grid ufunc refactor yet")
+# TODO: Relaease the periodic test here once we unified the API with padding.
```
jbusecke, line 229:
```suggestion
```

xgcm/test/test_grid.py
```diff
@@ -954,5 +972,146 @@ def test_boundary_global_input(funcname, boundary, fill_value):
...
+        grid = Grid(ds)
+        msg = "Dictionary inputs must have a DataArray as value. Got .*?"
+        with pytest.raises(
+            TypeError,
+            match=msg,
+        ):
+            grid.diff({"X": "not_a_dataarray"}, "X")
+
+    def test_wrong_axis_vector_input_axis(self):
+        # TODO: @Tom is this explicitly tested elsewhere? Then ill remove it here.
```
jbusecke, line 1010:
```suggestion
```

TomNicholas:
Thanks for the heroic effort here @jbusecke !

Issue #414, opened 2022-01-10
Plans for upcoming breaking changes (maybe v1.0.0?)


jbusecke:
I just wanted to create a main issue to collect/discuss things which will be breaking changes on top of the (so far mostly backwards compatible) changes in @TomNicholas `grid_ufunc_refactor` branch (#344). 

As outlined in #409, we will:
- Streamline and clarify the difference between padding (extending array boundaries) and boundary physical boundary conditions (not implemented yet, but will be possible in the future thanks to the grid ufuncs, #324 and #293 are related)
    - Rename `boundary` to  `padding` kwarg, and enable to set this only on the grid object, not the methods.
    - fill_value default being changed from zero to NaN
    - extrapolate functionality might get dropped. It was never consistently working and is more suited for actual physical boundaries
    - `periodic` kwarg is being dropped, and implemented as option for `padding`, as originally suggested in #195.
- Axis is going to go entirely in favour of always using grid. This has several important ramifications:
    - We need to migrate the transform logic to work only on the grid object (and could possibly add support for #313 in the process)
    - Removing the axis methods will break all of the autogenerate modules functionality. I believe we can implement that functionality much easier with the new padding module (#408) and can implement better defaults (#79)
    - More complete treatment for complex grid topology for vectors in #410 will likely eliminate the need for helper methods like interp_2d_vector

TomNicholas:
> extrapolate functionality might get dropped. It was never consistently working and is more suited for actual physical boundaries

Are we definitely doing this then? Because there are quite a few xfailed tests that use extrapolate in `grid_ufunc_refactor_project` right now. Can I just remove them?

TomNicholas:
> Are we definitely doing this then? Because there are quite a few xfailed tests that use extrapolate in grid_ufunc_refactor_project right now. Can I just remove them?

We are removing this in #470.

Also 
[ ] Rewrite boundary conditions docs eventually once there is a distinction between internal and external boundaries.

hdrake:
Closing as superseded. This thread's role — tracking the breaking changes for the first stable release — has been taken over by the **Road to xgcm 1.0.0** roadmap (#733), which explicitly absorbs it and is now the single source of truth for 1.0 scope. Every item here is either done or tracked there:

- `boundary` → `padding` rename → #696 / #678
- `periodic` deprecation → #746
- `fill_value` default → `nan` → tracked in §A of #733
- `keep_coords` removal → done in #745
[...]

PR #464, opened 2022-04-06
Ufunc examples docs

TomNicholas:
- [x] Helps closs the "how-to" part of #423
 - [x] Passes `pre-commit run --all-files`

Builds on top of #431, adding more complicated examples of defining grid ufuncs, such as:
 - [x] divergence
 - [ ] gradient
 - [ ] curl
 - [ ] advection

PR #470, opened 2022-04-18
Remove extrapolate

TomNicholas:
Removes all mention of extrapolate from the docs on boundary conditions, because extrapolate does not work with the refactor (because xarray.pad doesn't support it). We could potentially add extraplate back in later.

This should fix the docs failures.

Also moves the docs on boundary conditions to it's own new page, instead of being at the bottom of the page on simple grids.

Part of #414.


TomNicholas:
So far this just removes extrapolate from the docs, allowing them to pass. 

Should I also disable extrapolate as a valid boundary kwarg and remove the tests which use it @jbusecke ?

jbusecke:
Difficult question. We do not have an ideal solution here I believe. The two options are:
- Remove from docs but leave the functionality unchanged, which will not introduce breaking changes in the code, but result in an inconsistent docs/code relationship
- Remove the `extrapolate` option early (e.g. now as compared to later), and break backwards compatibility.

I vote for the first option, since this release was supposed to be the last backwards compatible one, before we introduce a bunch of breaking changes #414.

TomNicholas:
> raise when extrapolate is supplied to either Axis/Grid method and xfail speficially only the parameterized test that invoke this option.

That's now done

xgcm/test/test_grid.py
```diff
@@ -701,29 +710,32 @@ def test_multi_axis_input(all_datasets, func, periodic, boundary):
...
+    eager_result = grid_method(ds.tracer, "X", boundary=boundary)
 
     ds = ds.chunk({"xt": 1, "yt": 1, "time": 1, "zt": 1})
     grid = Grid(ds, coords=coords)
-    dask_result = grid.diff(ds.tracer, "X").compute()
+    grid_method = getattr(grid, func)
+    dask_result = grid_method(ds.tracer, "X", boundary=boundary).compute()
 
-    xr.testing.assert_identical(dask_result, eager_result)
+    xr.testing.assert_allclose(dask_result, eager_result)
```
TomNicholas, line 738: I realised this test was supposed to test many different method but it was only testing `diff` - now fixed.

doc/boundary_conditions.rst
```diff
@@ -0,0 +1,131 @@
...
+            "x_c": (
+                [
+                    "x_c",
+                ],
+                np.arange(1, 10),
+                {"axis": "X"},
+            ),
+            "x_g": (
+                [
+                    "x_g",
```
jbusecke, line 23: See my comment in [#464](https://github.com/xgcm/xgcm/pull/464#discussion_r852342050). I think this can be made more compact.

jbusecke (approved):
LGTM. Do we note clearly enough in #414 that this section needs to be rewritten once we introduce padding as a term (or array vs physical boundary).

doc/boundary_conditions.rst
TomNicholas: Let's just deal with all docs formatting in one go in another PR.

PR #475, opened 2022-04-19
Vector padding docs

### LINKED ISSUE #410: Extended functionality for vector data with complex grid topologies

jbusecke:
- [ ] Closes #410

This PR adds docs to explain the new 'vector_syntax'. Still very rough and needs pictures/examples, but I would appreciate an early review re the wording.


jbusecke:
Closed in favor of #476

jbusecke:
Superceded by #476

PR #479, opened 2022-04-20
Grid ufunc refactor

### LINKED ISSUE #344: Proposal: grid_ufunc

TomNicholas:
The **big** PR to implement the internal refactoring suggested by @rabernat in #344. After this we won't need a parallel `grid_ufunc_refactor_project` branch any more.

Means that `Grid` no longer uses `neighbour_binary_func` at all (so `_grid_func` has been deleted). However `Axis` still does use  `neighbour_binary_func` - this will be removed in the next version (via #405).

 - [x] Closes #344
 - [x] Tests added (the whole of `test_grid_ufunc.py`)


TomNicholas:
Something is going wrong with the pre-commit on the repo - it all passes locally for me, but on the CI mypy fails with

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

xgcm/padding.py:395: error: Item "None" of "Optional[Dict[str, Tuple[int, int]]]" has no attribute "values"
xgcm/grid_ufunc.py:413: error: Unexpected keyword argument "include_extras" for "get_type_hints"
/home/runner/.cache/pre-commit/repoyno4hcj9/py_env-python3.8/lib/python3.8/site-packages/mypy/typeshed/stdlib/typing.pyi:633: note: "get_type_hints" defined here
xgcm/test/test_grid_ufunc.py:2: error: Module "typing" has no attribute "Annotated"
Found 3 errors in 3 files (checked 24 source files)
```
(which shouldn't be an error for python 3.9+), and black fails with all sorts of tiny and pointless changes.

TomNicholas:
> Something is going wrong with the pre-commit on the repo

mypy error fixed by #480, but the black error is still there

xgcm/test/test_metrics_ops.py
```diff
@@ -6,8 +6,21 @@
 from xgcm.grid import Grid
 from xgcm.test.datasets import datasets_grid_metric
 
+GRID_UFUNC_REFACTOR_ERR = (
+    "Metrics have not yet been implemented for the grid ufunc refactor"
+)
+
```
jbusecke, line 12:
```suggestion
```

jbusecke: This does not seem to be used anywhere?

jbusecke (commented):
Couple of small things to check. Otherwise I am ok with this! Cannot wait to merge this.

xgcm/grid.py
```diff
@@ -2263,42 +2586,53 @@ def transform(self, da, axis, target, **kwargs):
...
-        ----------
-        vector : dict
-            A dictionary with two entries. Keys are axis names, values are
-            vector components along each axis.
-        to : {'center', 'left', 'right', 'inner', 'outer'}
-            The direction in which to shift the array. If not specified,
-            default will be used.
-        boundary : {None, 'fill', 'extend'}
-            A flag indicating how to handle boundaries:
+def _select_grid_ufunc(funcname, signature: _GridUFuncSignature, module, **kwargs):
```
rabernat, line 2590: Just noting that this function is a gorgeous little piece of python.

rabernat (approved):
Looking at everything all together in one PR makes me realize what a staggering piece of work this has been. 👏  Tom and Julius for pushing this through.

The overall code architecture is brilliant. The tests are incredibly comprehensive. The docs are detailed and useful. I honestly looked for nits to pick and didn't find any.

🚀

setup.py
```diff
@@ -28,7 +28,7 @@
 extras_require["dev"] = extras_require["complete"] + [
     "pytest",
     "pytest-cov",
-    "flake8",
+    "scipy" "flake8",
```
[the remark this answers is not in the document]
TomNicholas:
```suggestion
    "scipy",
    "flake8",
```

PR #483, opened 2022-04-20
Implementation of transform with logarithmic interpolation (method='log')

jthielen:
As discussed way back in https://github.com/xgcm/xgcm/issues/222#issuecomment-731267505 and https://github.com/xgcm/xgcm/issues/222#issuecomment-731308398, it would be handy for atmospheric datasets to have a `method='log'` option for vertical coordinate interpolation. https://github.com/ncar-xdev/xwrf/issues/69 finally prompted me to return to this, so here is a PR for it! Only one test (based on some values from a WRF file and output from `metpy.interpolate.log_interpolate_1d`) has been added so far; please let me know if this is sufficient or not!

 - [x] Tests added


xgcm/grid.py
```diff
@@ -2242,12 +2252,13 @@ def transform(self, da, axis, target, **kwargs):
...
-            are masked with nan, by default True. Only applies to 'linear' method.
+            are masked with nan, by default True. Only applies to 'linear' and
+            'log' methods.
         bypass_checks : bool, optional
-            Only applies for `method='linear'`.
+            Only applies for `method='linear'` and `method='log'`.
             Option to bypass logic to flip data if monotonically decreasing along the axis.
             This will improve performance if True, but the user needs to ensure that values
-            are increasing alon the axis.
+            are increasing along the axis.
```
jbusecke, line 2261: Thanks for catching all of those embarrassing typos haha.

xgcm/transform.py
```diff
@@ -62,12 +67,19 @@ def interp_1d_linear(
...
+        Apply a logarithmic transform to theta and target_theta_levels so that the
+        interpolation is done in logarithmic, rather than linear, space. In turn, theta
+        must be positive. Defaults to False.
 
     Returns
     -------
     phi_interp : array
         Array of shape (..., m) of phi interpolated to theta isosurfaces.
     """
+    if logarithmic:
```
jbusecke, line 80: Love it. There were literally only 4 lines needed to implement this option! Everything else are docstrings!

jbusecke (changes requested):
This looks fantastic @jthielen. Thanks for putting this in! I honestly see nothing wrong with this PR and I could merge this rn. But could you maybe add an example for this method to the docs? I think that would be very useful to demonstrate this method both for a toy example and maybe some real world output?

jthielen:
> But could you maybe add an example for this method to the docs? I think that would be very useful to demonstrate this method both for a toy example and maybe some real world output?

Sorry for the delay, but yes, I can definitely do so. However, given this method's relative uniqueness to pressure coordinates (as far as I'm aware based on experience), I wanted to ask a couple questions before putting together those added examples:

- would it be okay to add "Logarithmic Interpolation" as new subsections to https://xgcm.readthedocs.io/en/latest/transform.html that come *after* Conservative Interpolation, despite being more closely related to Linear Interpolation, simply because of the need to use different data?
- I'm having a hard time finding non-post-processed (i.e., in native model vertical coordinates rather than isobaric) atmospheric data on a public data catalog for a real-world example. Do you have any suggested workarounds?
    - Skip for now? 
    - Add a docs dependency on `xwrf` and use one of its sample files?
    - Upload a suitable file somewhere (where?) and point to a static URL?
    - Something else?

jbusecke:
> would it be okay to add "Logarithmic Interpolation" as new subsections to https://xgcm.readthedocs.io/en/latest/transform.html that come after Conservative Interpolation, despite being more closely related to Linear Interpolation, simply because of the need to use different data?

Absolutely! This sounds great.

> I'm having a hard time finding non-post-processed (i.e., in native model vertical coordinates rather than isobaric) atmospheric data on a public data catalog for a real-world example. Do you have any suggested workarounds?

I would prefer to keep this as light as possible. Do you think you could demonstrate this on some synthetic data? 

> Add a docs dependency on xwrf and use one of its sample files?

I would not want to add this dependency to xgcm directly, but we have this [example-repo](https://github.com/xgcm/xgcm-examples), where we store more long-form examples (and we can link them in the xgcm main docs too: see [this section](https://xgcm.readthedocs.io/en/latest/xgcm-examples/01_eccov4.html) and the [source notebook](https://github.com/xgcm/xgcm-examples/blob/master/01_eccov4.ipynb)). Very happy to help with setting that up, but I think that can be nicely separated from this PR and the actual implementation.

dcherian:
:+1: for using a synthetic dataset. We can always link to xwrf for a more realistic example...

jthielen:
I've added the following logarithmic interpolation examples to the `transform.ipynb` docs notebook:
- 1D toy data (same as test)
- 3D synthetic data loosely based on the Sanders 1971 Analytic Model (I modified to use sigma coordinate and sloping terrain)

I'm not sure how the doc pages are executed/rendered, but if they are simply rendered without being re-run, then I ran into an error with an existing portion of the notebook that I'd request assistance on (as I am not familiar with the ocean model dataset in use). The first cell in the "Realistic Data Example > Conservative Interpolation" subsection raises the following error:

<details>

```
Input In [27], in <cell line: 1>()
----> 1 grid = Grid(ds, coords={'Z': {'center': 'lev', 'outer': 'level_outer'},
      2                         'X': {'center': 'x', 'right': 'x_c'},
      3                         'Y': {'center': 'y', 'right': 'y_c'}
      4                         },
      5             periodic=False,
      6             )
      7 grid

File ~/develop/xgcm/xgcm/grid.py:1271, in Grid.__init__(self, ds, check_dims, periodic, default_shifts, face_connections, coords, metrics, boundary, fill_value)
   1269 for pos, dim in positions.items():
   1270     if not (dim in ds.variables or dim in ds.dims):
-> 1271         raise ValueError(
   1272             f"Could not find dimension `{dim}` (for the `{pos}` position on axis `{axis}`) in input dataset."
   1273         )
   1274     if dim not in ds.dims:
   1275         raise ValueError(
   1276             f"Input `{dim}` (for the `{pos}` position on axis `{axis}`) is not a dimension in the input datasets `ds`."
   1277         )

```

</details>

```
ValueError: Could not find dimension `x_c` (for the `right` position on axis `X`) in input dataset.
```

lpilz:
Hi everybody, just the gentlest of nudges from my side ;) Keen to get this over the finish line asap for our upcoming release of xwrf :)

jbusecke:
SOOO sorry for forgetting about this. If you could add a short item in the `doc/whats-new.rst`, I can merge this right away!

rabernat:
Perhaps we (maintainers) could just push the whats-new update directly? @jthielen has already responded to our reviews twice. 😉

jthielen:
> SOOO sorry for forgetting about this. If you could add a short item in the `doc/whats-new.rst`, I can merge this right away!

> Perhaps we (maintainers) could just push the whats-new update directly? @jthielen has already responded to our reviews twice. wink

No worries, whats-new entry has been added and the branch has been rebased on master!

jbusecke (approved):

Issue #109, opened 2018-07-21
Support SGRID conventions


rabernat:
Since we started working on xgcm, a new convention for staggered grid netCDF metadata has emerged:
http://sgrid.github.io/sgrid
(We learned about this thanks to @vrx- in #108.)

Reading through the document, I am pleased to see a huge amount of conceptual overlap between the sgrid and xgcm data models. This is a good thing! It means that there is one clear solution to this problem, and multiple independent groups have more-or-less converged on it.

We need to update xgcm to understand sgrid conventions (like it currently does with comodo conventions) so it can automatically build a Grid object based on just the sgrid metadata. This will require a few changes under the hood (see #108), bur should be doable. 

I wanted to ping some sgrid developers (@hrajagers, @rsignell-usgs) to get their thoughts on how we could work together. Particularly useful would be if you could point us towards some example netCDF files from real models that implement the conventions.

vrx-:
The [Texas-Louisiana Regional ROMS Shelf Model](http://barataria.tamu.edu:8080/thredds/catalog.html?dataset=txla_hindcast_agg) ( @hetland)

hrajagers:
Hi Ryan, Sorry for the slow ping. Following the success of the [UGRID](http://ugrid-conventions.github.io/ugrid-conventions/) we drafted SGRID as a lightweight meta data layer to add staggered grid information to NetCDF files using a syntax very similar to that of UGRID. Although we received several positive reactions, the uptake is slow. It's time for some tools to start leveraging the additional meta-data, so I'm glad you're reaching out.

We've added the meta-data to map-data of Delft3D in NetCDF. An example can be found at
http://opendap.deltares.nl/thredds/catalog/opendap/deltares/Delft3D/netcdf_example_files/catalog.html?dataset=varopendap/deltares/Delft3D/netcdf_example_files/trim-westernscheldt_sph.nc
Unfortunately that example file still contains some -999 values for the X and Y coordinates which should be missing values.

rsignell-usgs:
The [USGS COAWST forecast results](https://gamone.whoi.edu/thredds/catalog/coawst_4/use/fmrc/catalog.html?dataset=coawst_4/use/fmrc/coawst_4_use_best.ncd) are also SGRID compliant.

hrajagers:
[...]

PR #502, opened 2022-05-20
Improve error message around invalid axis shifts (and slight refactoring)

dorchard:
If an invalid axis shift is specified for `c_grid_axis_shift`, this change explains the correct range of values, e.g: given code

```
g = Grid(xr.Dataset(
        coords={
            "x_c": (
                ["x_c"],
                np.arange(1, 10),
                {"axis": "X"},
            ),
            "x_g": (
                ["x_g"],
                np.arange(0.5, 9),
                {"axis": "X", "c_grid_axis_shift": 1},
            ),
        }
    ))
```

The error is now

```
ValueError: Coordinate x_g has invalid `c_grid_axis_shift` attribute `1.0`. `c_grid_axis_shift` must be one of: 0.5, -0.5
```
I did some slight refactoring to the code around here to pull out number constants to the top level.

### DIFF
--- a/xgcm/comodo.py
+++ b/xgcm/comodo.py
@@ -1,5 +1,10 @@
 from collections import OrderedDict
 
+# Representation of axis shifts
+axis_shift_left  =  0.5
+axis_shift_right = -0.5
+# Characeterizes valid shifts only
+valid_axis_shifts = [axis_shift_left, axis_shift_right]
 
 def assert_valid_comodo(ds):
     """Verify that the dataset meets comodo conventions
@@ -102,7 +107,7 @@ def get_axis_positions_and_coords(ds, axis_name):
             axis_coords["outer"] = name
         elif clen == axis_len - 1:
             axis_coords["inner"] = name
-        elif shift == -0.5:
+        elif shift == axis_shift_left:
             if clen == axis_len:
                 axis_coords["left"] = name
             else:
@@ -110,7 +115,7 @@ def get_axis_positions_and_coords(ds, axis_name):
                     "Left coordinate %s has incompatible "
                     "length %g (axis_len=%g)" % (name, clen, axis_len)
                 )
-        elif shift == 0.5:
+        elif shift == axis_shift_right:
             if clen == axis_len:
                 axis_coords["right"] = name
             else:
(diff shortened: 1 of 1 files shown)


dorchard:
Sorry - I've borked the test procedure. I will get this fixed soon and update the PR!

xgcm/comodo.py
```diff
@@ -119,12 +124,21 @@ def _maybe_fix_type(attr):
...
+                # string representing valid axis shifts
+                valids = str(valid_axis_shifts)[1:-1]
 
+                raise ValueError(
+                    "Coordinate %s has invalid "
+                    "`c_grid_axis_shift` attribute `%s`. "
+                    "`c_grid_axis_shift` must be one of: %s" % (name, repr(shift), valids)
+                )
+            else:
+                raise ValueError(
```
jbusecke, line 137: I think the [comodo spec](https://web.archive.org/web/20160417032300/http://pycomodo.forge.imag.fr/norm.html#attribute-c-grid-axis-shift) inteprets a 'missing' `c_grid_axis_shift` as `0` or `centered`. So this might be too strict of a requirement?

rabernat:
Thanks a lot @dorchard for your contribution here! 🙌  We are very lucky to have your eyes on xgcm. This PR seems like a nice improvment.

Overall I find these whole area of "comodo conventions" to be a bit hard to navigate because the original website for the conventions has gone offline! AFAIK we are effectively the only ones maintaining this convention. 🙃  A more sustainable path will be #109.

xgcm/comodo.py
dorchard: I've added this to list. The code doesn't actually prevent 0 being used, but better to have this in the error message, so I've included it in the update.

dorchard:
> Thanks a lot @dorchard for your contribution here! 🙌 We are very lucky to have your eyes on xgcm. This PR seems like a nice improvment.
> 
> Overall I find these whole area of "comodo conventions" to be a bit hard to navigate because the original website for the conventions has gone offline! AFAIK we are effectively the only ones maintaining this convention. 🙃 A more sustainable path will be #109.

Good to know!
I fixed an issue and everything seems to pass locally now.

raehik:
The CI failure might be to do with the Python version Conda downloads with the `ci/environment-upstream-dev.yml` environment file - it downloads 3.7, but xarray needs 3.8. I had a similar issue locally, and I fixed it by specifying the Python version. Having trouble confirming locally because of apparent caching.

jbusecke:
Oh that should definitely not happen. Xgcm itself needs [>=3.9](https://github.com/xgcm/xgcm/blob/41b82bee8d18b8ab5a9daa81b2e12f4a59060a50/setup.py#L50-L51).
I have explicitly specified python 3.10 in #521, which might help here.

PR #513, opened 2022-06-14
Add @jthielen

jbusecke:
@jthielen could you check the spelling and add your ORCID if you have one?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -60,6 +60,7 @@ authors:
   orcid: "https://orcid.org/0000-0002-8654-6009"
 - family-names: "Thielen"
   given-names: "Jon"
+  orcid: "https://orcid.org/0000-0002-5479-0189"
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -58,6 +58,8 @@ authors:
 - family-names: "Uchida"
   given-names: "Takaya"
   orcid: "https://orcid.org/0000-0002-8654-6009"
+- family-names: "Thielen"
+  given-names: "Jon"
```
jthielen, line 62:
```suggestion
  given-names: "Jon"
  orcid: "https://orcid.org/0000-0002-5479-0189"
```

jthielen (changes requested):
Sounds good! (Sorry about the double reply, mistakenly first added as comment rather than review to be able to apply the suggestion)

PR #516, opened 2022-06-20
Fix missing link references in docs

raehik:
Tweak to add missing link references. Appears that when some content was moved between files, some references weren't copied through. I had a look for other instances, but this was all I found.

PR #517, opened 2022-06-22
Unpin ipython

jbusecke:
We saw some issues pulling from the PR in #516. Hopefully this works.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -14,6 +14,7 @@ dependencies:
   - matplotlib
   - sphinx_rtd_theme
   - ipykernel
+  - ipython
 #  - python-graphviz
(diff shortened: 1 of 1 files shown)


TomNicholas:
> the code produces an error

This is expected because you're now importing a version of ipython that doesn't have my fix in it. Ipython haven't actually released since my fix was merged (9 days ago), so either you need to wait for that and pin to the new version (recommended) or pin to ipython master.

jbusecke:
Oh duh, that makes a lot of sense. Let me pin to the main branch

jbusecke:
Ok this seems to work. Thanks for the pointers @TomNicholas, Ill merge this now.

PR #524, opened 2022-07-18
fix formatting in comodo.py

dorchard:
I'm sorry, I just realised that my previous PR (https://github.com/xgcm/xgcm/pull/502) didn't have the right formatting generated from `black`- this is fixed here.


jbusecke (approved):
Ah excellent. Thanks for adding this @dorchard.

PR #538, opened 2022-09-07
require ipython version which fixes decorator issue

### LINKED ISSUE #455: [BUG] Decorated lines don't show up in docs
Opening this to keep track of this very irritating [bug in ipython's sphinx extension](https://github.com/ipython/ipython/issues/13531).

The bug basically strips all lines beginning with `@` from ipython code before building the sphinx docs page, which is a problem when you're [trying to document a decorator](https://github.com/xgcm/xgcm/pull/431).

We either have to fix this upstream in ipython (which I am trying to do) or just demote the relevant code blocks to be static, pre-written inputs and outputs, instead of real code that runs each time the docs are created.

TomNicholas:
Ipython finally released a new version, which means we no longer have to pin our docs build to their master branch in order to fix #455.

 - [x] Closes #455

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -14,6 +14,7 @@ dependencies:
   - matplotlib
   - sphinx_rtd_theme
   - ipykernel
+  - ipython>=8.5.0
 #  - python-graphviz
   - pip:
     - nbsphinx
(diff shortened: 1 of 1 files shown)


jbusecke (approved):
Awesome. Thanks @TomNicholas. Merging this now.

PR #544, opened 2022-10-04
Remove release reminder

jbusecke:
As suggested by @TomNicholas, I am removing the release reminder, to avoid spam


jbusecke:
@TomNicholas can we merge this?

PR #557, opened 2022-11-15
Remove computation methods on Axis class

### LINKED ISSUE #405: Lets get rid of `Axis`

TomNicholas:
ToDo list:
- [x] Fix `transform` WIP in https://github.com/TomNicholas/xgcm/pull/1
- [x] Fix `default_shifts`
- [x] Rip out redundant tests
- [x] Fix typing errors (`TypeError: unsupported operand type(s) for |: 'type' and '_GenericAlias'`) for python 3.9
- ~~[ ] Fix numba issues for core-dependency ci~~ (We are going to ignore this and track the dev-upstream status for a bit of time to see if things work themselves out)
- ~~[ ] Remove autogenerate docs~~ (Postponed to a fix for #564)

---
- [x] Replaces #415
- [x] Closes #405
- [x] Tests added


TomNicholas:
Next step is to dive in to rewrite `Grid` around just containing a list of `Axis` instances

TomNicholas:
@rabernat we were wondering if there is any possible case for an Axis without a single coordinate? There are old tests for this but it seems like something we could just forbid?

jbusecke:
image

Now we are getting there! Whoop

jbusecke:
Ok after fixing the CI in #577 I hope this will actually pass (after I removed `autogenerate.py` and `duck_array_ops.py` (+respective tests) entirely.

jbusecke:
Found a minor issue with the transform refactor (checking for optional numba dependency), which is hopefully sorted now.

xgcm/axis.py
```diff
@@ -0,0 +1,208 @@
...
+            if pos in default_shifts:
+                self._default_shifts[pos] = default_shifts[pos]
+            else:
+                for possible_shift in FALLBACK_SHIFTS[pos]:
+                    if possible_shift in self.coords:
+                        self._default_shifts[pos] = possible_shift
+                        break
+
+            if pos in self._default_shifts and self._default_shifts[pos] == pos:
+                # TODO stricter checking? e.g. non-adjacent positions?
```
jbusecke, line 111: Not sure I still understand this comment.

xgcm/grid.py
```diff
@@ -4,1199 +4,57 @@
...
-    Iterable,
-    List,
-    Mapping,
-    Optional,
-    Sequence,
-    Tuple,
-    Union,
-)
-
-# Only need this until python 3.8
```
jbusecke, line 46: Does this mean we still need it for py3.8?

xgcm/grid.py
```diff
@@ -4,1199 +4,57 @@
...
+    from typing_extensions import Literal  # type: ignore
 
 
-_XGCM_BOUNDARY_KWARG_TO_XARRAY_PAD_KWARG = {
-    "periodic": "wrap",
-    "fill": "constant",
-    "extend": "edge",
-}
+def _maybe_promote_str_to_list(a):
+    # TODO: improve this
```
jbusecke, line 53: How do we need to improve this?

xgcm/grid_ufunc.py
```diff
@@ -31,7 +31,9 @@
 #   - Specify xgcm.Axis name and "axis positions" instead of numpy axes as (ax_name:ax_pos)
 
 _AXIS_NAME = r"\w+"
-_AXIS_POSITION = "(?:center|left|right|inner|outer)"
+_AXIS_POSITION = (
+    "(?:center|left|right|inner|outer)"  # TODO use VALID_POSITION_NAMES here instead
```
jbusecke, line 35: Can/Should we do that in this PR?

xgcm/test/test_axis.py
```diff
@@ -0,0 +1,117 @@
...
+        coords={
+            "XG": (
+                ["XG"],
+                2 * np.pi / N * np.arange(0, N),
+                {"axis": "X", "c_grid_axis_shift": -0.5},
+            ),
+            "XC": (["XC"], 2 * np.pi / N * (np.arange(0, N) + 0.5), {"axis": "X"}),
+        },
+    )
+    return ds
```
jbusecke, line 25: Can we just get this from the dictionary in datasets.py?

xgcm/test/test_axis.py
```diff
@@ -0,0 +1,117 @@
...
+    ds = periodic_1d()
+    axis = Axis(name="X", ds=ds, coords={"center": "XC", "left": "XG"})
+
+    da = ds["data_g"]
+    num = axis._get_axis_dim_num(da)
+    assert num == da.get_axis_num("XG")
+
+
+def test_assert_axes_equal():
+    ...
```
jbusecke, line 117: Should we still add an axis.equals()?

xgcm/test/test_faceconnections.py
```diff
@@ -126,38 +126,6 @@ def cubed_sphere_connections():
...
-    np.testing.assert_allclose(right_edge_data_x, ds.data_c.isel(x=0).data[:, None])
-
-    yaxis = grid.axes["Y"]
-    left_edge_data_y = yaxis._get_edge_data(ds.data_c)
-    np.testing.assert_allclose(left_edge_data_y, ds.data_c.isel(y=-1).data[None, :])
-    right_edge_data_y = yaxis._get_edge_data(ds.data_c, is_left_edge=False)
-    np.testing.assert_allclose(right_edge_data_y, ds.data_c.isel(y=0).data[None, :])
-
-
 def test_connection_errors(ds):
```
jbusecke, line 129: Not sure we should leave this just as a passing test. Seems like this is an older test. Mayne @rabernat remembers what this was supposed to test? Probably something with proper error handling when supplying face connections?

xgcm/test/test_grid.py
```diff
@@ -14,600 +14,123 @@
...
     assert ax1.name == ax2.name
     for pos, coord in ax1.coords.items():
         assert pos in ax2.coords
         assert coord == ax2.coords[pos]
     assert ax1._periodic == ax2._periodic
     assert ax1._default_shifts == ax2._default_shifts
-    assert ax1._facedim == ax2._facedim
     # TODO: make this work...
+    # assert ax1._facedim == ax2._facedim
     # assert ax1._connections == ax2._connections
```
jbusecke, line 26:
```suggestion
```

I guess this is what we had in mind with an `Axis.equals()` method. Either way this information is not stored in the axis anymore.

jbusecke (commented):
Mostly noted things we still need to clean up. Great progress on this @TomNicholas. We are very close here

jbusecke:
Yoinks! So the python 3.9 version has this weird `TypeError: unsupported operand type(s) for |: 'type' and '_GenericAlias'` which still needs to be fixed, but this is amazing!
image

xgcm/axis.py
TomNicholas: Nor me.

xgcm/grid.py
TomNicholas: I think for 3.8 onwards we can just use `typing.Literal`.

TomNicholas: I've changed it.

xgcm/grid_ufunc.py
TomNicholas: I think we can leave it for later

xgcm/test/test_grid.py
TomNicholas: Was the idea that we would have a `Grid.equals()`, that itself called `Axis.equals()`? Why were we going to have a grid.equals again? 

I think if we can't remember its fine to just leave equals out until it comes up again.

jbusecke: I believe you remember that correctly. This would be useful, but is also pretty separable from this PR. I am going to open an issue to track this, and then we can remove the commented lines here and move on.

jbusecke: #579

xgcm/axis.py
jbusecke: Then lets remove it I guess?

xgcm/test/test_axis.py
TomNicholas: Done

xgcm/grid.py
TomNicholas: I don't remember... `_maybe_promote_str_to_list` is only used within `set_metrics`, and seems conceptually similar to our `_map_kwargs_over_axes`, so maybe could be refactored to use that. This is v low priority though.

xgcm/test/test_faceconnections.py
TomNicholas: This question seems unconnected to this specific PR

jbusecke:
Getting close here. What are your thoughts on the upstream env? Over in #577 Justus [suggested](https://github.com/xgcm/xgcm/pull/577#issuecomment-1406439959) to pin numpy, but I am not sure we need to do that. I would suggest we ignore this failure for now, and keep an eye on it. 

We also need to remove the docs pertaining to autogenerate, since we ripped that out for good.
I played with cf-xarray a bit and am fairly certain that we can use it to implement #564. I am open to leave the docs refactor up to a dedicated PR for that, given that the autogenerate functionality was semi-broken in the last release anyways. Thoughts @TomNicholas? The reason to do this would be to make way for #559 asap, and work on #564 independently.

TomNicholas:
@jbusecke the biggest issue here is that the CI is doing something wild. Everything passes for me locally now (including mypy), and I'm using a local conda environment copied from xgcm's CI.

This todo up top might be related? 
> Fix numba issues for core-dependency ci

Up top it auto mentions removing autogenerate docs.

doc/whats-new.rst
```diff
@@ -14,6 +14,11 @@ New Features
 Breaking Changes
 ~~~~~~~~~~~~~~~~
 
+- All computation methods on the :py:class:`xgcm.Axis` class have been removed, in favour of using the corresponding
+  methods on the :py:class:`xgcm.Grid` object. The :py:class:`xgcm.Axis` class has also been removed from public API.
+  (:issue:`405`, :pull:`557`).
+  By `Thomas Nicholas <https://github.com/tomnicholas>`_.
+
```
jbusecke, line 21:
```suggestion
- All computation methods on the :py:class:`xgcm.Axis` class have been removed, in favour of using the corresponding
  methods on the :py:class:`xgcm.Grid` object. The :py:class:`xgcm.Axis` class has also been removed from public API.
  (:issue:`405`, :pull:`557`).
  By `Thomas Nicholas <https://github.com/tomnicholas>`_.

- All functionality for generating c-grid dimensions on incomplete datasets via :py:meth:`xgcm.Grid.autogenerate`,  was removed (:pull:`557`). 
   By `Julius Busecke <https://github.com/jbusecke>`_.
```

Issue #555, opened 2022-11-15
[FEATURE] Rename `coords` input to `dims`


jbusecke:
There have been multiple instances of confusion () around what the inputs to `coords` should be. 
These should in fact be xarray dimensions (which can have, but do not require coordinate values). I propose renaming the input to `dims` and maybe go through a deprecation cycle where 
```python
Grid(dims={'yada':{'yada'}})
```
produces identical results to 
```python
Grid(coords={'yada':{'yada'}})
```

An open question remains how to deal with input of both, e.g.: 

```python
Grid(coords={'yada':{'yoda'}}, dims={'yada':{'yada'}})
```

PR #559, opened 2022-11-18
Sgrid

jatkinson1000:
Addition of `sgrid.py` and modifications to `grid.py` to allow reading of SGRID compliant data as per issue #109.
Tested on a COAWST and a Delft3D dataset.
Ideally further testing and writing of unit-tests required following this update.


jatkinson1000:
@TomNicholas This pull request should be ready for review. As a new user the workflows won't run without approval.

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
...
+    Parameters
+    ----------
+    ds : xarray.dataset
+
+    Returns
+    ----------
+    True/False : boolean
+    """
+
+    if "SGRID" in ds.Conventions:
```
TomNicholas, line 17: This fails the docs build because you first need to check if `ds.Conventions` exists before you can safely check if "SGRID" is in there.

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
...
+        return True
+    elif "sgrid" in ds.Conventions:
+        return True
+    elif "Sgrid" in ds.Conventions:
+        return True
+
+    return False
```
TomNicholas, line 24: Minor style nitpick:

```suggestion
    elif "Sgrid" in ds.Conventions:
        return True
    else:
        return False
```

TomNicholas (commented):
Great start @jatkinson1000 ! I left just a couple of minor comments for now, but really the next thing this needs is some tests!

xgcm/sgrid.py
jatkinson1000: Added a new commit that should deal with this.

xgcm/sgrid.py
jatkinson1000: Tidied this up, and as part of the issue above (checking for conventions) now involving nested ifs I think the final return is now required, rather than being inside a final else.

jatkinson1000:
I've just added some example SGRID datasets to the `xgcm/test/datasets.py` file. Given there are already routines to test the gridding process is this sufficient, or do we need further tests for the individual functions? I am going off of what was been done for COMODO at the moment.

jatkinson1000:
I've just added some example SGRID datasets to the `xgcm/test/datasets.py` file. Given there are already routines to test the gridding process is this sufficient, or do we need further tests for the individual functions? I am going off of what was been done for COMODO at the moment.

jatkinson1000:
@TomNicholas Would it be possible to run the workflows for the updates and take a look at this when convenient. Thanks. :)

xgcm/grid.py
```diff
@@ -1325,8 +1328,14 @@ def __init__(
...
         if coords:
             all_axes = coords.keys()
+            print("Generating grid from user-defined information.")
+        elif sgrid.assert_valid_sgrid(ds):
+            all_axes = sgrid.get_all_axes(ds)
+            print("SGRID compliant dataset. Generating grid from SGRID parameters.")
+            coords = {}
         else:
             all_axes = comodo.get_all_axes(ds)
+            print("Generating grid following COMODO conventions.")
```
rabernat, line 1338: Same

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
...
+        b) Syntax match face/vol dimensions to node dimension
+        c) Record Shift and assign to nodes as appropriate
+    3) Facei and edgei are optional and depend on 2D vs 3D.
+       They add no additional information about dimensions so ignored for now.
+
+    SGRID cell padding types can be converted to xgcm node positions as follows:
+      - padding low  refers to 'right'
+      - padding high refers to 'left'
+      - padding both refers to 'inner'
+      - padding none refers to 'outer'
```
rabernat, line 104: This is a very useful Rosetta Stone to understand relationship between SGRID and xgcm. Thanks for writing it up! :+1:

xgcm/test/datasets.py
```diff
@@ -27,6 +27,233 @@
...
+            "node_vert": (
+                ["knode", "jnode", "inode"],
+                np.random.rand(10, 10, 20),
+                {
+                    "standard_name": "pressure",
+                    "units": "bar",
+                },
+            ),
+        },
+    ),
```
rabernat, line 256: These look like very nice and carefully constructed examples. But are they actually being run? Just adding datasets to this list does not make them run. What does your local workflow look like? How are you verifying that the tests run?

These fixtures are imported as follows

https://github.com/xgcm/xgcm/blob/0934c3722d0b8e4842fe8e1bc3daba2ee70f665b/xgcm/test/test_grid.py#L7-L14

AFAICT, unless the datasets are added to one of the fixtures such as `all_datasets` or `all_2d`, they will not actually run as part of the `test_grid.py` tests.

rabernat (commented):
Thanks so much for this excellent work @jatkinson1000.

The code looks good to me. However, I'm not convinced your tests are actually running. As you can see from the [coverage report for your PR](https://app.codecov.io/gh/xgcm/xgcm/pull/559) none of the sgrid code is actually getting called in the test suite.

I think more work is required to wire together the tests datasets you have created into proper fixtures that are run against the `test_grid.py` tests. Happy to help guide you on how to do this if it's not clear.

jatkinson1000:
I was implementing some tests for the SGRID additions and came across something that causes a test to fail.
However, I’m not sure it should do for the inputs I provide, so thought I’d open discussion.

In the test `test_axis_wrap_and_replace_2d` in test_grid.py the function the method `_wrap_and_replace_coords` is called.

As part of this method (line 1135 in grid.py) there is a statement `if new_dim in self._ds.coords:` which checks to find a coordinates with names matching the dimension that we are swapping.

This works because in the test datasets all of the coordinates share the name with a dimension.
However, this is not true in general, and does not apply for some of the SGRID datasets I have seen.
e.g. dimensions may be `icell, jcell` but coordinates are `x_cell, y_cell` where x_cell has dimension icell and y_cell has dimension jcell.
As a result the appropriate coordinates are not picked up for swapping.

1) Is this a legitimate problem? Or is there a restriction in xgcm that assumes coordinates share a name with a dimension?

2) If yes, shall I try and amend this function? However, I am not sure where similar issues may occur elsewhere. Changes here might also want to be under a different PR…

3) A number of the SGRID datasets I have used have 2D arrays as coordinates and no 1D arrays. I think this will cause issues with any xgcm functions like the above that assume 1D coordinates.

I have pushed the test updates (with test_grid.py failing) to my fork here: https://github.com/jatkinson1000/xgcm/tree/sgrid in case you want to talk a look.

jatkinson1000:
After some more searching it appears that the 1D coordinate sharing the name of a dimension is a CF convention requirement (and we discussed last time how xgcm makes something of an assumtion that it is dealing with CF compliant data).

However, a number of our test datasets do not match this requirement, despite being labelled as CF compliant.
I have raised this with their authors in https://github.com/xgcm/xgcm/issues/109#issuecomment-1327718382 .

Perhaps we can discuss removing this restriction if appropriate, though it may need its own issue and PR to do this.

xgcm/test/datasets.py
jatkinson1000: Added the examples (with some changes) into the various sets for running. I know they are now being run because some fail (see discussion below). 🙃

rabernat:
Thanks for raising this issue @jatkinson1000 

> 1D coordinate sharing the name of a dimension is a CF convention requirement

This is not my interpretation of the CF conventions (but I could be wrong!). CF defines "Coordinate Variable" as

> *coordinate variable*
> We use this term precisely as it is defined in the [NUG section on coordinate variables](http://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#coordinate_variables). It is a one-dimensional variable with the same name as its dimension [e.g., time(time) ], and it is defined as a numeric data type with values that are ordered monotonically. Missing values are not allowed in coordinate variables.

Their link is broken, but this works: https://docs.unidata.ucar.edu/nug/current/netcdf_data_set_components.html#coordinate_variables

> It is legal for a variable to have the same name as a dimension. Such variables have no special meaning to the netCDF library. However there is a convention that such variables should be treated in a special way by software using this library.
>
> A variable with the same name as a dimension is called a coordinate variable. It typically defines a physical coordinate corresponding to that dimension. The above CDL example includes the coordinate variables lat, lon, level and time, defined as follows:

The reason I think that coordinate variable are not **always required** is this [statement in CF](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#_overview)

> There are two methods used to identify variables that contain coordinate data. The first is to use the NUG-defined "coordinate variables." The use of coordinate variables is required for all dimensions that correspond to one dimensional space or time coordinates . In cases where coordinate variables are not applicable, the variables containing coordinate data are identified by the coordinates attribute.

In xgcm, we are often dealing with curvilinear grid data in which the logical dimensions (e.g. i, j, k) do not correspond to physical dimensions. Therefore, coordinate variables are not applicable. This corresponds to [CF example 7.3](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#cell-boundaries):

```
dimensions:
  imax = 128;
  jmax = 64;
  nv = 4;
variables:
  float lat(jmax,imax);
    lat:long_name = "latitude";
    lat:units = "degrees_north";
    lat:bounds = "lat_bnds";
  float lon(jmax,imax);
    lon:long_name = "longitude";
    lon:units = "degrees_east";
    lon:bounds = "lon_bnds";
  float lat_bnds(jmax,imax,nv);
  float lon_bnds(jmax,imax,nv);
```

`imax` and `jmax` are dimensions without a coordinate variable.

In the [Xarray data model](https://docs.xarray.dev/en/stable/user-guide/data-structures.html#coordinates), CF / Netcdf "Coordinate Variables" are called "Dimension Coordinates".  Dimension coordinates **used to be required** for all dimensions in Xarray, but this was relaxed several years ago. 

---

So to get to your question


> * Is this a legitimate problem? Or is there a restriction in xgcm that assumes coordinates share a name with a dimension?
> *

YES, I believe this is a legitimate problem. We likely modify `_wrap_and_replace_coords` somehow.

To proceed, it would be helpful to see a very verbose and detailed (yet minimal) example of a dataset that causes our current logic to fail. Would you mind sharing?

jatkinson1000:
Thanks for the updated NUG link. Reading that my interpretation is the same as yours. It had also occurred to me that even if the naming system were resolved there could be issues with 2D grids that don't align with physical axes.

However, reviewing the [Comodo Link](https://web.archive.org/web/20160417032300/http://pycomodo.forge.imag.fr/norm.html#at-location-label) it seems that the Comodo imposes a tighter restriction where coordinates MUST be 1D and match a dimension which is perhaps where this code arose.

A minimal example of an SGRID dataset that causes these tests to fail is as follows:
```
sgrid2Dfail = xr.Dataset(
    # 2D X-Y dataset with 2 variables.
    {
        "grid": (
            (),
            np.array(1, dtype="int32"),
            {
                "cf_role": "grid_topology",
                "topology_dimension": 2,
                "node_dimensions": "XG YG",
                "face_dimensions": "XC: XG (padding: high) YC: YG (padding: high)",
                "node_coordinates": "node_lon node_lat",
            },
        ),
        "data_c": (
            ["time", "YC", "XC"],
            np.random.rand(1, 10, 20),
        ),
        "data_g": (
            ["time", "YG", "XG"],
            np.random.rand(1, 10, 20),
        ),
    },
    attrs={"Conventions": "SGRID-x.x"},
    coords={
        "YGrid": (
            ["YG"],
            np.arange(0, 10),
        ),
        "YCell": (
            ["YC"],
            (np.arange(0, 10) + 0.5),
        ),
        "XGrid": (
            ["XG"],
            np.arange(0, 20),
        ),
        "XCell": (
            ["XC"],
            np.arange(0, 20) + 0.5,
        ),
    },
)
```

This is a 2D x-y grid of cells with dimensions `(XC, YC)`.
Gridpoints have the dimension `(XG, YG)`.
There is an array of data at the cell-centres (`c_data`) and at grid points (`g_data`).
There are 1D coordinates for each direction for both grid and cells (`YGrid, YCell, Xgrid, XCell`).
NB - SGRID padding high is equivalent to left in xgcm notation.

Previously this also had 2D grids with lat and lon coordinates as per a curvilinear grid, but I have removed this to keep things minimal.

This causes failure in the `_wrap_and_replace_coords` of `grid.py` which is used by the various tests in `test_grid.py`.

Let me know if this is useful or if it needs modifying in some way.

My suggestion would be that `_wrap_and_replace_coords` is opened as a separate issue and dealt with in a different pull request. I'm not sure what you want to do with this PR in the meantime as I am happy with the SGRID code and don't think I should amend my datasets to artificially pass the tests they are currently failing. At the same time, however, I understand that PRs that are failing tests shouldn't be merged. What is your preference in this situation?

xgcm/test/test_grid.py
```diff
@@ -691,8 +698,8 @@ def test_grid_ops(all_datasets):
         "fill",
         pytest.param("extrapolate", marks=pytest.mark.xfail(strict=True)),
         "extend",
-        {"X": "fill", "Y": "extend"},
-        {"X": "extend", "Y": "fill"},
+        {"X": "fill", "Y": "extend", "Z": "extend"},
```
jbusecke, line 701: Why is it necessary to add an additional axis here?

jbusecke:
Hi @jatkinson1000 thank you so much for working on this. Reading through this discussion I need to issue a warning that I believe the method `_wrap_and_replace_coords` is part of the [Axis object](https://github.com/xgcm/xgcm/blob/0934c3722d0b8e4842fe8e1bc3daba2ee70f665b/xgcm/grid.py#L64) which is going to be deprecated as part of #557 (@TomNicholas could you confirm that I am interpreting this correctly?). 
As part of that refactor we will also remove a set of tests (including the one you are mentioning here).

That said I do not suggest just ignoring the failures you encounter here, because I do not understand why particularly [these sort of failures](https://github.com/xgcm/xgcm/actions/runs/3546976838/jobs/5956591051#step:6:3517) occur. If I am reading this correctly this means that due to something going wrong, the results of a calculation are actually different. This should never occur and makes me think that maybe there is a bug in the parsing of axes/axis-positions. 

I propose to add a simpler test first. Something along the lines of:
```python

# examples for the input (this should probably include a few different datasets with sgrid/comodo convention (and maybe both?). more on that below
expected_axis_positions = {'X':{'center':'icell', 'right':'inode'}, 'Y':{'center':'jcell', 'right':jnode'}} # this is the same that you would enter manually if you did have no metadata!at 

def test_convention_parsing(dataset, expected_axis_positions_dict):
    grid = Grid(ds) #no inputs here, to see what the automatic parsing does
    # check that all expected axis are present
    assert set(grid.axes)==set(expected_axis_positions_dict.keys())
   # now check that all the values are the same
   for axis in grid.axes:
       assert grid.axes[axis].coords == expected_axis_positions_dict[axis]
```

This will give us a check on the parsed grid_positions (dimensions) earlier than the currently failing tests. In the end hope that we can completely decouple testing the metadata parsing and the computations on data. 

The idea here is that if our internal setup held in the values of grid.axes (and the .coords attribute in each axis) is the same, there should never be a difference in how stuff is calculated down the line!


Re the dimension/dimension-coordinates/coordinates discussion: 
Overall I am more and more realizing that we need to clean up our internal naming and treatment of dimensions and coordinates to make the code base more accessible. To my knowledge xgcm **NEVER** ever deals with coordinates internally! Every core operation should (and I believe does work) with datasets/arrays that have no coordinates at all. 
I believe that in this sense the SGRID conventions are 'overdetermined' (if that nomenclature makes sense here), i.e. they provide information that xgcm does not actually need. 
As a simple mental model, what xgcm needs from SGRID are 'grid_positions' (which are 1 or more dimensions) along an axis.

There are several parts of the code that IMO should be renamed to make this more clear for both users and devs (first and foremost in the input arguments, e.g. #555). But before I digress, I just want to make sure that I think this somewhat imprecise internal treatment might have been remedied by a lot of the code that went into the recent grid_ufuncs refactor (#479 and #344 - this is a longer read but just wanted to paste this for completeness).

TomNicholas:
> [@TomNicholas](https://github.com/TomNicholas) could you confirm that I am interpreting this correctly?

I believe so.

jbusecke:
Dear @jatkinson1000 happy New Year.

@TomNicholas and I are just getting back to xgcm after the winter break. Looking back at this PR, it looks like some of your work here is intertwined with the work we have been doing before Christmas. 
We are planning on having a meeting soon to talk through what is needed to get #557 across the finish line and were thinking we could just all have a chat together to also get your exciting work into main. 

Would you be available for a meeting in the next few weeks?

jatkinson1000:
Hi all,
I became busy on another project just before Christmas, but it would be great for me to get back on this and wrap it up as you say!

I took a pause to see how things developed in parallel with the grid class, so if that's made progress I can try and sync this work. 

I'm on vacation today and next week, but if I take a look again (and at your connected work) once I'm back perhaps something towards the end of the week beginning on the 29th, or the week after?

jbusecke:
Amazing! I would be available on Feb 3 or the week after (Feb 6-10).

jatkinson1000:
Hi @jbusecke and @TomNicholas 

I'm back in and catching up with everything you've been up to since I worked on the SGRID side of things (trying to at least!!)

Could you point me to the latest version you have where you've worked on changes to the grid system so I can try and manipulate the sgrid routines to match that?
I'll also look at converting my parsing routines into something like that discussed in #566 and #568

RE: Meeting 3rd is looking busy for me, but something on the 6th Feb (your morning, my afternoon)?

jbusecke:
Hi @jatkinson1000, excited to hear that you are getting back to this!

We just merged #557, so please feel free to rebase your work here onto main and you should not be blocked anywhere anymore. Thanks for your patience on this pretty massive refactor 😜

> I'll also look at converting my parsing routines into something like that discussed in https://github.com/xgcm/xgcm/issues/566 and https://github.com/xgcm/xgcm/pull/568

Amazing. I think this should serve as a good inspiration. There are some general considerations we still need to hash out with regard to the public API of a generic `metadata_parser` (which would then inspire both your work, as well as a refactor of the COMODO logic as you point out correctly). 

In general though, I think that we want to adhere to the pattern described in #566, where most importantly the parser provides all the input arguments for a call to `xgcm.Grid()`. 

>RE: Meeting 3rd is looking busy for me, but something on the 6th Feb (your morning, my afternoon)?

That works for me! Ill send a poll via email to nail down a specific time.

jatkinson1000:
OK, I've pulled the new code with the changes to grid and removal of axis into its own file.
I think I have re-worked the code I had previously to successfully interpret an SGRID dataset and read it in to an xgcm grid object.
I've added some logic to try for sgrid and then fall back to COMODO as I had in my pre-refactor implementation.
It's now passing required tests locally (pre-commit failed due to a duplicate dataset that I have now fixed) but I am not sure why Build (upstream-dev) is failing.

Next thing is to examine possibly re-working the sgrid code to conform to #566 though I am unsure whether it needs to wait until #568 is complete or if I can work in parallel?

xgcm/grid.py
```diff
@@ -21,7 +21,7 @@
 import xarray as xr
 from dask.array import Array as Dask_Array
 
-from . import comodo, gridops
+from . import comodo, gridops, sgrid
```
jbusecke, line 24: I think we should move both comodo and sgrid into a new module `metadata_parsers.py`?

xgcm/grid.py
```diff
@@ -178,13 +178,22 @@ def __init__(
             # TODO (Julius in #568) full hierarchy of conventions here
             # but override with any user-given options
 
+            # try sgrid parsing
+            if sgrid.assert_valid_sgrid(ds):
```
jbusecke, line 182: I really like the way of checking the validity. Very readable!

xgcm/grid.py
```diff
@@ -178,13 +178,22 @@ def __init__(
...
+                sgrid_ax_names = sgrid.get_all_axes(ds)
+                parsed_coords = {}
+                for ax_name in sgrid_ax_names:
+                    parsed_coords[ax_name] = sgrid.get_axis_positions_and_coords(
+                        ds, ax_name
+                    )
```
jbusecke, line 188: Can we encapsule this into a single method? 

Something like this?
```python
ds, sgrid_kwargs = sgrid.parse_metadata(ds)
```

xgcm/grid.py
```diff
@@ -178,13 +178,22 @@ def __init__(
...
-                parsed_coords[ax_name] = comodo.get_axis_positions_and_coords(
-                    ds, ax_name
-                )
+            else:
+                comodo_ax_names = comodo.get_all_axes(ds)
+                parsed_coords = {}
+                for ax_name in comodo_ax_names:
+                    parsed_coords[ax_name] = comodo.get_axis_positions_and_coords(
+                        ds, ax_name
+                    )
```
jbusecke, line 196: If we could refactor the comodo logic in the same way, that would be great. 

Happy to help with that. We can discuss this at the meeting.

xgcm/grid.py
jbusecke: Oh actullay i misunderstood the way this was setup. I wonder if it might be better to define an Sgrid class in a dedicated module? Maybe @TomNicholas can chime in here.

xgcm/test/datasets.py
```diff
@@ -131,6 +329,12 @@
 
 # include periodicity
 datasets_with_periodicity = {
+    "nonperiodic_sgrid2D": (datasets["sgrid2D"], False),
+    "periodic_sgrid2D": (datasets["sgrid2D"], True),
+    # "xperiodic_sgrid2D": (datasets["sgrid2D"], ["X"]),
+    # "yperiodic_sgrid2D": (datasets["sgrid2D"], ["Y"]),
+    # "sgrid2Dvert": (datasets["sgrid2Dvert"], False),
+    # "sgrid3D": (datasets["sgrid3D"], False),
```
jbusecke, line 337: Lets actually set up some separate test for the parsing instead of adding to these cases here. I think that makes it much easier to separate e.g. tests of grid operations from the setup. 

I propose to have a matching test module either `test_sgrid.py` or `test_metadata_parser.py` (depending on what we decide above). 

In those tests we can focus purely on the output values of the parser without ever needing to construct a grid object. 

Pseudo example:
```python
def test_sgrid_parser():
    ds = construct_test_dataset(...) # this would use your great test cases from here
    ds_parsed, kwargs_parsed = sgrid.parse_metadata(ds) # syntax might vary here, just an illustration for now
    assert some_attribute_on_dataset == 'expected value'
    assert 'X' in kwargs_parsed['coords'].keys()
    ....
```

We know from other tests that if the correct parameters are passed, xgcm will set up a proper grid. Here we only test that these parameters (for now coords, but could later include metrics?) are properly extracted from the given dataset.

xgcm/test/datasets.py
```diff
@@ -144,6 +348,44 @@
...
+            "Z": {"center": "vcell", "inner": "vnode"},
+        }
+    },
+    "sgrid3D": {
+        "axes": {
+            "X": {"center": "XC", "right": "XG"},
+            "Y": {"center": "YC", "right": "YG"},
+            "Z": {"center": "kcell", "left": "knode"},
+        }
+    },
```
jbusecke, line 388: This actually already contains all the expected values I mentioned in my example above. Refactoring these into separate tests should be fairly straightforward.

jbusecke (changes requested):
@jatkinson1000 this is amazing progress! Thank you so much for working on this. 

My main suggestion is to refactor the way the test work to enable more separation between the metadata parsing and anything related to a created grid object. 

This will help with maintenance in the long run (thinking about adding more features to the parser or implementing other metadata standards in the future), but will also not increase our already very high test count for long running computation tests (which can take long, and are overkill to test the parsing itself).

I am very excited for the meeting and getting this merged soon. It will be a great feature for many users.

xgcm/grid.py
jatkinson1000: Thanks, see discussion in https://github.com/xgcm/xgcm/issues/566#issuecomment-1410690075 for related query about doing this for CF-data.

xgcm/grid.py
jatkinson1000: I'd say there is certainly potential for a `gridparsers` module or similar that holds sgrid, comodo, and any other future additions. Alternatively perhaps a `metadata_parsers` module if we want to call cf_parsing from inside sgrid or comodo parsing (see https://github.com/xgcm/xgcm/issues/566#issuecomment-1410690075)

TomNicholas: A separate module is definitely the direction I would like to see things go in. The logical flow inside there might have to incorporate both CF and sgrid logic, but we want all of that to be clearly separate from just creating a `Grid` object from a given set of kwargs.

xgcm/grid.py
jatkinson1000: So a user would have to manually parse the dataset for sgrid `kwargs` (or comodo, or cf etc.) before passing them to `grid.Grid`?
Or do you still want this to be called as part of `grid.Grid` to incorporate some autoguessing as discussed in https://github.com/xgcm/xgcm/issues/566#issuecomment-1351705458

xgcm/grid.py
jatkinson1000: @TomNicholas so if I understand correctly a user would need to call `kwargs = metadata_parsers.parse_sgrid(ds)` (or similar for comodo etc) before then calling `da_grid = grid.Grid(kwargs)`?

Or do you want a generic `kwargs = metadata_parsers.parse_metadata()` method in there that then cycles through the various options (sgrid, cf, comodo etc) to try and auto-infer?

xgcm/grid.py
jbusecke: I lean towards making both possible and defaulting to parsing at grid creation. I think this is a good discussion point for the meeting.

xgcm/grid.py
jbusecke: > Or do you want a generic kwargs = metadata_parsers.parse_metadata() method in there that then cycles through the various options (sgrid, cf, comodo etc) to try and auto-infer?

Very good question, I think we should discuss this in the meeting.

xgcm/test/test_grid.py
jatkinson1000: I think all of the current test cases are 2D, but I added a couple of 3D SGRID tests (one 3D grid, and one 3D grid plus vertical coordinate) It failed here because it's only set for 2D. Just changed similarly in my current work.

xgcm/test/datasets.py
jatkinson1000: Started on this in `xgcm/tests/test_metadata_parsing.py`.
Can discuss more later today but foundations now there.

xgcm/grid.py
jatkinson1000: Made an effort to start on this in tandem with SGRID refactor.
Made changes in `grid.py` moving to `metadata_parsers.py` and left `comodo.py` alone for now.

xgcm/grid.py
jatkinson1000: We can discuss the specifics of the function return values in the meeting, but I have made the requested change to the code so `grid.py` now has `from . import gridops, metadata_parsers` and does not refer directly to either `sgrid.py` or `comodo.py`.

jatkinson1000:
TODO Following discussion on 6 Feb 2023:
- [ ] Autoparse metadata at top of Grid.init()
    - [ ] Call autoparse at top
    - [ ] Compare to user-provided kwargs and raise error if there are conflicts
- [ ] Each autoparse function should return its own ds and set of kwargs which are then combined in autoparse
- [ ] Move sgrid datasets out of `datasets.py` and into `test_metadata_parsing.py`
- [ ] Test unit functions as much as possible
- [ ] Autoparse hierachy??

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
+from collections import OrderedDict
```
dcherian, line 1: Dicts now  preserve insertion order by default so this should be unnecessary.

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
...
+    ----------
+    ds : xarray.dataset
+
+    Returns
+    ----------
+    var_name : str
+        The name of the grid variable
+    """
+
+    for var_name in ds.data_vars:
```
dcherian, line 41:
```suggestion
    for var_name in ds.variables:
```

ALternatively 
```
import  cf_xarray

return ds.cf["grid_topology"]
```

we'll  have to make an upstream change though

jatkinson1000: @jbusecke I think this will add cf_xarray as a dependency of xgcm.
Are you happy for me to do this?
I assume you'll probably want it for any future CF parsing code.

jatkinson1000: @dcherian I've tried the cf_array approach, but it doesn't work as I assume the upstream changes haven't made it to pypi yet?
Do you know when they might appear?
I can leave the code as is (with the cf_xarray alternative commented out) with a TODO note to switch once this feature becomes available?

xgcm/sgrid.py
jatkinson1000: Thanks!
This is the same approach as used in the `grid.py` and `comodo.py`, so I am tempted to perhaps leave it for now in favor of a larger refactor later.

xgcm/sgrid.py
dcherian: Could just ignore it for now. It'll be more  useful once you start parsing CF conventions for metrics like area and volume I think.

xgcm/sgrid.py
```diff
@@ -0,0 +1,230 @@
...
+    elif axis_name == "Z":
+        i_select = 2
+    else:
+        raise ValueError(
+            f"Axis name '{axis_name}' not recognised as one of the default SGRID values 'X', 'Y', 'Z'."
+        )
+
+    # Due to the way vertical coordinates are represented with a 2D dataset
+    # using SGRID catch this case and handle before general cases:
+    if (axis_name == "Z") & ("vertical_dimensions" in ds[sgrid_grid_name].attrs):
```
dcherian, line 149: I wrote a regex based parser in https://github.com/xarray-contrib/cf-xarray/pull/421
Right now it just returns a set of dimension names, but it is capturing the padding info, just not doing anything with it.

Examples
```python
import cf_xarray as cfxr
from cf_xarray.datasets import roms_sgrid

cfxr.sgrid.parse_axes(roms_sgrid)
```
```python
{'X': {'xi_psi', 'xi_rho', 'xi_u', 'xi_v'},
 'Y': {'eta_psi', 'eta_rho', 'eta_u', 'eta_v'},
 'Z': {'s_rho', 's_w'}}
```

We could easily have `ds.cf.sgrid` return a custom object that has all the info you need here (and even more!)

jatkinson1000: Thanks for linking this.
From what I can tell this is tailored specifically to the ROMS use of SGRID, rather than the generic case, is that correct?

dcherian: It should (in theory) work with the others because the regex is general : `re.compile("\\s?(.*?):\\s*(.*?)\\s+(?:\\(padding:(.+?)\\))?")`

Haven't tested it yet though. I'd like to add it to cf-xarray anyway because it gives us a way of identifying dimensions with axes names (X, Y, Z) when there is no metadata associated with the dimension itself.

and since it'll be in cf-xarray anyway, xgcm might as well reuse it.

jatkinson1000:
Hi Guys, I've made updates as discussed with the exception of updating any docs.
Please do take a look.
@TomNicholas I'm still getting mypy errors on `grid.py`, but not from any code I've touched, and I can't easily see how my changes would have affected these areas.

Last questions/things to highlight:
- In the future you may want to to use try/except for combined metadata parsing? Currently uses if sgrid, else comodo.
- Hierarchy of conventions looks superfluous at the moment but written in anticipation of future additions.
- I implemented @dcherian's suggestion of cf_xarray above, but currently commented out as feature not yet available in main release.
- To pass existing tests I had to add a check at the start of `Grid.__init__` for `ds` being of type `xr.dataset`. This is probably a good additional level of safety. You/We may want to be more specific about the error message and remove the identical check from the `Axis` class now that it's private.
- Comparing autoparsed data wasn't super straightforward. After a brief chat with @jbusecke we decided to ignore `periodic` (as it is soon to be deprecated) and compare each other kwarg to the default of `None`. This assumes that a user would never call `Grid()` with any of the constructors set as `None` - I can't see that they would, but can @TomNicholas confirm?
- A number of tests in `test_grid.py` had to be updated to include `autoparse_metadata=False` to preserve original behavior. I added a couple of new ones to check for failure when there are conflicts in user-supplied kwargs and.
The only notable change was to 3 tests which previously autoparsed comodo data, but had passed the kwarg `boundary` passed to the constructor. I changed it to autoparse comodo before calling `Grid` and use the `coords` kwarg with `autoparse_metadata=False` which we agreed should be the approach for this case.
- Tests for individual sgrid functions and metadata parsing of sgrid added in `test/test_metadata_parsers.py`

xgcm/test/test_grid.py
jatkinson1000: Removed with removing sgrid datasets from grid tests.

xgcm/sgrid.py
jatkinson1000: Ah, yes, gotcha.
The regex is nice.
Can add a todo to update method along with locating grid_topology identification once upstream changed are available.

jatkinson1000:
See [comment above](https://github.com/xgcm/xgcm/pull/559#issuecomment-1422976227) for discussion to the latest changes.

- [X] Closes #109 
- [X] Tests added
- [ ] Passes pre-commit run --all-files
- [X] User visible changes (including notable bug fixes) are documented in whats-new.rst
- [ ] New functions/methods are listed in api.rst

API list seems rather sparse, so not sure what you want adding here...
Tests not passing on CI but are locally. Due to changes - see [comment above](https://github.com/xgcm/xgcm/pull/559#issuecomment-1422976227)

xgcm/grid.py
```diff
@@ -129,8 +129,60 @@ def __init__(
         ----------
         .. [1] Comodo Conventions https://web.archive.org/web/20160417032300/http://pycomodo.forge.imag.fr/norm.html
         """
+        if not isinstance(ds, xr.Dataset):
+            raise TypeError(
+                f"ds argument must be of type xarray.Dataset, but is of type {type(ds)}"
```
jbusecke, line 134:
```suggestion
                f"ds argument to `xgcm.Grid` must be of type xarray.Dataset, but is of type {type(ds)}"
```

jbusecke: This makes the error unique and we can test this one separately from the axis version?

xgcm/sgrid.py
jbusecke: Fine with this here. Could you make a short issue to track this? That would be very helpful.

jbusecke:
Sorry for this annoying approval process on the CI. I am not quite sure how I can greenlight this for good...

dcherian:
Should be good now:

jatkinson1000:
Hi @jbusecke I've updated the docstrings in grid.py to pass the read the docs build.
Strangely the errors weren't caused by things I've edited (as far as I am aware) so not sure how that happened.
Also fixed a few similar warnings elsewhere that were being raised in docstrings and .rst docs.

As for tests, they're failing due to running the old tests (without `autoparse=False` kwarg in `Grid()`) on the new code.
I'm not sure what your approach is for these situations, whether you want to pull and test locally yourself, set up a workflow for this branch, and/or manually review. Let me know if I can help.

jatkinson1000:
Hi @jbusecke I think I've sorted it.
There were some changes to main that I synced up here.
Some of the ufunc and padding tests were failing because they also needed the autoparse_metadata=False in places where the user was specifying a set of coords to the Grid constructor.
I've added these in now where required.

The need to explicitly add this as an argument is because we agreed that the 'correct' workflow for a user who wanted to override any autoparsed metadata would be to autoparse the kwargs first and then change as desired before passing to the `Grid` constructor.
We decided this because it means that any overriding done by the user is done as a very clear and deliberate action within their code.
This avoids potential cases where a user autoparses and adds their own argument and it is ambiguous (to the naïve user) as to which one is being used within the `Grid` constructor.

I have added a description of this workflow to the docs that should appear when you do an in-depth review.
I think this is now ready!

xgcm/sgrid.py
jbusecke: Might be nice to capture this exchange and actionable items in an issue.

jbusecke (approved):
THANK YOU SO MUCH for your work (and the continued patience) @jatkinson1000. This looks great and Ill merge this now.

PR #562, opened 2022-11-29
Fix confusion between forwards and backwards diff in grid ufunc docs

### LINKED ISSUE #504: [DOC] Issue on page /grid_ufuncs.html
In cell [2], where diff_forward is defined. It seems like it has been defined in the opposite sense of the usual.
The usual definition would be diff_a = a[i+1] - a[i], whereas (if I understand roll correctly) it seems that here the function is calculating diff_a = a[i] - a[i+1]. (so there is a sign error).

Also, regardless of the sign error the diff_a would be a variable that ends up on ax1:right, but instead in the paragraph below it seems to suggest that things end up on ax1:left (which is what would happen if we were doing a backward difference.

TomNicholas:
- [x] Closes #504
 - [ ] ~~Tests added~~
 - [ ] ~~New functions/methods are listed in `api.rst`~~

### DIFF
--- a/doc/grid_ufuncs.rst
+++ b/doc/grid_ufuncs.rst
@@ -106,7 +106,7 @@ To give you an idea of how we might use grid ufuncs here is a table of possible 
 Defining New Grid Ufuncs
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


TomNicholas:
@dhruvbalwada does this fix the problem you reported in #504?

jbusecke:
Pinging @dhruvbalwada. Do you think we can merge this?

PR #563, opened 2022-12-06
Reduce the frequency of pre-commit autoupdates

jbusecke:


dcherian:
Yes less spam! quarterly is my preferred frequency btw but even this is good

PR #577, opened 2023-01-26
Switch CI to micromamba

### LINKED ISSUE #576: [BUG] Our CI is broken!
All of our recent builds fail with this less than helpful error during the creation of the conda environment.

```
Error: The process '/usr/share/miniconda/condabin/conda' failed with exit code 1
```

see e.g. here: https://github.com/xgcm/xgcm/actions/runs/3986809142/jobs/6835819892#step:4:354

I noticed that e.g. [xarray](https://github.com/pydata/xarray/blob/d385e2063a6b5919e1fe9dd3e27a24bc7117137e/.github/workflows/ci.yaml#L94-L103) and [intake-esm](https://github.com/intake/intake-esm/pull/500/files) have switched to micromamba. 
Ill try to do that now and see if it fixes the problem (Seems like this has also improved performance!). But if anyone has some more details as to what might have happened here, I would be curious. 

cc @andersy005 @keewis

jbusecke:
Towards #576

 - [x] Closes #576 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10"]
     steps:
       - uses: actions/checkout@v3
       - name: Create conda environment
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,10 @@ on:
       - "*"
   schedule:
     - cron: "0 13 * * 1"
+    
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


jbusecke:
Oh wow! Creating the conde env went from ~4 min:
image
to less than a minute!
image

jbusecke:
Still [too early](https://github.com/xgcm/xgcm/actions/runs/4020066730/jobs/6907593897) to get py3.11 in the mix, but that is ok for me right now.

jbusecke:
The upstream dev seems to be failing due to something [related to numba](https://github.com/xgcm/xgcm/actions/runs/4020132301/jobs/6907721971) which I do not fully understand, but the fact that this is pulling a dev version of numba
```
numba                     0.57.0dev0      np1.21py3.10ha54aa7d_g2ca22c20f_1048 
```
makes me think we should ignore this for now and keep an eye on this!

keewis:
the error is caused by `numba` currently not supporting `numpy>=1.24`.

And you can speed up the CI even further by [caching the environment](https://github.com/pydata/xarray/blob/d385e2063a6b5919e1fe9dd3e27a24bc7117137e/.github/workflows/ci.yaml#L99), not just the downloads. The result would be that instead of ~1min you'd get ~15-20s on cache hits (on linux, at least).

PR #578, opened 2023-01-27
Rename `boundary` to `padding`. 

jbusecke:
Got really inspired by #557 and decided to try to build on top of that with a PR that renames `boundary` to `padding` (which makes it consistent with the nomenclature we discussed, and is already used in the `padding.py` module). This should lead to less confusion over the fact that we do not in fact implement physical boundary conditions yet.

This worked pretty well, except for a bunch of errors which all seem to have the same source. 

```
FAILED xgcm/test/test_grid.py::test_dask_vs_eager[yperiodic_2d_left-extend-diff] - TypeError: diff_center_to_right() got an unexpected keyword argument 'padding'
```

There must be some place in the grid_ufunc internals where I have missed to rewire the logic to take `padding` instead of `boundary`. @TomNicholas, do you have a hunch what could be going on?

- [ ] Fix remaining tests failures
- [ ] Rebase/Merge onto main once #557 is merged.
- [ ] Make sure to rename all occurences/sections dealing with boundary/padding (maybe also finish #476)

------
 - [ ] Towards #414


hdrake:
Closing as superseded by #696, which carries the `boundary` → `padding` rename across the current codebase and is queued near the end of the v1.0.0 merge order (see #733) so the rename only has to be reconciled once. Thanks for starting this!

PR #583, opened 2023-02-14
Propagated unpadded coords in pad

### LINKED ISSUE #575: [BUG] Time coordinate being dropped by apply_as_grid_ufunc

TomNicholas:
- [ ] Closes #575
 - [x] Tests added
 - [ ] ~~New functions/methods are listed in `api.rst`~~

### DIFF
--- a/xgcm/test/test_padding.py
+++ b/xgcm/test/test_padding.py
@@ -135,6 +135,24 @@ class TestPadding:
         xr.testing.assert_allclose(expected, result)
 
 
+def test_strip_only_padded_coords():
+    """Tests coordinates along dimensions not involved in padding are still propagated. See GH issue #575."""
+    ds, coords, _ = datasets_grid_metric("C")
+    grid = Grid(ds, coords=coords)
+    original = ds.tracer
+
+    result = pad(
+        original,
+        grid,
+        boundary={"X": "periodic"},
+        boundary_width={"X": (1, 1)},
+    )
+
+    xr.testing.assert_equal(original.coords["time"], result.coords["time"])
+    xr.testing.assert_equal(original.coords["yt"], result.coords["yt"])
+    xr.testing.assert_equal(original.coords["zt"], result.coords["zt"])
+
+
 # TODO: Make sure that we cannot specify mixed methods for padding if the input is something like `cube-sphere` or `tripolar`
 
 


TomNicholas:
@jbusecke to fix #575 we need to change `_strip_all_coords` to instead strip only the coords which are a function of the padded dimensions. Therefore we need a function like
```python
def _strip_axis_coords(da, padding_axes):
   ...
```

For `_pad_basic` this is fine, but for face-connected grids its the implementation is a bit more complicated because `pad_axes` is not known until after you know you have a face-connected grid.

Instead how about we move the coordinate stripping inside of `_pad_basic`?

jbusecke:
> For _pad_basic this is fine, but for face-connected grids its the implementation is a bit more complicated because pad_axes is not known until after you know you have a face-connected grid.

I am not sure I follow this entirely. But could we do something like this:
```python
def _strip_all_axis_coords(da, grid):
    # get rid of any dim that is associated with a grid axis
   ...
```

If I understand this correctly, the problem was not with dims associated with an axis that is not being padded, but instead with dims that are not associated with any axis? This approach would strip more than is needed, but would still probably work?

hdrake:
Superseded by #721 (merged), which resolves #575.

To confirm, I checked out this PR's test (`test_strip_only_padded_coords`) against current `master` (which includes #721). Two things:

1. The test no longer runs as written — `Grid(ds, coords=coords)` now raises `ValueError: Autoparsed Grid kwargs 'coords' conflict with user-supplied kwargs` and needs `autoparse_metadata=False` (API drift since 2023).
2. Once that's fixed, the test still fails: `pad()` returns an object with **no** coordinates, so the `time`/`yt`/`zt` assertions don't hold.

That failure is expected, because #575 was fixed at a **different layer** than this PR targeted:

- **This PR:** make the low-level `pad()` primitive propagate non-core coords.
- **#721 (merged):** keep `pad()` as a deliberate coord-stripper (see `xgcm/padding.py`: *"we will strip all coordinates ... This ensures that any output from this function is stripped"*) and instead restore input coords one level up, in `apply_as_grid_ufunc` via `_reattach_coords`.

So the user-facing bug is resolved — #721's `test_non_core_coord_*` tests pass on `master` and issue #575 is closed — but this PR's approach of preserving coords inside `pad()` itself was intentionally not adopted. Closing as superseded. Thanks @TomNicholas for the original test and diagnosis!

PR #602, opened 2023-04-27
Remove mentions of `extrapolate` in docstrings

jbusecke:
Thanks to @dhruvbalwada for noticing this. When removing `extrapolate` as option in #470 there were a bunch of docstrings left which still mention `extrapolate` as viable option (and did not yet mention the `periodic` option either).

BONUS: This gets rid of some xfail tests. 


jbusecke:
@TomNicholas does this all look ok to you?

Issue #581, opened 2023-02-03
[BUG] Issue in grid_ufunc when original_arg is a dictionary


andrewdelman:
Encountered this apparent bug in v0.8.1 when calling diff_2d_vector, it originated in grid_ufunc.py (about line 1022):

```python
        original_arg_chunks = original_arg.variable.chunksizes
```
gives error:
```
'dict' object has no attribute 'variable'
```

`original_arg` is expected to be an xarray DataArray, but instead is a dictionary with a *value* that is an xarray DataArray.  (This does not seem to happen with regular xgcm diff.)  This is my inelegant solution that seems to work, replacing the `original_arg_chunks` assignment with:

```python
        if isinstance(original_arg,dict):
            original_arg_chunks = tuple(original_arg.values())[0].variable.chunksizes
        else:
            original_arg_chunks = original_arg.variable.chunksizes
```

andrewdelman:
Recently encountered more errors in `grid_ufunc.py` with `original_arg` of type dictionary, this time in the `_map_func_over_core_dims` function.  Tried fixes similar to those above (replacing `arg` with `tuple(arg.values()[0])`, i.e. the first value of the `arg` dictionary), which seemed to resolve the problem there but then triggered some kind of dimension chunking mismatch in `dask.array`.  For example, in `dask/array/blockwise.py`:

```python
    270 elif isinstance(adjust_chunks[ind], (tuple, list)):
    271     if len(adjust_chunks[ind]) != len(chunks[i]):
--> 272         raise ValueError(
    273             f"Dimension {i} has {len(chunks[i])} blocks, adjust_chunks "
    274             f"specified with {len(adjust_chunks[ind])} blocks"
    275         )
    276     chunks[i] = tuple(adjust_chunks[ind])
    277 else:

ValueError: Dimension 0 has 13 blocks, adjust_chunks specified with 12 blocks
```

So the fix I tried before is no longer sufficient, at least for me.  Not sure if this issue is recreated by other users when calling `diff_2d_vector`.

Issue #466, opened 2022-04-07
[FEATURE] Major refactor of the test suite


jbusecke:
Our test suite has balooned out of proportion to the point where CI actions actually time out some times. 

After the 0.7.0 release, we should spend some time to prune/refactor tests. 

Just using this issue for now to collect relevant bits: 

- https://github.com/xgcm/xgcm/pull/459/files#r841908726

TomNicholas:
A better pattern to use for the tests would be a factory pattern. Then we parametrize the tests with arguments to the factory. e.g.

```python

@pytest.mark.parametrize(position='left', dimensions=2)
def test_xgcm_methods(dataset_factory):
    ds = dataset_factory(position=position, dimensions=dimensions)

    assert ds.diff == ...
```

The advantage of this is that you don't need to look back at a fixture defined in a different file to see everything that your test was parametrized with.

(Idea from Daniel Rothenberg)

jbusecke:
Related to the factory idea proposed: I think this would be great and we could have an option to generate different metadata conventions in the same dataset, or not produce metadata at all. The latter case would help make some tests more clear. 

E.g. in [this recent change](https://github.com/xgcm/xgcm/blob/93d4017ba8c34dbb64f2521851bed9c629339866/xgcm/test/test_grid.py#L35) we could make the intention of the tests clearer by 'producing' a dataset without metadata instead of setting `parse_metadata=False`.

PR #607, opened 2023-05-11
Add Tests for vector input

jbusecke:
- [ ] Towards #581
 - [ ] Tests added

---

Tom: Suggested code improvements to prevent lots of bugs like this occurring:
- Properly type hint all places where a vector could be passed,
- Use the `typing` module's `@overload` decorator to type hint functions in an unambiguous way (i.e. scalar->scalar & vector->vector, but not vector->scalar),
- Type hinting `pad` and its internals is the most important part, because that's the only place that vectors are treated meaningfully differently,
- Other cases could potentially be handled by a `@handle_vectors` decorator wrapping methods, that unpacks, calls method, then re-wraps.
- Test padding dask arrays in functions like `diff_vector_2d`,
- Add an `is_vector` flag to any test dataset factory created (see #https://github.com/xgcm/xgcm/issues/466#issuecomment-1382451003)

### DIFF
--- a/xgcm/test/test_grid_ufunc.py
+++ b/xgcm/test/test_grid_ufunc.py
@@ -609,7 +609,8 @@ class TestGridUFuncNoPadding:
 
 
 class TestGridUfuncWithPadding:
-    def test_1d_padded_but_no_change_in_grid_position(self):
+    @pytest.mark.parametrize("vector", [True, False])
+    def test_1d_padded_but_no_change_in_grid_position(self, vector):
         def diff_center_to_center_second_order(a):
             return 0.5 * (a[..., 2:] - a[..., :-2])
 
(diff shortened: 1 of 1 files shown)


jbusecke:
Added a test to check functionality with vector input. 

This exposed another (but perhaps related bug) to #581:
From what I can see the vector dict is passed to `_pad_basic` as is, which should never happen. 
Digging into the pad function now.

jbusecke:
Actually I think this bug is distinct (it happens inside the pad function, while the original bug occured in `_rechunk_to_merge_in_boundary_chunks` which happens after padding) from #581, but both are related to the vector syntax. 
Trying to add a test that reproduces #581 first, before digging into `pad`. 

Also I am not sure what the best testing strategy is here. 

I think we need to parameterize the input (vector, non-vector) at least in a couple of places:
- For all padding option on the stand-alone tests for padding
- In the grid ufuncs tests, but do we need this for all of the tests?

Issue #625, opened 2024-03-27
[BUG] cumsum seems to apply periodic B.C. even when it knows that is not the case. 


dhruvbalwada:
We should expect that for non-periodic axis cumsum should start integration with first value as 0 (if using outer or left for the axis being cumsumed to), and increase or decrease accordingly.

Instead, the current implementation seems to be applying periodic conditions and filling the first value with the sum. 

```
ds = xr.Dataset(
    coords={
        "zl": (["zl"], np.arange(1,15)),
        "zi": (["zi"], np.arange(0.5, 15.5))
    }
)

grid = Grid(ds, 
            coords={"Z": {"center": 'zl', 'outer': 'zi'}},
           periodic=False)

temp = grid.cumsum(ds.zl, 'Z')
temp1 = grid.cumsum(ds.zl, 'Z', boundary='fill', fill_value=0.)
```

Both cumsums start from 0 and increase, instead the output is: 
![Screen Shot 2024-03-27 at 2 48 31 PM](https://github.com/xgcm/xgcm/assets/18236610/a16fd1d3-dcf9-4784-a38d-f78fb32336aa)

As seen above, when explicit B.C. is not provided we end up with very large value at start.

jbusecke:
Sorry for the long wait here @dhruvbalwada. 

> We should expect that for non-periodic axis cumsum should start integration with first value as 0 

Since we want to keep this as general as possible I think that we cannot expect this to be generally true? We can set some defaults but in the end the user has to specify e.g. the fill value appropriately (and in this case explicitly). 

The boundary however is a different story. There we chose to set 'fill' as a default, which is not happening in your example. I can reproduce this on v0.8.1 (on the leap hub), but not on the current main branch. Maybe we should just release a new version?

I think the robust solution here is to not use `periodic` at all (we should just deprecated it and be done with it). Instead set the boundary explicitly (this is more readable and logical than 'not periodic' which is ambigous)
```
grid = Grid(ds, coords={"Z": {"center": 'zl', 'outer': 'zi'}}, boundary='fill')
```

Issue #624, opened 2024-03-27
[BUG] Non-intuitive behavior if periodic not set as list


dhruvbalwada:
If the periodic argument to Grid is not passed as a list then the resultant grid object has non-intuitive result. 

In example below user may expect both object to have only X dim as periodic. However, in case where list is given X dim is periodic and Z is non-periodic, while when the dimension is specified as string the option True is likely being used. 

```
import xarray as xr
import numpy as np
from xgcm import Grid

ds = xr.Dataset(
    coords={
        "zl": (["zl"], np.arange(1,15)),
        "zi": (["zi"], np.arange(0.5, 15.5)),
        "xl": (["xl"], np.arange(1,15)),
        "xi": (["xi"], np.arange(0.5, 15.5)),
    }
)

grid_list = Grid(ds, 
            coords={"Z": {"center": 'zl', 'outer': 'zi'},
                    "X": {"center": 'xl', 'outer': 'xi'}},
           periodic=['X'])

grid_non_list = Grid(ds, 
            coords={"Z": {"center": 'zl', 'outer': 'zi'},
                    "X": {"center": 'xl', 'outer': 'xi'}},
           periodic='X')
```

The solution would be to raise an error to the user and force them to use a list, or automatically convert string to a list of a single value.

Issue #604, opened 2023-05-03
[BUG] Default boundary condition is periodic even when specified otherwise


dhruvbalwada:
The issue: Even when xgcm has been told that a coordinate is not periodic, it still chooses that as the default behavior - instead of throwing an error or warning. 

```
import xarray as xr 
import xgcm
import numpy as np

dx = 0.5
L = 20
x_left = np.arange(0,L-dx, dx)
x = np.arange(dx, L, dx)

slope = 2.
y = slope*x

da_y = xr.DataArray(y, dims=['xc'], coords={'xc':x})
ds_y = xr.Dataset(coords={'xc':x, 'xg':x_left})
ds_y['y'] = da_y

grid = xgcm.Grid(ds_y, 
                 coords={"X": {"center": "xc", "left": "xg"}},
                periodic=None)

dy_default = grid.diff(ds_y.y, 'X')
dy = grid.diff(ds_y.y, 'X', boundary='extend')
```

In this example we told grid that `X` is not a periodic direction. When we print grid, it says that `X` is not periodic. However, the dy_default (where we did not pass a boundary condition) still assumes `X` is periodic without alerting the user.

dhruvbalwada:
Might be related to #509

hdrake:
Confirming I'm running into this same issue.

jbusecke:
Is this fixed by using the `boundary=` input syntax? In that case deprecation of `periodic` might be the solution here.

hdrake:
> Is this fixed by using the `boundary=` input syntax? In that case deprecation of `periodic` might be the solution here.

Yep–see https://github.com/xgcm/xgcm/issues/606#issuecomment-1702742283.

But that's not in the most recent release `v0.8.1` yet, right?

hdrake:
Revisiting this on current `master`.

**Status of the reported reproducer.** The specific cases that originally silently wrapped no longer do. Building a `Grid` with `periodic=None`, `periodic=False`, or an explicit `boundary=...` now yields a non-periodic (`fill`/`extend`) axis, so `grid.diff(..., 'X')` without a per-call `boundary` no longer wraps:

```python
for kw in [dict(periodic=False), dict(periodic=None), dict(boundary='fill')]:
    g = xgcm.Grid(ds_y, coords={"X": {"center": "xc", "left": "xg"}},
                  autoparse_metadata=False, **kw)
    print(kw, "->", g.axes['X'].boundary)   # all -> 'fill'
```

[...]

PR #534, opened 2022-08-26
Replace `periodic` kwarg with an option in `padding`

raehik:
(Note that `padding` is the intended updated name for `boundary` -- that change isn't (for now?) covered in this PR.)

Related issues:

  * https://github.com/xgcm/xgcm/issues/195
  * https://github.com/xgcm/xgcm/issues/509
  * overview issue https://github.com/xgcm/xgcm/issues/414

PR #626, opened 2024-04-11
Deprecate `periodic` input ...

### LINKED ISSUE #195: boundary="periodic"?

jbusecke:
This is my attempt to resolve a bunch of issues that stem from the fact that we still support `periodic` as input. This should stop right here. 

This is a reimplementation of #534 which is quite far behind at this point (my apologies to the original author for falling behind on the maintainer duties here). 

Totally just taking a first stab here to give people a point of reference. This needs some careful (not past 9pm) consideration of how to change the tests. The fact that periodic is so deeply embedded into the test suite is really the reason that it was not removed earlier. 

 - [ ] Closes #195,#509, #604, #624, #625
 - [ ] Tests added

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -63,7 +63,6 @@ class Grid:
         self,
         ds: xr.Dataset,
         coords: Optional[Mapping[str, Mapping[str, str]]] = None,
-        periodic: bool = True,
         fill_value: Optional[Union[float, Mapping[str, float]]] = None,
         default_shifts: Optional[
             Mapping[str, str]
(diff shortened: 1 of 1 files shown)


hdrake:
Closing as superseded by #746, which carries this out on the current codebase (the `periodic` kwarg is removed, boundary handling is explicit and non-periodic by default, and the whole test suite + executed docs are migrated off `periodic=`). Your first stab here seeded the approach — credited via `Co-authored-by` in the successor. Closes out roadmap item §A of #733; the associated bugs (#195/#509/#604/#624) close with #746.

Issue #605, opened 2023-05-03
[BUG] Cumsum flips axis


dhruvbalwada:
Issue (identified by @NiekK19): xgcm's cumsum seems to operate with an additional sort. 

In the example below we compare different cumsum functions - xarray's default and xgcm's version. 

```
import xarray as xr 
import xgcm
import numpy as np
import matplotlib.pyplot as plt

dx = 0.5
L = 20
x_left = np.arange(0,L-dx, dx)
x = np.arange(dx, L, dx)

slope = 2.
y = slope*x

da_y = xr.DataArray(y, dims=['xc'], coords={'xc':x})
ds_y = xr.Dataset(coords={'xc':x, 'xg':x_left})
ds_y['y'] = da_y

grid = xgcm.Grid(ds_y, 
                 coords={"X": {"center": "xc", "left": "xg"}},
                periodic=None)

ds_y.y.cumsum().plot(label='Xarray cumsum')
grid.cumsum(ds_y.y, 'X', boundary='fill', fill_value=0.).plot(linestyle='--',label='xgcm cumsum')
ds_y.y.sortby(ds_y.xc, ascending=False).cumsum().plot(label='xarray cumsum after sort')
grid.cumsum(ds_y.y.sortby(ds_y.xc, ascending=False),  'X', boundary='fill', fill_value=0.).plot(label='xgcm cumsum after sort')
plt.legend()
plt.grid()
```
![Screen Shot 2023-05-03 at 10 17 42 AM](https://user-images.githubusercontent.com/18236610/235943380-055de2a7-c8b5-44d0-9d1b-22aeaa031eda.png)

As seen in the example: when x is increasing, both cumsum function give same answer (slight shift is to be expected as xarray is not grid aware). However, when we flipped the x axis the two functions give different answers. In the flipped case xarray works as expected: doing a cumulative sum starting at the other side of the x-axis (since the end of the axis is now the beginning). However, xgcm seems to do the cum in the same way, but then give a result where only the coordinate has been flipped.  I think xarray is doing is what a user might expect, but xgcm is doing something extra without prompting the user.

jbusecke:
Thanks for raising this @dhruvbalwada. 

Some findings from digging more into this:

```python
ds_y_flipped= ds_y.y.sortby(ds_y.xc, ascending=False)
ds_y_flipped
```
sorts both the values and coordinates

[...]

PR #665, opened 2025-03-18
Implement axis directionality (in low-level ufuncs)

### LINKED ISSUE #344: Proposal: grid_ufunc

hdrake:
This is an attempt to revive prior efforts to add a sense of directionality to xgcm grid axes, especially https://github.com/xgcm/xgcm/pull/338.

Unfortunately, @gmacgilchrist's prior attempt at this PR predated the ufunc-ification of grid operations (see PR https://github.com/xgcm/xgcm/pull/479 and the issue it closes https://github.com/xgcm/xgcm/issues/344), which meant that it eventually became stale and was closed.

 - [ ] Closes https://github.com/xgcm/xgcm/issues/337
 - [ ] Tests added

### DIFF
--- a/xgcm/gridops.py
+++ b/xgcm/gridops.py
@@ -15,7 +15,7 @@ through when `xgcm.Grid.diff()` is called. See xgcm.grid_ufunc._select_grid_ufun
 
 
 # TODO can we allow for grouping these definitions into classes? Similar to pytest tests?
-
+# TODO what is the best way to generalize the Diff and Cumsum ufuncs so that they have a sense of directionality?
 
 # Diff
 


hdrake:
@TomNicholas, I started reading through your big ufunc PR and the source code. I would love to take you up on your offer of walking me through the current state of the code, and then maybe you can talk me through how you would approach this PR as a concrete example. Pinging @NoraLoose, @raphaeldussin, and my new postdoc @liu-ran to keep them in the loop.

Easiest to coordinate a call over email?

hdrake:
I think we can do this in a way that addresses https://github.com/xgcm/xgcm/issues/605 and picks back up where @jbusecke 
 left off on https://github.com/xgcm/xgcm/pull/256 as well.

TomNicholas:
Sounds good. I'm going to have to remind myself of how all this works 😅

My email is Tom at Earthmover dot io

hdrake:
Closing in favor of #728, which implements axis directionality on the current grid-ufunc machinery (this draft only carried a TODO comment). #728 resolves #337; the complementary reverse-cumsum/cumint work is in #729.

PR #700, opened 2026-04-10
MAINT: Add Zizmor security linting

VeckoTheGecko:
- [x] Closes None

Similar to https://github.com/pydata/xarray/pull/11269

I think this would be better for security.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,10 +18,13 @@ concurrency:
 env:
   PIXI_VERSION: "v0.63.0"
 
+permissions: {}
 
 jobs:
   cache-pixi-lock:
     runs-on: ubuntu-slim
+    permissions:
+      contents: read
     outputs:
       cache-key: ${{ steps.pixi-lock.outputs.cache-key }}
       pixi-version: ${{ steps.pixi-lock.outputs.pixi-version }}
@@ -42,6 +45,8 @@ jobs:
     needs: ["cache-pixi-lock"]
     if: github.repository == 'xgcm/xgcm'
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     timeout-minutes: 60
     defaults:
       run:
@@ -89,6 +94,7 @@ jobs:
     name: GHA Security Analysis using Zizmor
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       security-events: write # Required for upload-sarif (used by zizmor-action) to upload SARIF files.
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/pythonpublish.yaml
```diff
@@ -31,12 +37,11 @@ jobs:
...
-        env:
-          TWINE_USERNAME: "__token__"
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
         run: |
           python -m build
           twine check dist/*
-          twine upload dist/*
+      - name: Publish to PyPI
+        if: github.event_name == 'release'
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
```
VeckoTheGecko, line 41: Note that this would mean that we would no longer use the PyPI token (instead we should set up trusted publishing in pypi - which results in better security)

.github/workflows/ci.yaml
```diff
@@ -18,20 +18,25 @@ concurrency:
...
   cache-pixi-lock:
     runs-on: ubuntu-slim
+    permissions:
+      contents: read
     outputs:
       cache-key: ${{ steps.pixi-lock.outputs.cache-key }}
       pixi-version: ${{ steps.pixi-lock.outputs.pixi-version }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
VeckoTheGecko, line 32: All these hashes were inserted by Zizmor (so are sure to match the tag in the comments).

We don't have to worry about these getting upgraded - dependabot is able to parse this, and will bump these (alongside the commented version) when it does its updates

.github/workflows/pythonpublish.yaml
```diff
@@ -10,33 +10,32 @@ on:
...
+      - name: Publish to PyPI
         if: github.event_name == 'release'
-        env:
-          TWINE_USERNAME: "__token__"
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-        run: |
-          python -m build
-          twine check dist/*
-          twine upload dist/*
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
```
VeckoTheGecko, line 41: Did a bit of a refactor here to simplify. In all cases, it builds. Only on releases does it upload to PyPI.

Before there was some unnecessary duplication

Issue #194, opened 2020-04-05
does xgcm handle models with a tri-polar?


navidcy:

jbusecke:
Hi @navidcy, thanks for using xgcm. Short answer: Yes. I have worked with several models with xgcm sucessfully. 

Do you have a concrete problem or example that you are concerned about?

dcherian:
@jbusecke this came up in our Ocean section meeting call today. It'd be nice to see an example notebook analyzing output on a tripolar grid (I have never worked with these so I'm pretty confused on how to represent things in xgcm).

jbusecke:
What does pop use? 

Happy to make a n example with MOM5/6, but I just have to say that I am quite backed up with stuff rn.

In principal it doesn’t matter what kind of grid you have if it fits into one “logical” tile. The only time when things get more complicated is smth like a cubed sphere grid I think. But Ryan has that implemented in xgcm too.

raphaeldussin:
is the question related to the folding of the north boundary? Which would basically be a face connection of a face onto itself with direction change from one side of the pivot point to the other.
I don't think I've seen this option in the docs/code.

dcherian:
@raphaeldussin I came to that conclusion too. And yes I hadn't seen that option and so wanted to see what @jbusecke was doing

jbusecke:
Hmm maybe I have been thinking about this a bit to simplified (forgive me I stay as far away from the Arctic as possible hahaha).
So you are asking what happens when you want to do e.g. a y derivative across the 'seam'? That sounds like something that could be addressed with some of the cubed-sphere logic that @rabernat has built. I however know next to nothing about that part of the code, so perhaps this question is best deferred to Ryan?

raphaeldussin:
that's exactly right, velocities reverse directions between the 2 parts of the fold. You need the information on the pivot point (index and point type). The most natural choice is to use a corner (f, q, psi,... depending on your convention) but there's configurations for which the pivot is on a V-point (ORCA05 if my memory is right). I think the face_connection logic can work if we build a virtual new face containing the "mirror" image of the northernmost row.

what do you think?

rabernat:
> I think the face_connection logic can work if we build a virtual new face containing the "mirror" image of the northernmost row.

These kinds of things are already done for the LLC grids, so should definitely be possible!

dcherian:
[...]

PR #707, opened 2026-05-31
Add bipolar-fold support to face_connections for tripolar grids

hdrake:
## Summary

Adds support for the **bipolar Arctic fold** of tripolar ocean grids to xgcm's `face_connections` machinery.

A tripolar grid's northernmost row folds onto itself, identifying column `i` with its mirror column `N − i`. The existing `reverse` flag only flips the **orthogonal** (connection-axis) dimension, so it cannot represent this fold, which requires a **tangential** mirror on a *same-axis* (Y↔Y) connection.

The added functionality is much more general than the tripolar grid and there may be other use cases, but I am not aware of any at this time.

This implementation is a straight-forward extension of currently methods for face connections, which currently supports folding along vorticity (corner) cells, as in MOM6 and MOM5 configurations. Other folding formulations would additionally require changing the pivot point (by default the (Nx/2,Ny) vorticity point) and/or shifting indices.

## Changes

- **API**: connection tuples gain an optional 4th element `fold` → `(face, axis, reverse, fold)`, defaulting to `False`. All existing 3-tuple connections (cubed-sphere / LLC) are unchanged.
- **`padding.py`**: when `fold=True`, `_pad_face_connections` mirrors the exchanged halo along the in-seam (tangential) axis — inferred as the other connected (horizontal) axis, so it never touches Z/time — and reverses the sign of vector components (the seam is a 180° pivot about the pole).
- **`grid.py`**: validation tolerates the 4-tuple, checks `fold` reciprocity, and treats a fold like `reverse` for same-side edge matching.
- **docs**: documented the `fold` element in `grid_topology.md`.

A tripolar grid is then expressible as two full-height faces split in X, joined periodically in X and by the fold along their northern edge:

```python
F = False
face_connections = {"face": {
    0: {"X": ((1,"X",F),(1,"X",F)), "Y": (None,(1,"Y",F,True))},
    1: {"X": ((0,"X",F),(0,"X",F)), "Y": (None,(0,"Y",F,True))},
}}
```

## Testing

- New tests: scalar fold (X-mirror), vector fold (sign-flip), 1-face self-fold, and end-to-end `diff`/`interp`.
- Full suite: **4183 passed, 0 failed**; existing cubed-sphere/LLC/reverse tests unchanged.
- Verified against real MOM6 OM4p25 output: the fold halo reproduces the geometric neighbour exactly, and MOM6's own seam transport satisfies `vmo = −vmo[::-1]`, confirming the vector sign convention.

A companion notebook section (MOM6 tripolar example) is prepared in the `xgcm-examples` submodule repo as a separate PR: https://github.com/xgcm/xgcm-examples/pull/23

### DIFF
--- a/docs/grid_topology.md
+++ b/docs/grid_topology.md
@@ -207,6 +207,22 @@ connection) or a three element tuple with the following contents
 face to the left edge of another face. A reversed connection connects
 left to left, or right to right.
 
+An optional fourth element specifies a **fold** connection
+
+```
+(<FACE DIMENSION VALUE>, `<AXIS NAME>`, <REVERSE CONNECTION>, <FOLD>)
+```
+
+`<FOLD>` is a boolean (default `False`). A fold joins two same-side edges of the
+*same* axis (e.g. the northern edge of one face to the northern edge of another)
+and mirrors the exchanged halo along the *tangential* (in-seam) axis. This is the
+connection needed for the **bipolar Arctic seam of a tripolar grid** (as used by
+MOM6, NEMO and MOM5), where the northernmost row folds onto itself so that
+column `i` is identified with its mirror column `N − i`. For vector fields both
+horizontal components reverse sign across the fold (the seam acts as a 180°
+pivot about the pole). See the
+[MOM6 tripolar example](xgcm-examples/03_MOM6.ipynb) for a worked case.
+
 !!! note
     We may consider adding standard `face_connections` dictionaries for common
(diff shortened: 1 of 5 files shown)


hdrake:
Closes https://github.com/xgcm/xgcm/issues/194

VeckoTheGecko:
Hi @hdrake , I wont be able to review this since im not familiar with the topic (or Eulerian oceanography for a large part). I'm also just a contributor and not a maintainer.

I think @jbusecke would be a better reviewer

PR #711, opened 2026-06-02
Bipolar north-fold boundary condition for tripolar grids

### LINKED ISSUE #194: does xgcm handle models with a tri-polar?

hdrake:
Adds a **bipolar north-fold boundary** to xgcm (closes #194), expressed as a
single per-axis boundary on one grid tile:

```python
Grid(ds, coords={...}, boundary={"X": "periodic", "Y": {"fold": "corner"}})
```

**Terminology.** The global grids this serves are **tripolar** — they carry
three coordinate singularities: the ordinary South Pole plus two poles displaced
over Arctic land. The *fold itself* is **bipolar**: the northern edge of the
single logical tile folds onto itself along the **bipolar seam**, the line
joining the two northern poles. So "tripolar" describes the grid; "bipolar"
describes this fold.

The fold mirrors the periodic **seam** (zonal) axis about the pole and
**sign-flips** vector components, reproducing the north-fold conventions of
MOM6, NEMO, MOM5, and Oceananigans. The pivot names the sublattice the pole sits
on: `center`/`T`, `corner`/`F`, `U`, `V` (or an explicit `{axis: position}`
mapping). These four cases are exhaustive (each axis is either a cell center or
a cell edge). Only the **north** edge folds; the south edge of a fold axis takes
a per-call `boundary` override when given, otherwise a configurable `south` mode
(default `fill`).

### What's here
- **`xgcm/padding.py`** — fold padding kernels + pivot resolution; **`grid.py`** —
  fold validation and seam-axis inference. Unsupported combinations now fail with
  clear errors rather than cryptic ones (a fold combined with `face_connections`;
  an `inner` seam position under a center-type pivot, which has no mirror
  partner).
- **`xgcm/test/test_fold.py`** — unit tests over all pivots/positions, the vector
  sign flip, multi-row halos, simultaneous seam padding, dask parity, the
  per-call south-edge override, a **known-answer `interp`/`diff`-across-the-seam**
  test (hand-folded numpy reference, scalar and vector), and the guard errors.
- **docs**: a new **"The Bipolar North Fold (Tripolar Grids)"** section in
  `grid_topology.md` with two publication-quality schematics (the four pivot
  conventions, and how the halo is filled by mirror + sign flip), generated by
  `scripts/make_fold_schematics.py`; adds `05_tripolar_fold` to the nav and bumps
  the `xgcm-examples` submodule (xgcm/xgcm-examples#26).

### Examples (xgcm/xgcm-examples#26, via the submodule bump)
- `05_tripolar_fold.ipynb` — demonstrates the fold directly in **logical
  grid-index space** (no map projection, so the seam is just a horizontal line)
  on three real models — **MOM6 (GFDL-CM4), NEMO eORCA1 (IPSL-CM6A-LR), and
  Oceananigans**. It shows `interp` across the seam (surface speed stays
  continuous), the `u`/`v` component **sign flip** (the vector fold), a cross-seam
  transect, and `diff` across the seam (horizontal divergence — a true scalar —
  stays continuous). In each case the fold halo continues the real field while a
  naive boundary smears the edge, with the operator corrected exactly along the
  seam and untouched elsewhere.

### Sync with master
- Merged current `master` into the branch. The two new upstream commits were the
  source of the conflicts: #730 removed the now-deleted `05_autogenerate` nav
  entry, and #731 bumped the `xgcm-examples` submodule to the **modernized 01–04**
  examples. Resolved by dropping `05_autogenerate`, **renumbering the fold example
  `06`→`05`** (nav + `grid_topology.md` link), and pointing the submodule at the
  rebased xgcm-examples#26 commit (modernized 01–04 + `05_tripolar_fold`).
- All five example notebooks were re-executed against this branch's build (0 cell
  errors).

> The `docs/xgcm-examples` submodule pointer references xgcm/xgcm-examples#26;
> that PR should merge first so the pointer resolves on `master`.

---

### Review pass (latest commit)
An independent review pass added two small hardening/clarity fixes:
- `padding.py` now **validates the position strings** in an explicit `{axis: position}` fold pivot — a typo (e.g. `"centre"`) was previously coerced silently to an edge pivot (wrong halo, no error); it now raises with the allowed positions (with a test).
- `grid_topology.md` clarifies that a vector is folded by passing it as a single-key dict (that is what flags the sign flip); the 180° fold needs no partner, so `other_component` is optional on the fold path.
- Bumped the `xgcm-examples` submodule to the reviewed notebook commit.

---

**Updated 2026-07-12:** merged current `master` and reconciled this branch's new north-fold boundary API to master's `padding` vocabulary — the fold-spec dict (e.g. `{"fold": "corner"}`) is now a `padding` value, so the grid is built with `padding={"X": "periodic", "Y": {"fold": "corner"}}`. Renamed the internal fold surface accordingly (`_parse_fold_boundary`→`_parse_fold_padding`, `_is_fold_boundary`→`_is_fold_padding`, `Axis._boundary`→`_padding`) and adapted to master's removal of the `periodic` argument (the seam axis is now inferred from an axis whose `padding` is `"periodic"`; an unspecified axis is `padding=None`, i.e. no boundary). Updated `test_fold.py`, `grid_topology.md`, and `scripts/make_fold_schematics.py` to the `padding=`/`padding_width=` keywords. Full test suite green.

---

### AI Usage Disclaimer

I used Claude Code (Opus 4.8) to implement this PR, which supersedes an earlier attempt (#707) based on using face connections with reversed seams. I started by prompting Claude in "Plan Mode" with issue #194 and my previous PR, saying that I wanted to instead implement the fold as an extension to the `padding` API (which at the time was called `boundary`). The plan involved testing and me manually reviewing the results of simple tests and a comprehensive documentation notebook that applied the method to real example datasets and included visualizations of the core `_pad` functionality across the fold seams. I iterated quite a bit with Claude to iron out the plan and as a result it pretty much worked out of the box in the initial commit. Most of the follow-up commits are just improving the docs, adding test coverage, and resolving merge commits with master.


hdrake:
**Update — refinements from review (commits `45663dd`, `371db1c`)**

Two correctness/usability improvements plus a small doc clarification:

**1. Seam-axis inference now keys off *explicitly* periodic axes** (`grid.py`)
Previously the seam was inferred as "the unique other periodic axis," but unspecified axes default to `periodic` — so on a realistic 3D grid (`boundary={"X": "periodic", "Y": {"fold": ...}}`) the vertical `Z` silently counted as a second periodic candidate and fold construction raised *"ambiguous."* The seam is now drawn only from axes the user **explicitly** marked periodic (via `boundary={ax: "periodic"}` or the `periodic` list/dict). Net effect:
- A vertical (or any other) axis can simply be **left unspecified** — it no longer needs to be forced non-periodic to disambiguate the seam.
- `> 1` explicitly-periodic non-fold axis still raises *"ambiguous"*; `0` raises a clear *"requires an explicitly periodic seam axis"* error.
- Note: the bare `periodic=True` *default* is intentionally not treated as "explicit" (it's indistinguishable from the argument default).

**Bundled latent-bug fix:** making the above reachable surfaced a pre-existing crash — a partial `boundary` dict that omits an axis raised `KeyError` during boundary parsing *before* fold validation ran. Changed that lookup to fall back to the default for omitted axes (`boundary_dict.get(ax)`); purely additive (full-dict / scalar / `None` inputs behave identically, no test relied on the `KeyError`).

**2. North-fold halo now errors instead of silently under-filling** (`padding.py`, `_fold_north_halo`)
A requested halo width exceeding the available interior rows used to silently clamp, producing a too-short array with no error. It now raises a clear `ValueError` naming the axis, requested width, and max available rows.

**3. Doc comment** in `_pad_fold` explaining why no component swap is needed: a 180° point reflection sign-flips both vector components without swapping axes, unlike a 90° rotation (e.g. a cube-sphere edge) which would need `other_component`.

Tests added in `test_fold.py` for the defaulted-vertical seam case, the two-explicit-periodic ambiguity, and the over-wide halo error.

xgcm/axis.py
```diff
@@ -34,7 +37,7 @@ def __init__(
             Mapping[str, str]
         ] = None,  # TODO type hint as Literal of the allowed options
         boundary: Optional[
-            str
+            Union[str, Mapping]
```
hdrake, line 40: The Docstring should be updated accordingly.

xgcm/padding.py
```diff
@@ -360,6 +516,145 @@ def _pad_basic(
...
+        isvector = False
+
+    fold_axes = {
+        ax for ax in padding_width if ax in grid._folds and padding_width[ax][1] > 0
+    }
+
+    # 1. Attach the northern fold halo for each fold axis (computed from the
+    #    unpadded interior so the seam mirror sees the full periodic row).
+    #    Iterate in a stable (sorted) order so the concatenation is
+    #    deterministic when more than one fold axis is present.
```
hdrake: This seems unnecessarily defensive. We do not expect use cases with more than one fold axis; this can be less general and just fail loudly if there is more than one fold axis.

xgcm/axis.py
hdrake: Done in da7a853 — updated the `Axis.__init__` docstring to document the fold-spec `dict` boundary (pivot aliases + optional `south` key), matching the widened `Union[str, Mapping]` type hint.

xgcm/padding.py
hdrake: Done in da7a853 — dropped the defensive sorted-set handling; `_pad_fold` now raises `NotImplementedError` when more than one fold axis is requested. A tripolar grid has a single north fold, so multiple folds are unsupported and fail loudly.

hdrake (commented):
The changes for renaming of `boundary` to `padding` look good to me.

hdrake (commented):
I did a comprehensive line-by-line review of all of the source code changes + changes to the docs.

Looks good to me. I think the tripolar fold functionality is similarly "experimental" as `face_connections` were when they were released. Since this PR only adds new functionality and does not change/break any existing functionality, I am in favor of releasing it in a v0.10.1 minor release so we can get user feedback before an eventual v1.0.0 release.

Issue #496, opened 2022-05-07
interp change the format of time


miniufo:
Recently, I update xgcm from 0.5.0 to 0.7.0 and re-run a script doing some diagnostics using outputs from MITgcm.  The codes of loading data and assign a new time coordinates are:
```python
#%% load dataset
import numpy  as np
import xmitgcm
import xarray as xr
from xcontour.xcontour import add_MITgcm_missing_metrics

tlen = 200
dt   = 600
path = 'I:/breakingIW/nonHydroKH/'
ds = xmitgcm.open_mdsdataset(data_dir=path+'output/', grid_dir=path, delta_t=2,
                             prefix=['Stat']).isel(time=slice(tlen))

dset, grid = add_MITgcm_missing_metrics(ds, periodic=['X'],
                                        boundary={'Y':'extend','Z':'extend'})

dset = dset.assign_coords({'time':np.arange(tlen)*dt/3600.0}).astype(np.float32)
```

The default time values are `timedelta64[ns]`, which is not convenient for xarray's `dset.UVEL.plot()`, reporting error
>TypeError: The DTypes <class 'numpy.dtype[timedelta64]'> and <class 'numpy.dtype[float64]'> do not have a common DType. For example they cannot be stored in a single array unless the dtype is `object`.

So I change it to `dtype=np.float32`.

**What happened**:
After using `xgcm`'s core functions, possibly `grid.interp()`, the `time` of interpolated variables recovers to `timedelta64[ns]`.

**What you expected to happen**:
`dtype` of `time` do not change after applying xgcm's functions.

**Anything else we need to know?**:
I also get a warning during running the interpolation:
>C:\ProgramData\Anaconda3\lib\site-packages\xgcm\grid.py:1637: UserWarning: Metric at ('time', 'Zl', 'YC', 'XG') being interpolated from metrics at dimensions ('Z', 'YC', 'XC'). Boundary value set to 'extend'.

This does not show up in previous version of `xgcm`.  Just want to know what happens here and where the warning is triggered.

**xgcm version and environment**:
0.7.0

jbusecke:
Hey @miniufo, 

I think there are several things going on here: 

> C:\ProgramData\Anaconda3\lib\site-packages\xgcm\grid.py:1637: UserWarning: Metric at ('time', 'Zl', 'YC', 'XG') being interpolated from metrics at dimensions ('Z', 'YC', 'XC'). Boundary value set to 'extend'.
[...]

PR #719, opened 2026-06-27
Fix #575: preserve non-core input coordinates (e.g. time) across grid ufuncs

### LINKED ISSUE #575: [BUG] Time coordinate being dropped by apply_as_grid_ufunc

hdrake:
## What

Grid operations (`interp`, `diff`, etc.) silently dropped non-core coordinates that ride on the **input data** but are not stored on the grid's dataset — the classic case being a `time` coordinate on the array being operated on. Fixes #575.

## Why

Padding strips **all** coordinates before applying a grid ufunc (`xgcm/padding.py`, `_strip_all_coords`). They were then restored in `apply_as_grid_ufunc` by `_reattach_coords`, which only looked them up on `grid._ds.coords`. Any coordinate present on the input `DataArray` but absent from the grid's stored dataset was therefore never reattached and was lost.

## How

- `apply_as_grid_ufunc` now collects non-core coordinates from the input arguments (unpacking vector components) and passes them to `_reattach_coords`.
- `_reattach_coords` restores those input coordinates as long as **all** of their dimensions survive in the output. `grid._ds` still takes precedence on name conflicts, so transformed core-dimension coordinates are never clobbered.

## Tests

- Added `test_non_core_coord_on_input_is_preserved` in `xgcm/test/test_grid_ufunc.py`, parametrized over numpy and dask backends. It verifies a `time` dimension coordinate and a non-dimension coordinate (`label`) both survive a `diff` with padding, with original values and dtype intact.
- Confirmed the test fails on `master` and passes with this change.
- Full suite green (`pixi run tests`): 4181 passed.

## Changelog

Added a Bugfixes entry under v0.9.0 in `docs/whats-new.md`.

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -46,6 +46,12 @@
 
 ### Bugfixes
 
+- Fix grid operations (`interp`, `diff`, etc.) silently dropping non-core coordinates that ride on the
+  input data but are not stored on the grid's dataset (e.g. a `time` coordinate). Padding strips all
+  coordinates and they were only restored from `grid._ds`; such coordinates are now also restored from
+  the input arguments. ([#575](https://github.com/xgcm/xgcm/issues/575)).
+  By [Henri Drake](https://github.com/hdrake).
+
 - Fix bug in `xgcm.transform.transform` that violated tracer conservation when using conservative interpolation in the presence of nans. ([#635](https://github.com/xgcm/xgcm/pull/635))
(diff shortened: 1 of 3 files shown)


hdrake:
Superseded by #721, which fixes both #575 and #496 with a single, more complete `_reattach_coords` change (the grid owns position-shifted core dims; the input array owns non-core dims). #575's regression test has been carried over to #721. Closing in favor of the bundled PR.

PR #705, opened 2026-04-28
Bugfix vector diff/interp on grids with face connections

anthony-meza:
On grids with face connections, vector components require calls like `grid.diff({"X": U}, "X", other_component={"Y": V})` in order to full utilize the grid information. At the moment, this is **not** equivalent to `grid.diff(U, "X", other_component={"Y": V}).` As I understand it, passing `{"X": U} `tells xgcm that `U` is a vector component, which allows it to use `V` when padding across connected faces.

`_map_func_over_core_dims` in `grid_ufunc.py` was still assuming that the input was always a DataArray (instead of a dict, i.e., `{"X": U}`) when (1) setting up the chunked overlap step, and  (2) when reading chunk sizes after padding. This change unwraps the DataArray from that one-item dictionary before those steps, so vector diff/interp for vector components behaves correctly on grids with face connections instead of failing with `AttributeError: 'dict' object has no attribute 'variable'`.

 - [x] Closes #704

PR #720, opened 2026-06-27
Fix grid ufunc crash on dask-backed vector components (#581)

### LINKED ISSUE #581: [BUG] Issue in grid_ufunc when original_arg is a dictionary

hdrake:
## Summary

Fixes #581: applying a vector grid ufunc (`diff_2d_vector`, `interp_2d_vector`, or the equivalent vector-component `Grid.diff`/`Grid.interp`) crashes on grids **without** face connections when a vector component is supplied as a `{axis_name: DataArray}` dict — for both numpy- and dask-backed arrays.

## Root cause

`diff_2d_vector`/`interp_2d_vector` always wrap each component as `{axis_name: DataArray}`. In `xgcm/padding.py`, `pad()` strips coords (preserving the dict) and then dispatches:

- the **face-connection** path (`_pad_face_connections`) unpacks the dict (it needs the axis to orient the component), but
- the **basic** path (`_pad_basic`) received the dict unchanged and called `da.copy(deep=False)`, raising `TypeError: dict.copy() takes no keyword arguments`.

The originally reported `AttributeError` (`'dict' object has no attribute 'variable'`/`'transpose'`) in `grid_ufunc.py` was already fixed by #705; with that fixed, the reproducer now surfaces this padding crash, which occurs before the grid-ufunc apply step and so affects numpy and dask alike.

## Fix

Unpack the inner `DataArray` from the vector-component dict before calling `_pad_basic`, mirroring what `_pad_face_connections` already does. Minimal and confined to the non-face-connection branch of `pad()`; no change to non-dict or face-connection behavior.

## Test

Adds `test_2d_vector_dict_input_no_face_connections` in `xgcm/test/test_grid.py`, parametrized over `diff_2d_vector`/`interp_2d_vector`, `boundary` in `{fill, extend}`, and numpy vs. dask (chunked). It computes both returned components and asserts no NaNs. Verified it fails on master (8/8 cases) and passes with this fix. Full suite green (4192 passed; the 3 unrelated pre-existing `test_tripolar_realdata.py` failures are not affected).

Closes #581

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -46,6 +46,13 @@
 
 ### Bugfixes
 
+- Fix `TypeError: dict.copy() takes no keyword arguments` (and the `AttributeError` reported for dask
+  arrays) when applying vector grid ufuncs (e.g. `diff_2d_vector`, `interp_2d_vector`) on grids *without*
+  face connections, where a vector component supplied as a `{axis_name: DataArray}` dict was passed
+  unchanged to the basic padding routine. The dict is now unpacked before basic padding
+  ([#581](https://github.com/xgcm/xgcm/issues/581)).
+  By [Henri Drake](https://github.com/hdrake).
+
 - Fix `diff_2d_vector`/`interp_2d_vector` (and the equivalent vector-component `Grid.diff`/`Grid.interp`)
   on grids with face connections when a vector component is a dask array chunked into more than one chunk
   along its core dimension. The `map_overlap` path now derives the output chunk spec from the padded,
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -46,10 +46,11 @@
 
 ### Bugfixes
 
-- Fix `TypeError: dict.copy() takes no keyword arguments` (and the `AttributeError` reported for dask
-  arrays) when applying vector grid ufuncs (e.g. `diff_2d_vector`, `interp_2d_vector`) on grids *without*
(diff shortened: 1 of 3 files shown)


hdrake:
**Review summary.** One independent review round; no blocking issues found. The reviewer confirmed the root cause is in `padding.pad()` (the non-face-connection branch handed the `{axis: DataArray}` vector dict straight to `_pad_basic`, hitting `dict.copy()` `TypeError`), that unpacking loses no information on grids without face connections (the partner component is only needed for seam/fold sign handling), and that the dask path is unblocked. Two non-blocking suggestions were addressed in a follow-up commit: the regression test now asserts value correctness via `xr.testing.assert_allclose` (not just that it computes), and the changelog wording was corrected to describe the `pad()` fix. Targeted suite green (8 passed).

PR #721, opened 2026-06-27
Fix #496 and #575: preserve non-core input coordinates through grid operations

### LINKED ISSUE #496: interp change the format of time

### LINKED ISSUE #575: [BUG] Time coordinate being dropped by apply_as_grid_ufunc

hdrake:
## What

Grid operations (`interp`, `diff`, etc.) mishandled coordinates carried on the **input** `DataArray` that sit on non-core (not position-shifted) dimensions:

- **#496** — a coordinate present on both the input and the grid (e.g. a user-recast `time`) was **clobbered** by the grid's (possibly stale) copy.
- **#575** — a coordinate present on the input but **not** on the grid (e.g. a `time` coordinate added to the data) was **dropped** entirely.

## Why

Padding strips all coordinates before applying a grid ufunc, and `_reattach_coords` restored them **only** from `grid._ds.coords`. So input-only coords were never restored (#575), and input coords that also existed on the grid were overwritten with the grid's version (#496).

## How

`apply_as_grid_ufunc` now passes the input arguments' coordinates to `_reattach_coords`, which:
- restores coordinates on **non-core** dims from the **input array** (first input wins for repeated names), and
- still takes the newly position-shifted **core-dim** coordinate from `grid._ds`.

This single change resolves both issues.

## Tests

- `test_preserve_input_noncore_coords` (`test_grid.py`) — interp/diff × numpy/dask: a user-set non-core coord is preserved, a stale core-dim coord is not carried over (#496).
- `test_non_core_coord_on_input_is_preserved` (`test_grid_ufunc.py`) — numpy/dask: a `time` dim-coord and a non-dim coord present on the input but absent from the grid both survive (#575).
- Both confirmed fail-before / pass-after; targeted suites green.

## Note

Supersedes #719 (which fixed #575 alone). Closes #496 and #575.

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -46,12 +46,12 @@
 
 ### Bugfixes
 
-- Grid operations (e.g. `Grid.interp`, `Grid.diff`) no longer drop or clobber non-core coordinates carried
-  on the input `DataArray`. Padding strips all coordinates and they were only restored from the grid's own
-  dataset, so a coordinate that lived on the input but not on the grid (e.g. a `time` coordinate) was lost,
-  and a coordinate present on both was overwritten with the grid's (possibly stale) copy. Coordinates on
-  non-core dimensions are now preserved from the input array (first input wins for repeated names), while
-  the newly position-shifted core-dim coordinate still comes from the grid
+- Grid operations (e.g. `Grid.interp`, `Grid.diff`, `Grid.cumsum`) no longer drop or clobber non-core
+  coordinates carried on the input `DataArray`. Padding strips all coordinates and they were only restored
+  from the grid's own dataset, so a coordinate that lived on the input but not on the grid (e.g. a `time`
+  coordinate) was lost, and a coordinate present on both was overwritten with the grid's (possibly stale)
+  copy. Coordinates on non-core dimensions are now preserved from the input array (first input wins for
+  repeated names), while the newly position-shifted core-dim coordinate still comes from the grid
   ([#496](https://github.com/xgcm/xgcm/issues/496), [#575](https://github.com/xgcm/xgcm/issues/575)).
 
 - Fix `TypeError: dict.copy() takes no keyword arguments` when applying vector grid ufuncs (e.g.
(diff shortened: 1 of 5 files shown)


hdrake:
**Review summary.** One independent review round; no blocking issues found. The reviewer verified that user-set non-core coordinates (e.g. a re-cast `time`) now survive grid ops with correct dtype and values, while the position-shifted core-dim coordinate still comes from `grid._ds`; and checked the core+non-core spanning-coord case, vector-dict inputs, the cumsum/pad-after path, and that the `#575`/#719 bug is not reintroduced. Two non-blocking suggestions were addressed in a follow-up commit: multi-arg coord precedence now uses first-arg-wins (`setdefault`, consistent with #719) and the regression test was strengthened. `test_grid.py` green (488 passed). Note: this PR overlaps `_reattach_coords` with #719 (#575) and will need to be reconciled with it on merge (see PR description).

hdrake (commented):
Looks good to me.

PR #739, opened 2026-07-03
MAINT: Bump Pixi version in CI

VeckoTheGecko:
Latest Pixi (0.72.0) wasn't working with the `0.4.*` build backends. This PR updates to the `0.7.*` backend and also bumps the pixi version used in CI.

I think we can remove the PIXI_VERSION lock in CI entirely, but I'll leave that decision up to you.

 - [x] Closes None

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  PIXI_VERSION: "v0.63.0"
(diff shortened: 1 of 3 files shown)


hdrake:
@VeckoTheGecko , I'm new to Pixi so still learning the ropes. I tried resolving the merge conflicts but found the ReadTheDocs build broke, so I bumped the ReadTheDocs pixi pin accordingly.

Let me know what you think.

VeckoTheGecko:
Diff looks good @hdrake

Issue #734, opened 2026-06-30
Claude-accelerated development and maintenance of xgcm


hdrake:
While @NoraLoose, @raphaeldussin, and I (@hdrake) have taken over responsibilities to maintain (and, hopefully, continue to develop) `xgcm`, we are all very busy and the backlog of issues and PRs is daunting.

In recent weeks, I have been experimenting with using Claude code to facilitate the maintenance and development of xgcm (and xgcm-examples). I have found that it does a great job, although PR reviews require even more scrutiny than before. Here are some example PRs that I coauthored with @claude, which were mostly straight-forward bug fixes, workflow fixes, and improvements to the docs:
- https://github.com/xgcm/xgcm/pull/722
- https://github.com/xgcm/xgcm/pull/723
- https://github.com/xgcm/xgcm/pull/725

I am curious to hear what other maintainers, contributors, and users think about this.

What are best practices for collaborating with AI agents on an open-source package like this?

If these is somewhat of a consensus that this is a good idea, I would like to follow up by making a PR for a `Claude.md` and any other supporting documents, such as instructions for dedicated agents to do things like review PRs, come up with new human-readable regression tests, etc.

hdrake:
Opened #755 as a first concrete step on this — it adds a `CLAUDE.md` that encodes xgcm's engineering norms (deprecate-by-removing, fail-loudly, minimal core deps, testing across the four standard grid topologies, EffVer) and a *Pull request guidelines* section in the contributor guide covering PR conventions, changelog discipline, and open disclosure of AI assistance.

It's written to **complement** the human contributor guide rather than duplicate it, and to be **durable** (pointers to `pyproject.toml`/the guide instead of drift-prone specifics). Deliberately scoped to just the `CLAUDE.md` + norms — the other supporting docs floated above (dedicated PR-review and regression-test agents) can follow as separate PRs once there's consensus.

Would love maintainer/contributor eyes on the norms themselves (@NoraLoose, @raphaeldussin) — the goal is to write down the review-enforced-but-unwritten conventions so an AI (and new human contributors) follow them by default.

🤖 _Drafted by [Claude Code](https://claude.com/claude-code); reviewed and posted by @hdrake._

hdrake:
Also curious to hear from @VeckoTheGecko since you seem to also be using Claude to contribute!

VeckoTheGecko:
[...]

Issue #733, opened 2026-06-30
Road to xgcm 1.0.0


hdrake:
This is the tracking issue for **xgcm's first stable release, v1.0.0**. It is the single source of truth for 1.0 scope; the [`v1.0.0` milestone](https://github.com/xgcm/xgcm/milestone/3) mirrors this checklist.

A 1.0 tag is a contract: *this public API is stable, and we will not break it without a major version bump.* The scope below is deliberately **lean — "stabilize and freeze"**: finish the deprecations already in flight, fix the core-semantics bugs, land the two nearly-done features we want stable from day one, and reconcile the docs/changelog/API surface. Larger API-redesign debates are intentionally **deferred to 1.x** (listed at the bottom) so the freeze isn't held hostage to open discussions.

Supersedes/absorbs the older roadmap thread #414 (now closed).

---

## A. Finish the announced deprecations
Three deprecations are announced in the `Grid` constructor but not yet executed; 1.0 is where they get finalized. No warning should point at a kwarg that doesn't exist.
- [x] `boundary` → `padding` rename — #696 (supersedes #578). Fixes the broken intermediate state in #678.
- [x] Deprecate `periodic` input — #746 (supersedes #626/#534), in review.
- [ ] `fill_value` default → `nan` (or explicitly defer with a documented reason) — relates to #567.
- [x] Resolve the `keep_coords` deprecation — #382. *Done in #745.*

## B. Fix the core-semantics bug cluster
User-visible contradictions in `diff` / `cumsum` / `cumint` / boundary handling. The API can't be frozen as "correct" while these stand.
- [x] #604 — default boundary periodic even when specified otherwise
- [x] #624 — non-intuitive behavior if `periodic` not set as a list
- [x] #625 — cumsum applies periodic B.C. when it shouldn't
- [ ] #605 — cumsum flips axis
- [x] #609 — direction of integration for cumint
- [ ] #485 — boundary inconsistent between `grid.diff` and `grid.cumsum`
- [x] #509 — diff on non-periodic axis behaves periodic
- [ ] Land #729 (reverse `cumsum`/`cumint`, supersedes #256) and #728 (axis directionality, implements #337) — both intersect #605/#609. Both rebased on master and ready for review.

## C. Land the bipolar north-fold and promote topology to stable
- [x] #711 — bipolar north-fold boundary condition for tripolar grids (#194). Already reviewed and well-tested.
[...]

PR #696, opened 2026-03-03
Rename boundary to padding across the codebase

VeckoTheGecko:
This PR renames `boundary` to `padding` (and `boundary_width` to `padding_width`) across the codebase's public API according to the following deprecation guidance.

```md
When doing deprecations for renaming of variables, I want you to do the following.

## function signatures (also applies to class initialisers, and methods)

- Do the renaming of the variable in the function
- Provide a **kwargs in the function signature if not already there
- first item in the method should be checking if the old variable name is in the kwargs. If it is, raise a value error mentioning that the item is deprecated "Argument '{variable name}' has been renamed to ...". Then delete this old name from the kwargs dict.
- Add a comment `# TODO - remove deprecation handling`

## attributes

For class attributes:
- Do the renaming
- Add a property with the old name that when used raises the error "Attribute '{variable name}' has been renamed to ..."
- Add a comment `# TODO - remove deprecation handling`
```
I think this deprecation policy is the best way to go (it chooses explicit errors with very informative error messaging, rather than using deprecation cycles like in Xarray which would come with it significant maintenence burden with limited additional benefit).

I also updated the local variables/private API accordingly.

 - [x] Closes https://github.com/xgcm/xgcm/issues/678
 - [x] Tests added (added `test_deprecations.py`)
   - I only focussed here on the grid initialiser. Should I add the axis initialiser too? I imagine that testing for informative error messages across the entire API would be a bit overkill?
 - [x] Passes `pre-commit run --all-files`
 - [x] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
   - Let me know if this is sufficient, or if you want me to document all the public API that was changed

## pytest summary output

on master:

```
= 4179 passed, 2 skipped, 50 xfailed, 8 xpassed, 13344 warnings in 70.19s (0:01:10) =
```
on this PR:
```
= 4181 passed, 2 skipped, 50 xfailed, 8 xpassed, 13177 warnings in 71.45s (0:01:11) =
```
so the amount of warnings went down (likely due to fixing #678 

## LLM usage

I used AI in the following ways

- `claude 'Find all the public API functions, methods and classes that use "*boundary*" as either argument names or variable names. I want you to then document these in a separate file (note the file, and the item).'`
- Then I asked claude to do these renamings according to the deprecation policy above (without modifying the test suite at all or running it)
- Then I asked claude to update the test suite based on the test suite output

There was also quite a bit of manual iterations to help it along as claude missed a bunch of renamings etc.

Here is an 

[...]

PR #746, opened 2026-07-03
Remove the `periodic` argument from `xgcm.Grid`

### LINKED ISSUE #195: boundary="periodic"?

### LINKED ISSUE #509: [BUG] Diff in non periodic axis seems to be periodic
**What happened**:

When applying a diff operation on a simple example (modified from the doc https://xgcm.readthedocs.io/en/latest/grids.html), it seems that the X axis is considered as periodic even if stated otherwise.

In our minimal example, we apply the diff operation on a small array [2, 4, 5, 8, 10, 12, 14, 16, 18]. We expected to have [np.nan, 2, 2, 2...] or an array with one less element, but we got [-16, 2, 2, 2...] instead. It seems that the last element is taken to extend the boundary at the left, whereas it has no relation with the first element whatsoever.

**Minimal Complete Verifiable Example**:

```python
import xarray as xr
import xgcm
import numpy as np

ds = xr.Dataset(
	coords={
		"x_c": (
			["x_c"],
			np.arange(1, 10),
		),
		"x_g": (
			["x_g"],
			np.arange(0.5, 9),
		),
	}
)

da = ds.x_c*2
grid = xgcm.Grid(ds, coords={"X": {"center": "x_c", "left": "x_g"}}, periodic=False)

grid
#>> <xgcm.Grid>
#>> X Axis (not periodic, boundary=None):
#>>  * center   x_c --> left
#>>  * left     x_g --> center

grid.diff(da, "X")
#>> <xarray.DataArray 'x_c' (x_g: 9)>
#>> array([-16,   2,   2,   2,   2,   2,   2,   2,   2])
#>> Coordinates:
#>>   * x_g      (x_g) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5
```

xgcm 0.7.0

<details>

</details>

### LINKED ISSUE #604: [BUG] Default boundary condition is periodic even when specified otherwise

### LINKED ISSUE #624: [BUG] Non-intuitive behavior if periodic not set as list

### LINKED ISSUE #625: [BUG] cumsum seems to apply periodic B.C. even when it knows that is not the case.

hdrake:
## What

Removes the `periodic` argument of `xgcm.Grid`. Boundary behavior is now controlled exclusively by `boundary`. Passing `periodic=` raises an informative `ValueError` pointing at the replacement (per the deprecation policy from #696).

## Why

`periodic` defaulted to `True`, silently making every axis periodic. That default was the root of several boundary-condition bugs (#509, #604, #624) and created confusion between the `periodic` string/list forms and `boundary` (#624). Removing it makes boundary handling explicit and non-periodic by default. Part of the v1.0.0 push (#733, roadmap §A/§B); scope was discussed there and in #734, and the related `boundary`→`padding` rename is kept separate in #696 to keep this diff reviewable.

## New default semantics (maintainer-review point)

An axis with no `boundary` specified now applies **no boundary condition** (`boundary=None`, which is what the docstring always claimed). Operations that need to pad such an axis raise an informative error instead of silently wrapping; an op-level `boundary=` still works as before. Pass `boundary='periodic'` at grid construction to recover the pre-1.0 behavior. This also fixes cases where a declared non-periodic axis could still wrap.

**Face-connection carve-out** (from review): a **fully connected** edge does not require a boundary — for grids like a global cubed sphere, where every padded edge's halo is overwritten by connection data, operations work without any `boundary` (matching master's numerics). The informative error still fires when a *genuinely unconnected* edge of a partially-connected grid has no boundary at either the grid or op level.

## Migration

- `periodic=True` → `boundary='periodic'`
- `periodic=False` → `boundary='fill'` (the previous implicit mapping)
- `periodic=['X']` → `boundary={'X': 'periodic', 'Y': 'fill'}` (per-axis dict)
- Fully face-connected grids need no boundary; otherwise rely on the default only if every operation either doesn't pad or is called with an explicit `boundary=`.

## Tests

Migrated the whole suite and the executed docs off `periodic=`, preserving each test's intent (grids that relied on the old periodic default got explicit `boundary="periodic"` to preserve exact numerics). Added regression tests for the removal `ValueError` and for #509/#604/#624; `test_diff_interp_cubed_sphere` again builds its grid with **no** boundary (as on master) plus a hand-derived connected-halo assertion, and `test_unconnected_edge_without_boundary_raises` covers the partially-connected error path. Also fixes a latent test bug where `test_metrics_ops` parametrized boundary as the strings `"True"`/`"False"` (both silently mapped to `fill`). The migration also covers `test_conservative_transform_explicit_target_dim`, added on master after this branch was cut (it still passed `periodic=False`, which broke CI post-merge).

As part of the AI iteration described under *How AI was used*, adversarial review passes were used to check that the test-collection delta vs master is fully accounted for (coverage improved, none lost), that the op-level `boundary=` override still works at runtime, and that the #509/#604 regression test fails on master. These informed — but do not replace — the line-by-line human review I'm completing separately (to be summarized in a comment below).

Full suite green (`pixi run tests`), `pixi run lint` clean, `pixi run -e docs docs-build` clean.

Closes #195, closes #509, closes #604, closes #624, closes #625. Supersedes #626 — earlier work by @jbusecke.

## How AI was used

I used [Claude Code](https://claude.com/claude-code) (Fable 5 model) to draft this PR. I pointed it towards @jbusecke 's earlier PR #626 and asked it to pick up that work but merging in the current upstream/master. Since this PR is conceptually pretty straight-forward but requires a large diff, I think it is a good candidate for off-loading to AI. It helps that there are multiple issues summarizing the motivation for this PR and an entirely human-written PR #626 that started the work.

I iterated with it a little bit, e.g. by asking it to have independent review agents review it to make sure no instances of `periodic` were missed. Most of these changes were done in a single commit, with only minor iterations later to add guards and tests. I reviewed the summaries of code changes as Claude was making them, but will conduct a full line-by-line review separately later (and summarize that in a comment on this PR).

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.
    Tools: Claude Code (Claude Fable 5)


xgcm/test/test_grid.py
```diff
@@ -384,7 +452,9 @@ def test_keep_coords_removed():
 
     ds, coords, metrics = datasets_grid_metric("B")
     ds = ds.assign_coords(yt_bis=ds["yt"], xt_bis=ds["xt"])
-    grid = Grid(ds, coords=coords, metrics=metrics, autoparse_metadata=False)
+    grid = Grid(
+        ds, coords=coords, metrics=metrics, boundary="fill", autoparse_metadata=False
+    )
```
hdrake, line 630: Why does this get `boundary="fill"` while R422-428 get `boundary="periodic"` if both are meant to replicate the original default behavior?

hdrake: Good catch — inconsistent, but harmless. On master both tests relied on the bare default (`periodic=True`). `test_keep_coords` was migrated to `boundary="periodic"` because it actually runs padding ops whose numerics depend on it. `test_keep_coords_removed` never pads, though — the `keep_coords` guard raises at the top of `diff`/`cumsum`/`apply_as_grid_ufunc`, before any boundary handling — so `"fill"` here was arbitrary. Switched it to `boundary="periodic"` to match the sibling and the default→periodic migration rule.

🤖 _AI-generated with [Claude Code](https://claude.com/claude-code); reviewed by @hdrake._

hdrake:
I reviewed all of the line-by-line changes as well as the updated docs. This looks good to me.

PR #642, opened 2024-10-11
Transform target dim

NoraLoose:
Still WIP

 - [ ] Closes #614
 - [ ] Documentation added
 - [ ] Tests added (WIP)
 - [ ] Passes `pre-commit run --all-files`
 - [ ] User visible changes (including notable bug fixes) are documented in `whats-new.rst` (WIP)

I added a section to the existing `doc/transform.ipynb` notebook to document the new capability. While working on this notebook, I also fixed pre-existing issues with this notebook including:

* Insert `autoparse_metadata=False` into `Grid()` to avoid error messages
* Correct dictionary keys for pangeo cloud data library
* Remove non-existing `xc` coordinate from Grid object to avoid existing `ValueError`
* Insert a `.squeeze()` for plotting AMOC to avoid error

Issue #614, opened 2023-06-17
[DOC] Depth to depth xgcm.Grid.transform onto terrain-following depth levels


kevinrosa:
**What were you looking for/trying to do?**:
Transform a field such as temp(x,y,z,t) from its original depths onto the non-uniform levels of a ROMS model grid.  This is a common step when preparing inputs for a nested model in which initial conditions and boundary conditions are sourced from a parent model. 

**What was missing from the docs or not clear enough?**:
The docs show how to do depth to depth transformation for 1-D spatially-uniform target depths but not for a 3-D array of depths.  From what I could tell, this does not seem to be within the standard functionality of `transform` and so requires a bit of workaround, but I could be wrong about this.

**How could the docs be improved? Additional examples, different wording, etc.**
This could either be a documentation improvement or a feature request (I chose the former because that's the easier solution).

I'd be willing to provide code for a minimal working example of how to do this within the current version of xgcm.  My method involves a 2-step transformation: first transforming source depths into destination sigma layer space, and then back into depths.

jbusecke:
Hi @kevinrosa thanks for using xgcm, and providing this example.

A minimal working example would be great to have! I am not sure if the intermediate step is actually necessary here. If you provide an example Ill see if we can get this done in one step.

rcaneill:
Hi @kevinrosa , I don't think that the intermediate step is necessary. For a study with the NEMO model, I did the opposite: go from terrain following coordinates to constant depth, in 1 step. I guess that going the other way around should work the same.
If you want to have a look, here is the code https://github.com/rcaneill/caneill-et-al-JPO-nemo-transition-zone/blob/main/scripts/diags/moc.py

Note: it is not a documentation but the code used for the paper, I am happy to answer your questions if you have some regarding this.

jbusecke:
It seems that this is an issue of documenting the feature better. Is there a way we could get a small sample dataset or have some ROMS data in the cloud to work out an example for the docs?

jbusecke:
And just in general: If you have  (or can broadcast) the inputs/outputs in the same shape except the dimension associated with the transform axis, we should be able to support any sort of 1D transformation.

Maybe @TomNicholas has some ROMS data in the cloud soon😉

kevinrosa:
[...]

PR #747, opened 2026-07-03
Support multi-dimensional targets for the conservative transform

### LINKED ISSUE #666: [FEATURE] Enable conservative vertical transform for spatially-varying vertical coordinate
#642 enables the vertical transform for a **spatially-varying** vertical coordinate, however only for the **linear** method.

We should implement the same for the conservative method.

### Notes to whoever implements this new feature

* There is already some test data in the `test_transform.py` module that can be used to verify that the multi-dimensional conservative transform works as expected: 

https://github.com/xgcm/xgcm/blob/1139fb06a43f6345d4d793a2e0bcc08e002d0cc9/xgcm/test/test_transform.py#L215-L255

* Once implemented, the following exception can be removed:

https://github.com/xgcm/xgcm/blob/1139fb06a43f6345d4d793a2e0bcc08e002d0cc9/xgcm/test/test_transform.py#L776

* Once implemented, this test which currently checks for the raising of `NonImplementedError` would also need to be removed:

https://github.com/xgcm/xgcm/blob/1139fb06a43f6345d4d793a2e0bcc08e002d0cc9/xgcm/test/test_transform.py#L1059-L1070

hdrake:
## What

Enables `Grid.transform(..., method="conservative")` to remap onto spatially-varying (multi-dimensional) target coordinates, e.g. terrain-following vertical coordinates. Previously this raised `NotImplementedError`; only the `linear`/`log` methods supported multi-dim targets (shipped in #642 / #614).

## Why

Roadmap item C2 for v1.0.0 (#733). Ocean/atmosphere models routinely use vertical coordinates that vary in the horizontal (sigma, s-coordinates, isopycnal layers); conservative remapping of extensive quantities onto such coordinates is a core use case.

## Design

- `interp_1d_conservative` now vectorizes over the final axis of multi-dim target bins (the underlying numba gufunc `(n),(n),(n),(m),(m)->(m)` already broadcasts leading dims). A uniformly-decreasing target is flipped along its last axis; mixed-orientation columns raise `ValueError`.
- `conservative_interpolation` sizes the new dimension from the target's core dim (the old `len(...)` grabbed the wrong axis for 2-D targets) and assigns per-column midpoint cell-centers as a non-dimension coordinate named after the target, matching the linear method's output shape.
- A multi-dim target without `target_dim` now raises a clear `ValueError` (validated in `_parse_target`, which also fixes a previously-dead mismatch check on the linear path).
- Works with numpy and dask, including chunked source (extra broadcast dim) and chunked multi-dim targets — verified identical to the numpy reference.

## Docs

- `docs/transform.ipynb`: the "Transforming onto a spatially varying vertical coordinate" section now closes with a **linear-vs-conservative comparison** — the vertically-integrated temperature (`thetao * dz`) is conservatively remapped onto the terrain-following layer *interfaces* and divided by the layer thickness to give a layer-mean temperature, contrasted with the linear result via side-by-side sections and difference statistics (median 0.006 °C, p95 0.074 °C over the GLORYS→ROMS example).
- The notebook's pre-existing CMIP6 section no longer executed (the cloud store now delivers variables chunked along `lev`, which `apply_ufunc(dask='parallelized')` rejects for a core dimension); fixed by rechunking the three data variables to a single `lev` chunk. Full notebook re-executed with zero cell errors; `docs-build` clean.

## Limitations

- All columns of the target must share one orientation (all increasing or all decreasing); mixed per-column orientation raises `ValueError`.
- As before, conservative transform applies to extensive quantities only.

## Tests

- The multi-dim conservative case is now a known-answer test with a hand-verified expected result plus an independent column-by-column numpy reference.
- Per-column and total conservation asserted for the spatially-varying target.
- Dedicated dask test chunks both source and multi-dim target.
- The old `NotImplementedError` test is repurposed to assert the missing-`target_dim` `ValueError`.
- Full suite green (`pixi run tests`: 4214 passed, 2 skipped, 50 xfailed, 8 xpassed), lint clean.
- Real-data validation on the GLORYS→ROMS example posted in the comments: bitwise match against a hand-rolled overlap-integral reference, column conservation to machine precision.

Closes #666.

Note on #741: resolved — #741 merged first and the merge of master into this branch (b2ba617) keeps this branch's deletion of the guard region.

---

**Updated 2026-07-12:** merged current `master` into this branch. Reconciled the two breaking API changes that landed upstream — the removal of the `periodic` argument (#746; deleted the `periodic=False` kwargs) and the `boundary`/`boundary_width` → `padding`/`padding_width` rename (#761) — across the tests and the `transform.ipynb` notebook. The multi-dim conservative feature and the `whats-new` entry are preserved. `transform.ipynb` hard-conflicted with master's #756 refresh; I kept this branch's version (for the multi-dim demonstration), ported its API usage to `padding=`, and **re-executed it against the merged code with zero cell errors**. `pytest xgcm/test/test_transform.py` passes (105 passed, 1 skipped, 2 xfailed). This master-merge/conflict-resolution step was carried out by Claude Code at my direction; I am responsible for the review per xgcm's AI Usage Policy.


hdrake:
**Real-data validation: GLORYS → ROMS terrain-following remap** (the spatially-varying example from `docs/transform.ipynb` / #642, dataset `doi:10.6084/m9.figshare.27214182.v1`).

Built the conservative analogue of the notebook's linear example: heat content `thetao·dz` (GLORYS, 43 constant-depth levels, with NaNs below topography) remapped onto the ROMS layer bins bounded by `interface_depth_rho` (`s_w`×`eta_rho`×`xi_rho` — 2040 columns, all uniformly decreasing along `s_w`, so this exercises the multi-dim flip path on real data). Source and target share the horizontal dims, i.e. the column-aligned case.

Results on this branch (merge commit b2ba617):
- Output vs an independent hand-rolled per-column overlap-integral reference on 8 random wet columns: **max abs deviation 0.0** (bitwise identical).
- Per-column conservation (sum over bins vs source integral over the bin-covered range): **rel. error ≤ 3.2e-16** (machine precision).
- Sanity vs the notebook's linear method: layer-mean conservative T vs linear T at layer centers agrees to median |Δ| = 0.006 °C (p95 = 0.07 °C) over 163k wet cells.
- Land columns (all-NaN source): all-NaN output, as expected.
- Same call on current master: `NotImplementedError` — confirming this PR enables the #666 use case end-to-end.

PR #756, opened 2026-07-08
Refresh grid_metrics & transform doc notebooks; fetch grid_metrics data from Zenodo

hdrake:
## Summary

Brings the in-repo documentation up to date with `xgcm/master` and makes the example notebooks reproducible by other maintainers.

### `grid_metrics.ipynb` — now reproducible + executed in CI
The MITgcm example dataset (`mitgcm_example_dataset_v2.nc`) was removed from the repo years ago, so the notebook's local `../datasets/...` path was broken — it could neither be reproduced nor re-executed to keep its outputs current. It now downloads the data at run time from Zenodo via `pooch`:

```python
fname = pooch.retrieve(
    url="doi:10.5281/zenodo.21251181/mitgcm_example_dataset_v2.nc",
    known_hash="md5:3176deb0b4d556d2981dff03ba2ed624",
)
```

The archived file is the standard MITgcm `tutorial_global_oce_latlon` / `global_ocean.90x40x15` verification experiment (4°×4°, 90×40×15), recovered byte-for-byte from old git history and deposited at [zenodo.org/records/21251181](https://zenodo.org/records/21251181) (CC0). The notebook is **removed from the `mkdocs` execute-exclude list**, so it now executes on every ReadTheDocs build; it is also added to `mkdocs-jupyter`'s `ignore` list so the `execute` plugin both runs *and* renders it (otherwise both plugins wrap the same `.ipynb` and `--strict` errors). One `xr.testing.assert_equal` was adjusted for the current keep-coords behavior. Outputs refreshed.

### `transform.ipynb` — re-executed against current xarray/xgcm
Re-executed end-to-end. The CMIP6 transform cells rechunk the vertical (`lev`) axis to a single dask chunk (`.chunk({"lev": -1})`), which `Grid.transform` requires; the kernelspec was fixed (`test_env_xgcm` → `python3`). This notebook stays **excluded** from the build (its multi-GB CMIP6 + figshare downloads are too heavy for CI) and is refreshed manually. Outputs refreshed.

### Config / deps
`mkdocs.yml` execute-plugin config, and the doc-notebook dependencies added to the `notebooks` pixi feature: `pooch` (build-time, for the grid_metrics download) and `intake`, `intake-esm`, `gcsfs`, `cf_xarray` (for manual `transform.ipynb` re-execution).

All `.md` doc pages were validated against current xgcm via `mkdocs build --strict` (no changes needed).

## Related
[...]

PR #758, opened 2026-07-08
Fix spurious interpolation warnings in Grid.get_metric

### LINKED ISSUE #757: Grid.get_metric emits spurious "being interpolated" warnings when an exact-position metric exists
## Summary

When `Grid.get_metric` has to assemble a metric by combining metrics registered
on separate axes (for example a volume built from an `(X, Y)` area and a `(Z)`
thickness), it can emit one or more bogus
`UserWarning: Metric at ... being interpolated from metrics at dimensions ...
Boundary value set to 'extend'.` messages even though an exact-position metric
combination exists and is ultimately returned.

The returned value is **correct** — this is purely spurious warning noise — but
it is confusing to users (it implies an inexact/interpolated metric was used)
and pollutes logs. It surfaces through any operation built on `get_metric`,
e.g. `Grid.integrate` and `Grid.average`.

## Reproduction

Using an MITgcm-style set of multi-position metrics (area registered at multiple
horizontal positions, thickness registered at multiple vertical positions, with
no combined volume metric), integrating a staggered variable over `["X", "Y", "Z"]`
emits spurious warnings before settling on the exact metric. In the docs
`grid_metrics` example, `grid.integrate(ds.UVEL, ["X", "Y", "Z"])` (u-point)
emits ~9 warnings and `grid.integrate(ds.SALT, ["X", "Y", "Z"])` (tracer-point)
emits ~2, in each case before returning the correct exact-position metric.

Minimal reproduction against the synthetic test grid:

```python
import warnings
from xgcm import Grid

ds, coords, _ = datasets_grid_metric("C")
# area on (X, Y) and thickness on (Z), but NO combined (X, Y, Z) volume metric.
# The exact-position thickness for the tracer point (dz_t) is registered after
# several wrong-position thicknesses.
metrics = {
    ("X", "Y"): ["area_t", "area_n", "area_e", "area_ne"],
    ("Z",): ["dz_w", "dz_w_ne", "dz_w_n", "dz_w_e", "dz_t"],
}
grid = Grid(ds, coords=coords, metrics=metrics, autoparse_metadata=False)

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    m = grid.get_metric(ds.tracer, ("X", "Y", "Z"))

print(len(caught))  # -> 4 spurious "being interpolated" warnings
# m is nonetheless exactly area_t * dz_t
```

## Root cause

`xgcm/grid.py`, `Grid.get_metric`, the `else` branch (taken when no metric is
registered under the exact requested axes tuple). It loops over candidate metric
combinations via `itertools.product`. For each combination it checks
`metric_dims.issubset(array_dims)`:

- If true (Condition 3, exact position) it sets `metric_vars` and `break`s.
- If false (Condition 4) it calls `warnings.warn(...)` **and** interpolates, but
  does **not** `break`.

Because Condition 4 warns eagerly and does not stop the loop, any
position-mismatched combination iterated *before* the exact-position one emits a
warning; a later iteration then hits Condition 3 and overwrites `metric_vars`
with the correct exact metric. Net effect: correct value, but one or more bogus
warnings.

## Impact

Correct numeric result; spurious `UserWarning`s only. No incorrect data is ever
returned.

## Expected behavior

`get_metric` should search all candidate combinations for an exact-position
match first and only warn/interpolate if no exact-position combination exists
anywhere. When interpolation is genuinely required it should warn exactly once.

---
*This issue was researched and drafted with the assistance of Claude (Anthropic).*

hdrake:
Fixes #757

## What / why

`Grid.get_metric`'s multi-axis combination search (the `else` branch) warned
eagerly whenever a position-mismatched metric combination was tried, without
stopping the loop. When an exact-position combination existed but was not the
first candidate iterated, the correct metric was still returned, but one or more
bogus `UserWarning: Metric at ... being interpolated ...` messages were emitted
first. This is user-facing noise via `Grid.integrate` / `Grid.average` and
implies (falsely) that an interpolated metric was used.

## The fix

Restructure the `else` branch so it searches **all** candidate axis
combinations for an exact-position match (`metric_dims.issubset(array_dims)`)
before considering interpolation, and warns **at most once** when interpolation
is genuinely required (no exact-position combination exists for any candidate).

To avoid changing any existing interpolation result, the interpolation fallback
is chosen deterministically to mirror the pre-fix selection: the last product
combination of the first candidate whose metrics are all present. This is
documented in a code comment. The return contract (product of metrics via
`functools.reduce(operator.mul, ...)`), the `_get_dims_from_axis` validation, the
`KeyError`-swallowing over `iterate_axis_combinations`, and the final
`raise KeyError` are all preserved.

## Tests

Added `test_get_metric_no_spurious_interpolation_warning` in
`xgcm/test/test_metrics.py`, parametrized over a tracer point and a w point.
Each registers area `(X, Y)` and thickness `(Z)` metrics (built by reusing the
`datasets_grid_metric` synthetic grid) with the exact-position thickness placed
after wrong-position ones, then asserts that no "being interpolated" warning is
emitted and that the returned metric equals the exact-position product. The test
fails on `master` (spurious warnings) and passes with this change. The full
`test_metrics.py` + `test_metrics_ops.py` suite passes unchanged.

## Changelog

Added a Bug Fixes entry under the unreleased v0.10.0 section of
`docs/whats-new.md`.

## How I used AI

I was manually reviewing #756 (see https://app.reviewnb.com/xgcm/xgcm/pull/756/discussion/) when I noticed a nonsensical warning, which Claude Code (Opus 4.8 model) helped me trace to this bug. I then asked Claude to open an Issue describing the bug and a minimal draft PR that fixes it and adds a test that passes only after test. I already reviewed the summarized changes as the agent was working on the task but will extensively review all code changes manually.

### DIFF
--- a/xgcm/grid.py
+++ b/xgcm/grid.py
@@ -499,38 +499,64 @@ class Grid:
                 )
                 metric_vars = self.interp_like(mv, array, "extend", None)
         else:
+            # Search ALL candidate axis combinations for an exact-position match
+            # (Condition 3) before considering interpolation (Condition 4).
+            # Otherwise, position-mismatched combinations iterated before the
+            # exact one would emit spurious "being interpolated" warnings even
+            # though an exact-position metric exists (GH #756).
+            #
+            # `interp_fallback` records the combination to interpolate if NO
+            # exact-position combination exists for any candidate. To keep
+            # existing interpolation results (e.g. test_get_metric_with_
+            # conditions_04*) unchanged, it deliberately mirrors the pre-fix
+            # selection: the LAST product combination of the FIRST candidate
+            # whose metrics are all present. Once that first present candidate
+            # has been fully scanned without an exact match, the fallback is
+            # locked so later candidates cannot overwrite it.
+            interp_fallback = None
+            interp_fallback_locked = False
             for axis_combinations in iterate_axis_combinations(axes):
                 try:
                     # will raise KeyError if the axis combination is not in metrics
                     possible_metric_vars = [
                         self._metrics[ac] for ac in axis_combinations
                     ]
-                    for possible_combinations in itertools.product(
-                        *possible_metric_vars
-                    ):
-                        metric_dims = set(
-                            [d for mv in possible_combinations for d in mv.dims]
-                        )
-                        if metric_dims.issubset(array_dims):
-                            # Condition 3: use provided metrics with matching dimensions to calculate for required metric
-                            metric_vars = possible_combinations
-                            break
-                        else:
-                            # Condition 4: metrics in the wrong position (must interpolate before multiplying)
-                            possible_dims = [pc.dims for pc in possible_combinations]
-                            warnings.warn(
-                                f"Metric at {array.dims} being interpolated from metrics at dimensions {possible_dims}. Boundary value set to 'extend'."
-                            )
-                            metric_vars = tuple(
-                                self.interp_like(pc, array, "extend", None)
-                                for pc in possible_combinations
-                            )
-                    if metric_vars is not None:
-                        # return the product of the metrics
-                        metric_vars = functools.reduce(operator.mul, metric_vars, 1)
-                        break
                 except KeyError:
-                    pass
+                    continue
+                for possible_combinations in itertools.product(*possible_metric_vars):
+                    metric_dims = set(
+                        [d for mv in possible_combinations for d in mv.dims]
+                    )
+                    if metric_dims.issubset(array_dims):
+                        # Condition 3: use provided metrics with matching
+                        # dimensions to calculate the required metric.
+                        metric_vars = functools.reduce(
+                            operator.mul, possible_combinations, 1
+                        )
+                        break
+                    if not interp_fallback_locked:
+                        interp_fallback = possible_combinations
+                if metric_vars is not None:
+                    break
+                # Only the first present candidate seeds the interpolation
+                # fallback (see comment above).
+                interp_fallback_locked = True
+
+            if metric_vars is None and interp_fallback is not None:
+                # Condition 4: no exact-position combination exists anywhere, so
+                # interpolate the fallback metrics (warning exactly once).
+                possible_dims = [pc.dims for pc in interp_fallback]
+                warnings.warn(
+                    f"Metric at {array.dims} being interpolated from metrics at dimensions {possible_dims}. Boundary value set to 'extend'."
+                )
+                metric_vars = functools.reduce(
+                    operator.mul,
+                    tuple(
+                        self.interp_like(pc, array, "extend", None)
+                        for pc in interp_fallback
+                    ),
+                    1,
+                )
         if metric_vars is None:
             raise KeyError(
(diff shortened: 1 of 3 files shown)


xgcm/grid.py
```diff
@@ -499,38 +499,64 @@ def get_metric(self, array, axes):
                 )
                 metric_vars = self.interp_like(mv, array, "extend", None)
```
hdrake, line 500: Under what conditions does this get triggered?

hdrake (commented):
I am confused about Condition 2 in the original code and how it interacts with the changes here.

xgcm/grid.py
hdrake: Good question — and worth pinning down, because Condition 2 is easy to conflate with what this PR changes even though it lives in the *other* branch and is untouched here.

**When Condition 2 triggers.** The `if len(overlap_metrics) > 0:` branch is taken when a metric is registered under *exactly* the requested set of axes (in any order) — e.g. you asked for `("X", "Y")` and a metric exists under the key `("X", "Y")`. Condition 1 then looks for one of those metrics that already sits at the array's grid position (`metric_dims.issubset(array_dims)`) and uses it as-is. **Condition 2 is the fallback for when a metric with the matching axes exists but none of them is at the array's position**, so interpolation is genuinely required.

`test_get_metric_with_conditions_02a` is the minimal case: only `area_e` (at `XG, YC`) is registered under `("X", "Y")`, and you ask for the metric of `ds.v` (at `XC, YG`). No exact-position match, so Condition 2 warns and interpolates `area_e` onto v's position. (`02b` adds that this exact-axes match takes priority even when suitably-positioned metrics exist under *other* axis keys.)

**How it interacts with the changes here: it doesn't.** My change is entirely inside the `else` branch — old Conditions 3 & 4 — which is only reached when *no* metric is registered under the requested axes, so the metric has to be assembled by combining metrics from sub-axes (e.g. an `(X, Y)` area × a `(Z)` thickness). The two branches are mutually exclusive on `len(overlap_metrics) > 0`, so Condition 2's behavior is unchanged.

**Why Condition 2 never had the bug — and what this PR does.** Notice Condition 2's `warnings.warn` sits *after* the `for mv in possible_metrics` loop, guarded by `if metric_vars is None:`. It checks every candidate for an exact-position match first and only warns (once) if none is found — which is the correct behavior. The `else` branch was the odd one out: it warned *inside* the `itertools.product` loop (old Condition 4) before an exact-position combination later in that loop could win, so it emitted spurious warnings even though the exact metric was ultimately returned. This PR restructures the `else` branch to mirror what Condition 2 already does: search all candidates for an exact match first, and warn at most once only if none exists. So the net effect is that the two branches now behave consistently.

(One pre-existing smell I deliberately left out of scope: Condition 2 interpolates from `mv`, the leaked loop variable — i.e. the *last* metric registered under that key — which is a bit arbitrary when several are registered. Happy to tidy that in a follow-up if you'd like.)

<sub>↳ reply drafted with Claude Code</sub>

